Issue #13, opened 2019-02-13
Error when trying to use an other bakcend with device: 'qiskit.ibm'


danielmpla:
**Describe the bug**
When executing qml.device('qiskit.ibm', wires=1, backend='ibmqx4') I get the error: "__init__() got multiple values for keyword argument 'backend'". The Exception is thrown at devices.py:424 (super().__init__(wires, backend=backend, shots=shots, **kwargs)).

**To Reproduce**
1. Import: import pennylane as qml
2. Execute: qml.device('qiskit.ibm', wires=1, backend='ibmqx4')

**Expected behavior**
I expect to get a 'qiskit.ibm' device with the 'ibmqx4' backend.

**Desktop Environment:**
 - OS: Linux Manjaro
 - Python: 3.7.2
 - pennylane==0.2.0
 - qiskit==0.7.2
 - pennylane_qiskit==0.0.3

**Additional information**
I think the problem is, that the kwargs contain backend='ibmqx4' and additionally there is the argument backend=backend in line 424 (from reading the code this should also contain my specified backend). So I assume after assigning the backend from the kwargs to the variable backend, backend should be removed from the kwargs (if present).

danielmpla:
As it turns out you probably can't remove an item from **kwargs. So my temporary solution is:

```
if 'backend' in kwargs.keys():   
            super().__init__(wires, shots=shots, **kwargs)
else:   
            super().__init__(wires, backend=backend, shots=shots, **kwargs)
```

carstenblank:
Hi @danielmpla! Thank you so much for pointing this out. Ouch! Okay I will see to it.

carstenblank:
I think the best way to solve this, is to make the the arguments `backend` and `shots` in the `__init__` explicit such that it wont ever be part of `kwargs`. I believe this should resolve the issue. I am working on a pull request to fix this.

danielmpla:
Yeah I think this solution is cleaner and improves probably also the code completion when using an IDE. I would close the issue after testing the new release.

sebboer:
fix published in version 0.0.5

PR #14, opened 2019-02-13
Multiple keywords fix

carstenblank:
Regarding Issue #13 this is my suggestion to fix the bug:

make the arguments `backend` and `shots` on the `__init__` function explicit so that there cannot be any duplicates on the dict `kwargs`.

Changes have been done on the file `devices.py` and tests have been added to `test_device_initialization.py`.


sebboer:
fix published in version 0.0.5

PR #22, opened 2019-03-15
add device noise model support

sebboer:


carstenblank:
I am now reviewing this PR, sorry for the delay. It looks good, I have added a couple of changes still. For one, the devices should still be configurable during the tests, so a decided to just re-instantiate new devices if they are configured to run, but with a noise model.
The code itself looks good.

PR #29, opened 2019-06-08
Fix/after expectation fix

carstenblank:
After the expectation fix by @josh146 one test, test_compare_with_default_qubit.py, actually got re-activated and failed. 
It showed that base functionality was not working and also showed that for sime operations as the QubitStateVector and others deviated from the default behavior of the DefaultQubit device.

A lot of fixes are thus necessary, with plenty of refactorings done as well.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ info = {
         'pennylane.plugins': [
             'qiskit.aer = pennylane_qiskit:AerQiskitDevice',
             'qiskit.basicaer = pennylane_qiskit:BasicAerQiskitDevice',
-            'qiskit.ibm = pennylane_qiskit:IbmQQiskitDevice',
+            'qiskit.ibmq = pennylane_qiskit:IbmQQiskitDevice',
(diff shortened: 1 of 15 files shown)


setup.py
```diff
@@ -38,7 +38,7 @@
         'pennylane.plugins': [
             'qiskit.aer = pennylane_qiskit:AerQiskitDevice',
```
josh146: Hey @carstenblank, I'm just curious - in the documentation, you mention `qiskit.basicaer` and `qiskit.ibmq`, but never `qiskit.aer`. Is this a working device that you can also use with this plugin?

carstenblank: Hey @josh146, it is indeed possible and recommended to use the `qiskit.aer` device. The reason this wasn't documented was because of historic reasons. During the time when the qiskit team pulled out the aer from `qiskit-terra` it was dysfunctional. But all issues seem to have been fixed.
I am adding more documentation right now to the readme.

Issue #28, opened 2019-05-18
Support more expectation values


josh146:
Currently, the Qiskit devices only allow for `PauliZ` expectation values.

However, it is simple to support _all_ the remaining PennyLane expectation values: `PauliX`, `PauliY`, `Hadamard`, and `Hermitian`.

---

For the first three, this is relatively simple, and can be done as follows:

![image](https://user-images.githubusercontent.com/2959003/57975062-c3e00200-7990-11e9-8894-eba25eab8e29.png)

That is,

1. Peak at the expectation queue using `self.expval_queue`

2. Apply the corresponding change of basis unitary U as per the table above

3. Return to the user the resulting \<Z\> expectation value

---

For arbitrary 2x2 Hermitian matrices, the process is only slightly more complicated:

![image](https://user-images.githubusercontent.com/2959003/57975087-34871e80-7991-11e9-8f81-51581a90148a.png)

That is:

1. Calculate the eigenvector matrix U and the corresponding eigenvalues lambda of the provided Hermitian matrix

2. Apply the change of basis matrix U^\dagger to the end of the users circuit

3. Measure the wavefunction probabilities p

4. The resulting expectation value is given by the dot product lambda.p

josh146:
For example, see how this is done in the PennyLane-Forest plugin, here:

https://github.com/rigetti/pennylane-forest/blob/multi_qubit_observables/pennylane_forest/qvm.py

Specifically, the `pre_expval` and `expval` methods.

carstenblank:
Nice! Thank you so much Josh! I will implement this the next days I think!

josh146:
Let me know if you have any questions!

carstenblank:
Do you think we solved this issue with #32 ?

josh146:
I think it is, yes :slightly_smiling_face: - more features can still be added (such as multi-qubit expectations), but that is out of scope and can be discussed in a new issue.

Issue #8, opened 2019-01-29
Expose the various kwargs of the qiskit backends


cgogolin:
The qiskit backends have a wealth of keyword arguments with which their behavior can be fine tuned. It would be great to expose them via kwargs of the corresponding devices and also document them in the pennylane-qiskit docs.

This is related to #3

PR #32, opened 2019-06-20
Feature/backend options and more expectation values

carstenblank:
This PR will address issues #8 and #28 . Albeit having a weird branch naming (it got a little hectic today while transferring the project to Xanadu!) this is a major change in features and code!

In more detail:
* #8 has been added especially in the `devices.QiskitDevice#_compile_and_execute` function where the known backends are listed and the `backend_options` are passed in (as with the noise model). The tests are found in `test_backend_options.py` and they caused some head-ache for me until I finally realized what it was I want to test. The tests `test_basicaer_initial_unitary` and `test_basicaer_chop_threshold` are basically useless but helped me to root out bugs that the current tests didn't catch.

* #23 Now this one was basically inspired by the `pennylane-forest` plugin and @josh146 explanations. I added the unit tests from `pennylane-forest` too. A lot of adjustments were necessary and another bug found in the (rather ugly) `qiskitops.QubitUnitary`. I added the documentation too.

This should be reviewed as soon as CI as up and running again! 

@josh146  can you review this maybe?

### DIFF
--- a/pennylane_qiskit/devices.py
+++ b/pennylane_qiskit/devices.py
@@ -114,12 +114,13 @@ class QiskitDevice(Device):
         'model': 'qubit'
     }  # type: Dict[str, any]
     _operation_map = QISKIT_OPERATION_MAP
-    _expectations = {'PauliZ'}
+    _expectations = {'PauliX', 'PauliY', 'PauliZ', 'Identity', 'Hadamard', 'Hermitian'}
     _backend_kwargs = ['verbose', 'backend']
     _noise_model = None  # type: Optional[NoiseModel]
     _unitary_result_backends = [UnitarySimulator().name(), UnitarySimulatorPy().name()]
     _statevector_result_backends = [StatevectorSimulator().name(), StatevectorSimulatorPy().name()]
     _unitary_backend_initial_state = None
+    _eigs = {}  # type: Dict[str, Dict[str, np.ndarray]]
 
     _no_measure_backends = _unitary_result_backends + _statevector_result_backends
 
@@ -221,28 +222,46 @@ class QiskitDevice(Device):
 
         # Add unitaries if a different expectation value is given
         for e in self.expval_queue:
-            wires = e.wires
+            wire = [e.wires[0]]
 
-            if e.name == 'PauliZ':
-                pass # nothing to be done here!
+            if e.name == 'Identity':
+                pass # nothing to be done here! Will be taken care of later.
+
+            elif e.name == 'PauliZ':
+                pass # nothing to be done here! Will be taken care of later.
 
             elif e.name == 'PauliX':
                 # X = H.Z.H
-                self.apply('Hadamard', wires=wires, par=[])
+                self.apply('Hadamard', wires=wire, par=[])
 
             elif e.name == 'PauliY':
                 # Y = (HS^)^.Z.(HS^) and S^=SZ
-                self.apply('PauliZ', wires=wires, par=[])
-                self.apply('S', wires=wires, par=[])
-                self.apply('Hadamard', wires=wires, par=[])
+                self.apply('PauliZ', wires=wire, par=[])
+                self.apply('S', wires=wire, par=[])
+                self.apply('Hadamard', wires=wire, par=[])
 
             elif e.name == 'Hadamard':
                 # H = Ry(-pi/4)^.Z.Ry(-pi/4)
-                self.apply('RY', wires, [-np.pi / 4])
+                self.apply('RY', wire, [-np.pi / 4])
 
             elif e.name == 'Hermitian':
-                # TODO
-                raise ValueError("Hermitian expectation value is not implemented yet!")
+                # For arbitrary Hermitian matrix H, let U be the unitary matrix
+                # that diagonalises it, and w_i be the eigenvalues.
+                H = e.parameters[0]
+                Hkey = tuple(H.flatten().tolist())
+
+                if Hkey in self._eigs:
+                    # retrieve eigenvectors
+                    U = self._eigs[Hkey]['eigvec']
+                else:
+                    # store the eigenvalues corresponding to H
+                    # in a dictionary, so that they do not need to
+                    # be calculated later
+                    w, U = np.linalg.eigh(H)
+                    self._eigs[Hkey] = {'eigval': w, 'eigvec': U}
+
+                # Perform a change of basis before measuring by applying U^ to the circuit
+                self.apply('QubitUnitary', wire, [U.conj().T])
 
             else:
                 raise ValueError("The expectation %s is unknown!", e.name)
@@ -291,6 +310,21 @@ class QiskitDevice(Device):
         one = sum(p for (measurement, p) in probabilities.items() if measurement[::-1][wire] == '1')
 
         expval = (1 - (2 * one) - (1 - 2 * zero)) / 2
+
+        # for single qubit state probabilities |psi|^2 = (p0, p1),
+        # we know that p0+p1=1 and that <Z>=p0-p1
+        p0 = (1 + expval) / 2
+        p1 = (1 - expval) / 2
+
+        if expectation == 'Identity':
+            # <I> = \sum_i p_i
+            return p0 + p1
+
+        if expectation == 'Hermitian':
+            # <H> = \sum_i w_i p_i
+            Hkey = tuple(par[0].flatten().tolist())
+            w = self._eigs[Hkey]['eigval']
+            return w[0] * p0 + w[1] * p1
 
         return expval
 
@@ -347,7 +381,12 @@ class BasicAerQiskitDevice(QiskitDevice):
       :class:`pennylane.BasisState`
 
     Supported PennyLane Expectations:
+      :class:`pennylane.expval.PauliX`
+      :class:`pennylane.expval.PauliY`
       :class:`pennylane.expval.PauliZ`
+      :class:`pennylane.expval.Identity`
+      :class:`pennylane.expval.Hadamard`
+      :class:`pennylane.expval.Hermitian`
 
     Extra Operations:
       :class:`pennylane_qiskit.S <pennylane_qiskit.ops.S>`,
@@ -422,7 +461,12 @@ class AerQiskitDevice(QiskitDevice):
       :class:`pennylane.BasisState`
 
     Supported PennyLane Expectations:
+      :class:`pennylane.expval.PauliX`
+      :class:`pennylane.expval.PauliY`
       :class:`pennylane.expval.PauliZ`
+      :class:`pennylane.expval.Identity`
+      :class:`pennylane.expval.Hadamard`
+      :class:`pennylane.expval.Hermitian`
 
     Extra Operations:
       :class:`pennylane_qiskit.S <pennylane_qiskit.ops.S>`,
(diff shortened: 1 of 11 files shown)


josh146:
This looks great @carstenblank, I'll have a look at it now!

pennylane_qiskit/devices.py
```diff
@@ -52,19 +52,28 @@
...
 from pennylane import Device, DeviceError
+from pennylane import numpy as np
 from qiskit import QuantumRegister, ClassicalRegister, QuantumCircuit
 from qiskit.circuit import Gate
 from qiskit.circuit.measure import measure
 from qiskit.converters import dag_to_circuit, circuit_to_dag
 from qiskit.extensions import XGate, RXGate, U1Gate, HGate, RYGate, RZGate, CzGate, CnotGate, YGate, ZGate, SGate, \
     TGate, U2Gate, U3Gate, SwapGate
 from qiskit.providers import BaseProvider, BaseJob, BaseBackend
+from qiskit.providers.aer import StatevectorSimulator, UnitarySimulator
```
josh146: Nice, are these two new backends available for PennyLane?

pennylane_qiskit/devices.py
```diff
@@ -105,18 +114,24 @@ class QiskitDevice(Device):
         'model': 'qubit'
     }  # type: Dict[str, any]
     _operation_map = QISKIT_OPERATION_MAP
-    _expectations = {'PauliZ'}
+    _expectations = {'PauliX', 'PauliY', 'PauliZ', 'Identity', 'Hadamard', 'Hermitian'}
```
josh146: Awesome!

pennylane_qiskit/devices.py
```diff
@@ -171,20 +191,25 @@ def apply(self, operation, wires, par):
...
+                self._current_job = backend.run(qobj, noise_model=self._noise_model, backend_options=self.kwargs)
+
+            elif isinstance(backend, QasmSimulatorPy) and isinstance(self, BasicAerQiskitDevice):
+                self._current_job = backend.run(qobj, backend_options=self.kwargs)
+
+            elif isinstance(backend, StatevectorSimulatorPy) and isinstance(self, BasicAerQiskitDevice):
+                self._current_job = backend.run(qobj, backend_options=self.kwargs)
+
+            elif isinstance(backend, UnitarySimulatorPy) and isinstance(self, BasicAerQiskitDevice):
+                self._current_job = backend.run(qobj, backend_options=self.kwargs)
```
josh146: Apart from `AerBackend`, where the noise model option is available, is it possible to merge all of these into one else statement?
```python
else:
    self._current_job = backend.run(qobj, backend_options=self.kwargs)
```

josh146:
It seems the continuous integration is no longer running, let me turn that on.

tests/test_expectations.py
```diff
@@ -0,0 +1,181 @@
...
+
+log.getLogger('defaults')
+
+
+class TestQVMBasic(BaseTest):
+    """Unit tests for the QVM simulator."""
+    # pylint: disable=protected-access
+
+    num_subsystems = 2
+    shots = 16 * 1024
```
quantshah: This leads to an error from qiskit saying that the number of shots exceed some limit. I changed this to 8*1024 and it worked.

tests/test_simple_circuits.py
```diff
@@ -35,20 +35,21 @@ class SimpleCircuitsTest(BaseTest):
     """
 
     num_subsystems = 4
+    shots = 16 * 1024
```
quantshah: The number of shots exceeds some limit set by IBM Q and this test fails. Changing this to 8*1024 works.

quantshah:
Hi, I would like to help with this. I fetched this PR and ran all the tests with my own token. The device initialization test fails and should be updated I guess with the new qiskit features.
The error message that I get is:  

```
pkg_resources.ContextualVersionConflict: (jsonschema 3.0.0a3 
(/Users/shahnawaz/anaconda3/lib/python3.7/site-packages), 
Requirement.parse('jsonschema<2.7,>=2.6'), 
{'qiskit-terra', 'qiskit-chemistry', 'qiskit-aqua'})
```

@josh146 How can one run CI on this without having a public token?

In addition, I would suggest:
 - Moving to pytests: I could do this after this PR is merged.
 - Adding some information about the new token and url based activation of qiskit. E.g., 

=========================================================================
Tests on the `ibm provider <https://pennylane-qiskit.readthedocs.io/en/latest/devices.html>`_ can only be run if a :code:`ibmqx_token` for the `IBM Q experience <https://quantum-computing.ibm.com/>`_ is configured. This can be done in two ways:  

- setting the IBMQX_TOKEN with `export IBMQX_TOKEN=token`
- configuring the `PennyLane configuration file in <https://pennylane.readthedocs.io/en/latest/code/configuration.html>`_

You also need to enable your QISKIT `account <https://qiskit.org/documentation/install.html>`_  for the first time:

        from qiskit import IBMQ
        IBMQ.save_account(token, url)
        IBMQ.enable_account(token)
=========================================================================

pennylane_qiskit/devices.py
carstenblank: certainly, they are specific backends of qiskit-aer. You need to use the argument `backend` to use one of them. They are also given by the `capabilities()` function under the key `'backends'`.

pennylane_qiskit/devices.py
carstenblank: You could, I don't want to though. I expect changes and wanted to 'document' that we have the ability to go into detail with any Device. If you insist we can change this of course!

tests/test_expectations.py
carstenblank: Oh that can actually be the case for the IBM Q backends. Very unfortunate though! Because many tests will fail because of very strong deviations from randomness. Any idea how to achieve both goals: satisfying constraints and failing by randomness?

carstenblank:
I changed the shots for IBMQ devices and did some refactoring and fixing when no token is present. The tests failed then. I think we want to have the tests without really checking if IBMQ is working, with all the new expectation values the tests take about 60 minutes!!

tests/test_expectations.py
carstenblank: I chose to solve this by using 16*1024 shots for the local backends and 8*1024 for the IBMQ. Then it won't fail as often.... not so convincing though!

quantshah:
Thanks for the updates @carstenblank. Would it be possible to make me a collaborator on this project?
 @josh146 

That way I could make changes directly to this PR or commit things if you have allowed collaborators to make changes to the PR. I have the following things in mind:

- get a basic working version out with this merge
- update the instructions on how to use this with pennylane and latest IBM tokens
- Add some more examples/tutorials

quantshah:
Thanks for adding me as a collaborator. It seems there are some small tests for exception handling that are leading to `codecov` being unsatisfied:

https://codecov.io/gh/XanaduAI/pennylane-qiskit/compare/01e38e2a2bb1e3403f255e80c09aafcb5e37ddc4...4e0d8b1e7d8feff97fb09cbbad8df7fde49be338/diff

I will try to smoothen them out here. Or we could merge this and then do another PR with all the test refactored to pytest and increase the coverage.

pennylane_qiskit/devices.py
```diff
@@ -171,20 +191,25 @@ def apply(self, operation, wires, par):
         else:
```
quantshah: This part of the code is a bit confusing. Particularly, this error should be raised higher up I think. We want to see if the user has supplied a correct operation which is mapped. So something like

```
device.apply("RXY", wires=[1])
```
should raise an error since it is not a mapped operation. But with the code now, it will raise a key error as 

```
Traceback (most recent call last):
  File "/Users/shahnawaz/Dropbox/dev/pennylane-qiskit/tests/test_unsupported_operations.py", line 67, in test_unsupported_operation
    device.apply('RXY', wires=[0], par=[0.1])
  File "/Users/shahnawaz/Dropbox/dev/pennylane-qiskit/pennylane_qiskit/devices.py", line 170, in apply
    mapped_operation = self._operation_map[operation]
KeyError: 'RXY'
```

So, I suggest the check to happen before we even do `mapped_operation = self._operation_map[operation]` as,
```
try:
    mapped_operation = self._operation_map[operation]
except KeyError:
    raise ValueError("")
```

quantshah: Also, if we make sure that the operation is one of the supported ones in `QISKIT_OPERATION_MAP`, we do not need to check beyond the two possible types that we allow in the map. Therefore, I am adding an error message and checking if the operation really is valid in the beginning with something like:

```
ValueError: The operation is not of an expected type. Supported QISKIT operations and instructions are: PauliX, PauliY, PauliZ, CNOT, CZ, SWAP, RX, RY, RZ, PhaseShift, Hadamard, Rot, BasisState, QubitStateVector, QubitUnitary, S, T, U1, U2, U3
```

pennylane_qiskit/devices.py
```diff
@@ -137,6 +152,11 @@ def operations(self):
     def expectations(self):
         return set(self._expectations)
 
+    @property
+    def backend(self):
+        # type: () -> BaseBackend
+        return self._provider.get_backend(self.backend_name)
+
     def apply(self, operation, wires, par):
```
quantshah: Some operations might not have parameters. Better to perhaps put `par = None` here?

pennylane_qiskit/devices.py
josh146: Something to note: the `Device` base class in PennyLane currently has built-in checks for supported observables and operations. When `Device.execute()` is called, before any `apply()` calls are made, the device calls `Device.check_validity` (see here: https://github.com/XanaduAI/pennylane/blob/478e0fa6fd692768105002d8a09d99e70ae7644b/pennylane/_device.py#L303).

This checks if the queued operations and observables are in the `Device.operations` and `Device.observables` properties; if not, an exception is raised.

pennylane_qiskit/devices.py
josh146: This is an overwritten abstract method of `Device`, so the signature will need to match :)

josh146: See https://github.com/XanaduAI/pennylane/blob/478e0fa6fd692768105002d8a09d99e70ae7644b/pennylane/_device.py#L321 (cf. [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle))

tests/test_expectations.py
```diff
@@ -0,0 +1,214 @@
...
+        phi = 0.123
+        for dev in self.devices:
+            dev.apply('RX', wires=[0], par=[theta])
+            dev.apply('RX', wires=[1], par=[phi])
+            dev.apply('CNOT', wires=[0, 1], par=[])
+
+            O = qml.expval.qubit.Identity
+            name = 'Identity'
+
+            dev._expval_queue = [O(wires=[0], do_queue=False), O(wires=[1], do_queue=False)]
```
quantshah: The queue for `expval` for the device seems to be `expval_queue` and not `_expval_queue`. I do not understand this override.

quantshah:
Hi, it seems all tests pass. If this is satisfactory then I would like to merge this and start to port the tests to pytest and change things to support the latest pennylane API with the new `measure` module which allows things like `pennlane.expval(PauliX(wires=[0]))`. Then, we can update the blogs and examples and release a new version perhaps?

@josh146 @carstenblank

tests/test_expectations.py
quantshah: Sorry, at the moment I have no idea how to do that but seems like the current settings work. Maybe we need to move the in-depth testing on the IBM device somewhere else? Like a dev-branch?

carstenblank: I agree! The testing takes really really long...!!

pennylane_qiskit/devices.py
quantshah: Thanks. Then, probably we should make the check happen before the `.apply` method is called in the `check_validity` function.

co9olguy:
@carstenblank we'd like to merge this in. Are there any more changes that you are planning to make, or are we ok to go ahead :+1: ?

carstenblank:
This PR looks good, thanks @quantshah for your input. I will merge this now as we have agreed!

carstenblank:
I will wait another 2-3 days and then delete the feature branch.

co9olguy:
Thanks @carstenblank!

PR #44, opened 2019-08-04
Plugin cleanup and refactor

josh146:
**Description of changes:**

This is quite a significant clean up and refactor of the plugin. In particular, 

* All code blocks not being used were removed

* Superfluous conditionals were deleted

* Unused methods and attributes were deleted

* Documentation and licenses were updated to bring them up to standard

* Long and confusing names were shorted

* `device.py` was split into separate files, one per device

* The long standing IBM Q bug (see #38 and #39) was fixed, and the IBM Q device now supports both v1 and v2 of the IBM API

* The deprecated custom gates in `qiskitops.py` were removed, and replaced with the recommended approach to defining Qiskit custom gates.

**To-do:** None of the tests were refactored, this is still on the todo list.

PR #45, opened 2019-08-05
Test refactor

quantshah:
This PR will refactor the tests with a port to pytest, allowing for a cleaner and more readable test suite.


josh146:
Hey @quantshah, before working too much on this, maybe we should merge in #44 so that they don't go out of sync!

quantshah:
Yes, I will do the review now and then continue this.

quantshah:
This is ready for review and merge. I left two tests which still need to be refactored -`test_backend_options` and `test_simple_circuits`. I would like to work on them in a different PR. They are a bit convoluted and we need to implement some way to get the IBM token working so that tests can run on the real devices too as we discussed. 

But I want to get this merged first. Feel free to suggest changes @josh146

tests/test_basis_state.py
```diff
@@ -14,105 +14,92 @@
...
+
+elif "IBMQX_TOKEN" in os.environ and os.environ["IBMQX_TOKEN"] is not None:
+    IBMQX_TOKEN = os.environ["IBMQX_TOKEN"]
+
+devices = [BasicAerDevice(wires=num_wires), AerDevice(wires=num_wires)]
+
+if IBMQX_TOKEN is not None:
+    devices.append(
+        IbmQQiskitDevice(wires=num_wires, num_runs=8 * 1024, ibmqx_token=ibmqx_token)
+    )
```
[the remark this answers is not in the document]
josh146: The above also allows for the different backends of the providers to be tested (currently, only the default Aer `qasm_simulator` backend is being tested).

tests/test_expectations.py
```diff
@@ -20,195 +20,239 @@
 from defaults import pennylane as qml, BaseTest
 from pennylane_qiskit import BasicAerDevice, AerDevice, IBMQDevice
 import unittest
+import pytest
+import os
 
-log.getLogger('defaults')
+log.getLogger("defaults")
```
josh146: We should be able to remove the logger calls in the tests now

tests/test_expectations.py
```diff
@@ -20,195 +20,239 @@
 from defaults import pennylane as qml, BaseTest
```
josh146: Best to replace this with `import pennylane as qml` here

tests/test_expectations.py
```diff
@@ -20,195 +20,239 @@
...
+           AerDevice(wires=num_wires, shots=shots)]
 
-    def setUp(self):
-        super().setUp()
+if IBMQX_TOKEN is not None:
+    devices.append(
+        IbmQQiskitDevice(wires=num_wires,
+                         num_runs=ibmq_shots,
+                         ibmqx_token=ibmqx_token)
+    )
```
josh146: Same comment re: previous test about testing additional backends/IBMQ token.

tests/test_expectations.py
```diff
@@ -20,195 +20,239 @@
...
+ibmq_shots = 100
+devices = [BasicAerDevice(wires=num_wires, shots=shots),
+           AerDevice(wires=num_wires, shots=shots)]
 
-    def setUp(self):
-        super().setUp()
+if IBMQX_TOKEN is not None:
+    devices.append(
+        IbmQQiskitDevice(wires=num_wires,
+                         num_runs=ibmq_shots,
```
josh146: this should be shots rather than num_runs

tests/test_expectations.py
```diff
@@ -20,195 +20,239 @@
...
-
-            O = qml.PauliZ
-            name = 'PauliZ'
+    # below are the analytic expectation values for this circuit (trace should always be 1)
+    assert np.allclose(
+        res,
+        np.array([1, 1]),
+        atol=3 / np.sqrt(device.shots),
+        rtol=3 / np.sqrt(device.shots),
+    )
```
josh146: nice!

tests/test_expectations.py
```diff
@@ -20,195 +20,239 @@
...
+    device.apply("CNOT", wires=[0, 1], par=[])
+
+    O = qml.PauliZ
+    name = "PauliZ"
+
+    device._obs_queue = [O(wires=[0], do_queue=False), O(wires=[1], do_queue=False)]
+    res = device.pre_measure()
+
+    res = np.array([device.expval(name, [0], []), device.expval(name, [1], [])])
+    print(res)
```
josh146: Don't forget to remove print statements!

tests/test_expectations.py
```diff
@@ -20,195 +20,239 @@
...
+    device.pre_measure()
+
+    res = np.array([device.expval(name, [0], []), device.expval(name, [1], [])])
+    # below are the analytic expectation values for this circuit
+    assert np.allclose(
+        res,
+        np.array([np.sin(theta) * np.sin(phi), np.sin(phi)]),
+        atol=3 / np.sqrt(device.shots),
+        rtol=3 / np.sqrt(device.shots),
+    )
```
josh146: Awesome tests :)

tests/test_basis_state.py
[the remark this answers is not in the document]
josh146: Thinking about this a bit more, you could maybe turn the above into fixtures, and move them into conftest.py. Further, you could attempt to instantiate a IBMQ device, and if it fails, mark the IBMQ device fixture to be skipped via pytest.skip.

tests/test_simple_circuits.py
```diff
@@ -26,9 +26,198 @@
...
+    """Tests if correct error is raised when attempting multi-qubit unitaries"""
+    test_input = np.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]])
+    @qml.qnode(device)
+    def circuit():
+        qml.QubitUnitary(test_input, wires=[0])
+        return qml.expval(qml.PauliZ(0))
+
+    with pytest.raises(ValueError,
+                       match="Only a single qubit unitary is supported."):
+        circuit()
```
josh146: Hopefully we should be able to change this once qiskit v0.12 is released :) It contains support for a two-qubit `QuantumCircuit.unitary` operation, as well as an `isometry` module for arbitrary unitary decomposition.

tests/test_simple_circuits.py
```diff
@@ -26,9 +26,198 @@
...
+    def circuit():
+        qml.QubitUnitary(test_input, wires=[0])
+        return qml.expval(qml.PauliZ(0))
+
+    with pytest.raises(ValueError,
+                       match="Not a unitary"):
+        circuit()
+
+
+class CompareWithDefaultQubitTest(BaseTest):
```
josh146: Agree that this is too convoluted - I would like to remove this test, and replace it with short explicit test functions. These can likely be copied over from PennyLane core, or PennyLane-Forest. But as you mentioned, better to tackle in a later PR.

tests/test_simple_circuits.py
```diff
@@ -26,9 +26,198 @@
...
+    """Compares the behavior of the ProjectQ plugin devices with the default qubit device.
```
josh146: Looks like this was originally copied from ProjectQ :laughing:

josh146:
Closing as this has been superceded by a recent test refactor

Issue #42, opened 2019-07-29
Support sampling observables


josh146:
**Context:**

With the latest development version of PennyLane, we now support returning `samples` from the QNode, as well as expectation values and variance:

```python
@qml.qnode(dev)
def circuit(x):
    qml.RX(x, wires=0)
    qml.CNOT(wires=[0, 1])
    return expval(qml.PauliZ(0)), sample(qml.PauliZ(1))
```

Plugins can support this new feature by subclassing the `Device.sample()` method, returning the requested number of samples for a given observable.

---

This should be relatively easy to support on hardware; for example,

```python
def sample(self, observable, wires, par, n=None):
    if n is None:
        n = self.shots

    if n == 0:
        raise ValueError("Calling sample with n = 0 is not possible.")

    if n < 0 or not isinstance(n, int):
        raise ValueError("The number of samples must be a positive integer.")

    if observable == "Identity":
        return np.ones([n])

    # resulting hardware samples
    res = [np.array([int(i) for i in state]*count) for state, count in results.get_count.items()]
    res = np.concatenate(res).reshape(-1, self.num_wires)
    res = res[:, np.array(wires)]

    if observable == "Hermitian":
        Hkey = tuple(par[0].flatten().tolist())
        eigvals = self._eigs[Hkey]["eigval"]

       # observable samples
        samples = np.zeros([n])

        for w, b in zip(eigvals, itertools.product([0, 1], repeat=len(wires))):
            # iterate through basis state and eigenvalues,
            # repeating the former with the latter
            samples = np.where(np.all(res == b, axis=1), w, samples)

        return samples

    return 1 - 2 * res
```

_Note: this is just a rough implementation. It could also be implemented using probabilities directly, or maybe there is a more direct way to extract sampling information_.

---

A major benefit of supporting sampling is that is significantly reduces the complexity of the code  required to compute expectations and variances on hardware. In #40, probabilities need to be computed for multi-qubit observables. Now, we can simply call `np.mean(self.samples())` and `np.var(self.samples())`.

PR #46, opened 2019-08-07
Add samples and variance support

josh146:
The last PR to get the Qiskit plugin up to feature parity with the latest PennyLane. This PR adds the following:

* Support for sampling. If a hardware device, the keyword argument `memory=True` is passed to the executor to force Qiskit to return raw sampling data. If a state simulator, sampling data is generated from the statevector probabilities by doing a spectral decomposition of the observable.

* Support for variance. On hardware devices, this is done by simply taking the variance of the samples generated via probability distributions. On state simulators, this is done by doing either (a) doing an inner product directly, or (b) if shots > 0, taking the variance of the samples.

* Speed up expectation value calculations on state simulators. Previously, all state simulators were treated as hardware devices. Now, the expectation value is calculated using the same approach as variance:  (a) doing an inner product directly, or (b) if shots > 0, taking the mean of the samples.

This PR addresses #42


tests/test_backend_options.py
```diff
@@ -90,27 +90,27 @@ def circuit():
 
                 log.info("Outcome: %s", measurement)
 
-    def test_basicaer_chop_threshold(self):
-        """Test BasisState with preparations on the whole system."""
-        if self.devices is None:
-            return
-        self.logTestName()
+    # def test_basicaer_chop_threshold(self):
```
quantshah: Is this test supposed to be commented out?

josh146: A bit embarrassing, but I spent a good couple of hours attempting to debug this test, but (a) couldn't work out what it was supposed to do, and (b) couldn't work out why any of my changes broke it. I figured with your test refactor PR happening at the same time, we could replace it and then see what happened :laughing:

Issue #40, opened 2019-07-23
Hermitian op only works for single-qubit observables


co9olguy:
The base PennyLane library now how support for multi-qubit Hermitian observables (see [here](https://github.com/XanaduAI/pennylane/pull/192)). However, the pennylane-qiskit plugin only works for single-qubit observables. 

@quantshah @josh146 Are there any restrictions in qiskit that prevent us from implementing support for multi-qubit Hermitian observables?

josh146:
No, it is just a matter of implementing it. This has already been implemented on hardware devices over at https://github.com/rigetti/pennylane-forest, see [qvm.py](https://github.com/rigetti/pennylane-forest/blob/master/pennylane_forest/qvm.py). I can outline the process here.

1. Before executing the device, 'peak' at the observable queue in `pre_measure`. If an Hermitian observable exists in the queue, compute the eigendecomposition U D U^\dagger.

    Store the eigenvalues for later, and prepend the unitary U^\dagger to the end of the operation queue.

    _This step should already be implemented, as it is required for arbitrary single-qubit Hermitian observables._

2. After device execution, in both `var` and `expval`:
    
    - For only the subsystems in which the multi-qubit observable is measured on, compute the probability distribution over all computational basis states.

      This could be done manually from the sampling results (see the PennyLane-Forest [`QVMDevice.probabilities()`](https://github.com/rigetti/pennylane-forest/blob/master/pennylane_forest/qvm.py#L214) method), or perhaps Qiskit provides built in functions for this ( `results.get_counts()`?)

3. Expectation values are then given by `D @ probs`

4. Variance values are given by `(D ** 2) @ probs - (D @ probs) ** 2`

quantshah:
Thanks @josh146 for explaining that. Will get on to it.

PR #49, opened 2019-09-06
Adds support for multi-qubit tensor observables

josh146:
**Context:** Previously, the Qiskit plugin did not support multi-qubit observables.

**Description of changes:**

* Adds support for multi-qubit observables via the following approach:

  1. All wires terminating in anything other than `PauliZ` are rotated via single qubit observables (`PauliX`: `H`, `PauliY`: `H.S.Z`, `Hadamard`: `RY(pi/4)`, `Hermitian`: numerical eigenbasis `U`).

  2. This modifies the circuit to measure (Z \otimes Z \otimes ... \otimes Z)

  3. The Pauli-Z basis is measured for each wire, and then classical post-processing is used to determine the expectation values (`ev = eigvals @ prob`) and variance (`var = (eigvals**2 @ prob) - (eigvals @ prob)**2`) where `eigvals` are the eigenvalues of the original observable that was to be measured pre-transform.

* To compute the eigenvalues of the `qml.Hermitian`, `np.linalg.eigvalsh` is used.

* To compute the eigenvalues of arbitrary tensor products of `{X, Y, Z, H}`, an efficient function `z_eigs` is provided that uses dynamic programming.

**Benefits:**

* Arbitrary multi-qubit tensor observables can now be measured, on both simulator and hardware, using the Qiskit plugin.

* By treating hardware devices and simulators _identically_, we avoid the naive approach of simply using `np.kron()` and `np.tensordot` on the simulator wavefunction, which does not scale as the number of qubits increases.

**Drawbacks:**

* The assumption is made that the PennyLane plugin API for specifying tensor observables is simply passing a list of observables to `Device.expval`, `Device.var`, and `Device.sample`. If this assumption is wrong or the API change, this will also have to change.

**Related issues:** #40


tests/test_sample.py
```diff
@@ -151,3 +151,173 @@ def test_sample_exception_wrong_n(self, device, shots):
...
+        var = np.var(s1)
+        expected = (
+            8 * np.sin(theta) ** 2 * np.cos(2 * varphi) * np.sin(phi) ** 2
+            - np.cos(2 * (theta - phi))
+            - np.cos(2 * (theta + phi))
+            + 2 * np.cos(2 * theta)
+            + 2 * np.cos(2 * phi)
+            + 14
+        ) / 16
+        assert np.allclose(var, expected, **tol)
```
johannesjmeyer: If you write a stochastic test, then the tolerance should depend on the number of shots, otherwise it is arbitrary.

johannesjmeyer: (Better not to write stochastic tests at all, I think settling to checking that the underlying function behaves as expected should be enough)

tests/test_sample.py
```diff
@@ -151,3 +151,173 @@ def test_sample_exception_wrong_n(self, device, shots):
...
+        expected = 0.5 * (
+            -6 * np.cos(theta) * (np.cos(varphi) + 1)
+            - 2 * np.sin(varphi) * (np.cos(theta) + np.sin(phi) - 2 * np.cos(phi))
+            + 3 * np.cos(varphi) * np.sin(phi)
+            + np.sin(phi)
+        )
+        assert np.allclose(mean, expected, **tol)
+
+        var = np.var(s1)
+        expected = (
```
johannesjmeyer: I would just hard-code the expected value here, having it calculated in-place adds no value because you did not parameterize the angles. This also relates to my above comment.

johannesjmeyer: This applies also to the other tests.

pennylane_qiskit/qiskit_device.py
```diff
@@ -458,6 +439,49 @@ def probabilities(self, wires=None):
...
+        Returns:
+            array[float]: an array of size ``(len(wires),)`` containing the
+            eigenvalues of the observable
+        """
+        # observable should be Z^{\otimes n}
+        eigvals = z_eigs(len(wires))
+
+        if isinstance(observable, list):
+            # determine the eigenvalues
+            if "Hermitian" in observable:
```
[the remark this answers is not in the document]
josh146: Yep, very good point! It did cross my mind, but at some point I must have implicitly assumed no user would ever do that :laughing:

tests/test_sample.py
josh146: Good point, I will modify `conftest.py` so that the 'stochastic' tolerance is shot dependent.

As a side note, I feel this is more of an integration test, as I wrote it so that it could be evaluated on hardware to ensure that the classical post-processing is doing what we would expect

tests/test_sample.py
josh146: I agree that it makes no difference re: hardcoding the values. At the same time, I would prefer to leave in the full expression for future use, i.e., in case the test does get parametrized, if the precision needs to be adjusted. Also it avoids the expected result looking a bit like a 'magic number'.

tests/test_expval.py
```diff
@@ -178,3 +178,93 @@ def test_multi_mode_hermitian_expectation(self, device, shots, tol):
         )
 
         assert np.allclose(res, expected, **tol)
+
+
+@pytest.mark.parametrize("shots", [0, 8192])
+class TestTensorExpval:
+    """Test tensor expectation values"""
+
+    def test_paulix_pauliy(self, device, shots, tol):
```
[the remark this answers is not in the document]
josh146: Fair point. The reason I coded it this way was to match other plugins, which don't support `QubitStateVector`.

pennylane_qiskit/qiskit_device.py
```diff
@@ -58,7 +58,7 @@ def z_eigs(n):
         n (int): number of wires
```
co9olguy: suggest changing name to `pauli_eigs` to make implicit assumptions later on more clear

pennylane_qiskit/qiskit_device.py
```diff
@@ -458,6 +439,49 @@ def probabilities(self, wires=None):
...
+        eigvals = z_eigs(len(wires))
+
+        if isinstance(observable, list):
+            # determine the eigenvalues
+            if "Hermitian" in observable:
+                # observable is of the form Z^{\otimes a}\otimes H \otimes Z^{\otimes b}
+                eigvals = np.array([1])
+
+                for k, g in itertools.groupby(
+                    zip(observable, wires, par), lambda x: x[0] == "Hermitian"
```
co9olguy:
```suggestion
                    zip(observable, wires, par), lambda x: "Hermitian" if x[0] == "Hermitian" else "non-Hermitian"
```

pennylane_qiskit/qiskit_device.py
```diff
@@ -458,6 +439,49 @@ def probabilities(self, wires=None):
...
+                ):
+                    if k:
```
co9olguy:
```suggestion
                    if k == "Hermitian":
```

pennylane_qiskit/qiskit_device.py
```diff
@@ -458,6 +439,49 @@ def probabilities(self, wires=None):
...
+                        p = list(g)[0][2]
+                        Hkey = tuple(p[0].flatten().tolist())
```
co9olguy: Need to aggregate in case there are more than one hermitian matrix

pennylane_qiskit/qiskit_device.py
co9olguy: for clarity

pennylane_qiskit/qiskit_device.py
```diff
@@ -458,6 +439,49 @@ def probabilities(self, wires=None):
...
+                        eigvals = np.kron(eigvals, self._eigs[Hkey]["eigval"])
+                    else:
+                        n = len([w for sublist in list(zip(*g))[1] for w in sublist])
```
co9olguy: This line seems to just be counting number of wires in a group of consecutive non=Hermitian operators. I wonder if there is an easier way to write this?

tests/test_sample.py
```diff
@@ -171,11 +173,11 @@ def test_paulix_pauliy(self, device, shots, tol):
         dev.apply("CNOT", wires=[1, 2], par=[])
 
         dev._obs_queue = [
-            qml.PauliX(wires=[0], do_queue=False) @ qml.PauliY(wires=[2], do_queue=False)
+            Tensor(["PauliX", "PauliY"], [[0], [2]], [[], []], qml.operation.Sample)
         ]
 
-        for idx in range(len(dev._obs_queue)):
-            dev._obs_queue[idx].return_type = qml.operation.Sample
```
co9olguy: How come these were removed? Why not just delete?

tests/test_sample.py
```diff
@@ -213,13 +215,11 @@ def test_pauliz_hadamard(self, device, shots, tol):
...
-            qml.PauliZ(wires=[0], do_queue=False)
-            @ qml.Hadamard(wires=[1], do_queue=False)
-            @ qml.PauliY(wires=[2], do_queue=False)
+            Tensor(["PauliZ", "Hadamard", "PauliY"], [[0], [1], [2]], [[], [], []], qml.operation.Sample)
         ]
 
-        for idx in range(len(dev._obs_queue)):
-            dev._obs_queue[idx].return_type = qml.operation.Sample
+        # for idx in range(len(dev._obs_queue)):
+        #     dev._obs_queue[idx].return_type = qml.operation.Sample
```
co9olguy: ditto

tests/test_sample.py
```diff
@@ -263,12 +263,10 @@ def test_hermitian(self, device, shots, tol):
...
 
-        dev._obs_queue = [
-            qml.PauliZ(wires=[0], do_queue=False) @ qml.Hermitian(A, wires=[1, 2], do_queue=False)
-        ]
+        dev._obs_queue = [Tensor(["PauliZ", "Hermitian"], [[0], [1, 2]], [[], [A]], qml.operation.Sample)]
 
-        for idx in range(len(dev._obs_queue)):
-            dev._obs_queue[idx].return_type = qml.operation.Sample
+        # for idx in range(len(dev._obs_queue)):
+        #     dev._obs_queue[idx].return_type = qml.operation.Sample
```
co9olguy: see above

tests/test_var.py
```diff
@@ -60,3 +60,132 @@ def test_var_hermitian(self, device, shots, tol):
...
+                + np.sin(phi)
+                * (
+                    15
+                    + 8 * np.cos(varphi)
+                    - 11 * np.cos(2 * varphi)
+                    + 42 * np.sin(varphi)
+                    + 3 * np.sin(2 * varphi)
+                )
+            )
+        ) / 16
```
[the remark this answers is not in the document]
josh146: lol, this one took me a long time to work out :laughing:

tests/test_expval.py
```diff
@@ -178,3 +178,93 @@ def test_multi_mode_hermitian_expectation(self, device, shots, tol):
...
+        dev.apply("RX", wires=[0], par=[theta])
+        dev.apply("RX", wires=[1], par=[phi])
+        dev.apply("RX", wires=[2], par=[varphi])
+        dev.apply("CNOT", wires=[0, 1], par=[])
+        dev.apply("CNOT", wires=[1, 2], par=[])
+
+        dev._obs_queue = [
+            qml.PauliX(wires=[0], do_queue=False) @ qml.PauliY(wires=[2], do_queue=False)
+        ]
+        res = dev.pre_measure()
```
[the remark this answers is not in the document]
josh146: removed

tests/test_expval.py
```diff
@@ -178,3 +178,93 @@ def test_multi_mode_hermitian_expectation(self, device, shots, tol):
...
+
+        res = dev.expval(["PauliX", "PauliY"], [[0], [2]], [[], [], []])
+        expected = np.sin(theta) * np.sin(phi) * np.sin(varphi)
+
+        assert np.allclose(res, expected, **tol)
```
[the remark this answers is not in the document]
josh146: done!

pennylane_qiskit/qiskit_device.py
```diff
@@ -119,7 +119,7 @@ class QiskitDevice(Device, abc.ABC):
     plugin_version = __version__
     author = "Carsten Blank"
```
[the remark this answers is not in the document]
josh146: How about just making it Xanadu?

pennylane_qiskit/qiskit_device.py
```diff
@@ -458,6 +439,49 @@ def probabilities(self, wires=None):
...
+                        Hkey = tuple(p[0].flatten().tolist())
+                        eigvals = np.kron(eigvals, self._eigs[Hkey]["eigval"])
+                    else:
+                        n = len([w for sublist in list(zip(*g))[1] for w in sublist])
+                        eigvals = np.kron(eigvals, z_eigs(n))
+
+        elif observable == "Hermitian":
+            # single wire Hermitian observable
+            Hkey = tuple(par[0].flatten().tolist())
+            eigvals = self._eigs[Hkey]["eigval"]
```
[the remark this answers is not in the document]
josh146: It should be safe, in that the existence of a `qml.Hermitian` in the observable queue causes the key to be created prior to measurement; the same observable queue is used later for extraction.

I can't think of an edge case where the observable queue could mutate between `pre_measure()` and `expval`, and an exception will be raised if something does go horribly wrong.

pennylane_qiskit/qiskit_device.py
josh146: done

pennylane_qiskit/qiskit_device.py
josh146: Had a long think about it, but couldn't come up with anything better. Open to suggestions!

tests/test_sample.py
josh146: This is because I was making a PL core tensor observable assumption; by removing the dependence on the PL core tensor branch, these are no longer needed.

tests/test_sample.py
josh146: Oh I think I left it commented just in case PL core requires this logic again :laughing:

tests/test_expval.py
```diff
@@ -266,3 +291,75 @@ def test_hermitian(self, device, shots, tol):
         )
 
         assert np.allclose(res, expected, **tol)
+
+    def test_hermitian_hermitian(self, device, shots, tol):
```
co9olguy: :smile:

pennylane_qiskit/qiskit_device.py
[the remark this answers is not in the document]
co9olguy: sg

PR #51, opened 2019-09-13
Fixes a bug where `backend_options` would be passed to backends that don't support it.

josh146:
**Context**: The current `QiskitDevice` assumes that all backends accept backend options, but this isn't the case --- hardware backends on IBMQ do not accept backend options. This leads to an `TypeError` for hardware-bound jobs.

**Description of change:**

* `inspect.signature` is used to determine what keyword arguments the backend supports.

* A new attribute `self.run_args` is now used to correctly store a dictionary of allowed keyword arguments.

* Version number incrementation for a minor bugfix release.

**Benefits:** Fixes the bug

**GitHub Issues**: https://discuss.pennylane.ai/t/pennylane-qiskit-error/204

### DIFF
--- a/pennylane_qiskit/aer.py
+++ b/pennylane_qiskit/aer.py
@@ -54,7 +54,6 @@ class AerDevice(QiskitDevice):
     Args:
         wires (int): The number of qubits of the device
         backend (str): the desired backend
-        noise_model (NoiseModel): NoiseModel Object from ``qiskit.providers.aer.noise``
         shots (int): number of circuit evaluations/random samples used
             to estimate expectation values and variances of observables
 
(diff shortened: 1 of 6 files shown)


tests/test_integration.py
```diff
@@ -66,3 +66,33 @@ def circuit(x, y, z):
...
+        assert dev.compile_backend == "test value"
+
+    def test_noise_model(self):
+        """Test that the noise model argument is properly
+        extracted if the backend supports it"""
+        dev = qml.device("qiskit.aer", wires=2, noise_model="test value")
+        assert dev.run_args["noise_model"] == "test value"
+
+    def test_invalid_noise_model(self):
+        """Test that the noise model argument is ignored
```
co9olguy: docstring doesn't quite match assertion

pennylane_qiskit/aer.py
```diff
@@ -54,30 +54,19 @@ class AerDevice(QiskitDevice):
...
 
     # pylint: disable=too-many-arguments
 
     short_name = "qiskit.aer"
 
-    def __init__(self, wires, shots=1024, backend="qasm_simulator", noise_model=None, **kwargs):
-        super().__init__(wires, qiskit.Aer, backend=backend, shots=shots, **kwargs)
-        self._noise_model = noise_model
-
-    def run(self, qobj):
```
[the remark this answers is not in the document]
josh146: I noticed that this was part of the same bug. There were no tests for noise models, so it wasn't picked up. Since the logic for kwargs is now part of the base class, this method is redundant and can be removed.

tests/test_integration.py
josh146: Oops, good catch. I made a new commit to make the logic cleaner, and forgot to update the docstring.
```suggestion
        """Test that the noise model argument causes an exception to be raised
```

PR #53, opened 2019-10-18
Update tests, version bump

josh146:
*Description of changes:*

* All Qiskit devices now support tensor observables using the
  `return expval(qml.PauliZ(0) @ qml.Hermitian(A, [1, 2]))`
  syntax introduced in PennyLane v0.6.

* Verified to work with Qiskit version 0.13

* Version bump

### DIFF
--- a/tests/test_sample.py
+++ b/tests/test_sample.py
@@ -5,7 +5,7 @@ import pennylane as qml
 
 from pennylane_qiskit import AerDevice, BasicAerDevice
 
-from conftest import U, U2, A, Tensor
+from conftest import U, U2, A
 
 
 np.random.seed(42)
@@ -134,11 +134,11 @@ class TestTensorSample:
         dev.apply("CNOT", wires=[1, 2], par=[])
 
         dev._obs_queue = [
-            Tensor(["PauliX", "PauliY"], [[0], [2]], [[], []], qml.operation.Sample)
+            qml.PauliX(0, do_queue=False) @ qml.PauliY(2, do_queue=False)
         ]
 
-        # for idx in range(len(dev._obs_queue)):
-        #     dev._obs_queue[idx].return_type = qml.operation.Sample
+        for idx in range(len(dev._obs_queue)):
+            dev._obs_queue[idx].return_type = qml.operation.Sample
 
         res = dev.pre_measure()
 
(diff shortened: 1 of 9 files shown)


tests/test_sample.py
```diff
@@ -216,10 +216,12 @@ def test_hermitian(self, theta, phi, varphi, device, shots, tol):
...
 
-        dev._obs_queue = [Tensor(["PauliZ", "Hermitian"], [[0], [1, 2]], [[], [A]], qml.operation.Sample)]
+        dev._obs_queue = [
+            qml.PauliZ(0, do_queue=False) @ qml.Hermitian(A, [1, 2], do_queue=False)
+        ]
 
-        # for idx in range(len(dev._obs_queue)):
-        #     dev._obs_queue[idx].return_type = qml.operation.Sample
+        for idx in range(len(dev._obs_queue)):
+            dev._obs_queue[idx].return_type = qml.operation.Sample
```
antalszava: Same comment as before

tests/test_sample.py
```diff
@@ -134,11 +134,11 @@ def test_paulix_pauliy(self, theta, phi, varphi, device, shots, tol):
...
 
         dev._obs_queue = [
-            Tensor(["PauliX", "PauliY"], [[0], [2]], [[], []], qml.operation.Sample)
+            qml.PauliX(0, do_queue=False) @ qml.PauliY(2, do_queue=False)
         ]
 
-        # for idx in range(len(dev._obs_queue)):
-        #     dev._obs_queue[idx].return_type = qml.operation.Sample
+        for idx in range(len(dev._obs_queue)):
+            dev._obs_queue[idx].return_type = qml.operation.Sample
```
antalszava: These lines do not seem to affect the outcome of the test, is there a particular reason to include them?

tests/test_sample.py
```diff
@@ -172,11 +172,11 @@ def test_pauliz_hadamard(self, theta, phi, varphi, device, shots, tol):
...
 
         dev._obs_queue = [
-            Tensor(["PauliZ", "Hadamard", "PauliY"], [[0], [1], [2]], [[], [], []], qml.operation.Sample)
+            qml.PauliZ(0, do_queue=False) @ qml.Hadamard(1, do_queue=False) @ qml.PauliY(2, do_queue=False)
         ]
 
-        # for idx in range(len(dev._obs_queue)):
-        #     dev._obs_queue[idx].return_type = qml.operation.Sample
+        for idx in range(len(dev._obs_queue)):
+            dev._obs_queue[idx].return_type = qml.operation.Sample
```
antalszava: Same comment as previously

tests/test_sample.py
```diff
@@ -134,11 +134,11 @@ def test_paulix_pauliy(self, theta, phi, varphi, device, shots, tol):
...
-            Tensor(["PauliX", "PauliY"], [[0], [2]], [[], []], qml.operation.Sample)
+            qml.PauliX(0, do_queue=False) @ qml.PauliY(2, do_queue=False)
         ]
 
-        # for idx in range(len(dev._obs_queue)):
-        #     dev._obs_queue[idx].return_type = qml.operation.Sample
+        for idx in range(len(dev._obs_queue)):
+            dev._obs_queue[idx].return_type = qml.operation.Sample
 
         res = dev.pre_measure()
```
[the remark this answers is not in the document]
josh146: FIxed!

tests/test_sample.py
josh146: Yep - since we are not using the high-level `sample()` function, the sample return type on the observables is not being set. Without this being set, the plugin will not instruct Qiskit to return the samples (this is optional in qiskit).

PR #55, opened 2019-10-31
QuantumCircuit to PennyLane template conversion

antalszava:
**Context:**
Each quantum software framework has its architecture when it comes to how they handle creating circuits and applying gates. Therefore a converter between Qiskit's ``QuantumCircuit`` and a custom template in PennyLane could be provided. This converter could then be utilized by users who are new to PennyLane but have prior experience with other quantum computing software frameworks.

Adds the converter function ``load`` in ``converter.py`` that creates a PennyLane template from an arbitrary
* ``QuantumCircuit`` object;
* ``QASM`` file;
* ``QASM`` string.

Adds the entrypoint ``pennylane.io`` such that PennyLane can use the ``load`` function.

**Benefits:**

Allows the conversion from Qiskit circuit object to Pennylane templates. Once the conversion has been done, these templates can then be placed within a quantum circuit in PennyLane.

Circuits containing parameters converted this way can be parametrized by passing a dictionary as an argument to ``load``.

Each converted circuit can be added to a quantum circuit in PennyLane an arbitrary many times. The user can also specify different parameters each time if desired.

Further allows shifting the converted circuit. Passing the ``wire_shift`` argument allows to shift the indices of wires already defined in the original ``QuantumCircuit`` object e.g. if the ``QuantumCircuit`` acts on wires `[0, 1]` and if `wire_shift == 1`, then the returned Pennylane template will act on wires `[1, 2]`.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,9 @@
 # Release 0.7.0-dev
 
 ### New features since last release
-
-* Added a `QuantumCircuit`/QASM to PennyLane template converter
-  that allows incorporating Qiskit circuits in a PennyLane quantum
-  circuit. 
+* Added the ability to automatically convert Qiskit `QuantumCircuits`
+  or QASM circuits directly into PennyLane templates. The loaded
+  operations can be used directly inside PennyLane circuits.
   [#55](https://github.com/XanaduAI/pennylane-qiskit/pull/55)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ### New features since last release
 
+* Added a 'QuantumCircuit'/'QASM' to PennyLane template converter
+  that allows incorporating Qiskit circuits in a PennyLane quantum
+  circuit. 
+  [#55](https://github.com/XanaduAI/pennylane-qiskit/pull/55)
+
+
 ### Breaking changes
 
 ### Improvements
@@ -13,6 +19,8 @@
 ### Contributors
 
 This release contains contributions from (in alphabetical order):
+
+Antal Száva
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### New features since last release
 
-* Added a 'QuantumCircuit'/'QASM' to PennyLane template converter
+* Added a `QuantumCircuit`/QASM to PennyLane template converter
   that allows incorporating Qiskit circuits in a PennyLane quantum
   circuit. 
   [#55](https://github.com/XanaduAI/pennylane-qiskit/pull/55)
--- a/pennylane_qiskit/__init__.py
+++ b/pennylane_qiskit/__init__.py
@@ -14,10 +14,8 @@
 """Top level PennyLane-qiskit module"""
 
 from ._version import __version__
-
 from .aer import AerDevice
 from .basic_aer import BasicAerDevice
-from .ibmq import IBMQDevice
-
-from .ops import *
 from .converter import load
+from .ibmq import IBMQDevice
+from .ops import *
--- a/pennylane_qiskit/__init__.py
+++ b/pennylane_qiskit/__init__.py
@@ -16,6 +16,6 @@
 from ._version import __version__
 from .aer import AerDevice
 from .basic_aer import BasicAerDevice
-from .converter import load
+from .converter import load, load_qasm_from_file, load_qasm
 from .ibmq import IBMQDevice
 from .ops import *
--- a/pennylane_qiskit/__init__.py
+++ b/pennylane_qiskit/__init__.py
@@ -16,6 +16,6 @@
 from ._version import __version__
 from .aer import AerDevice
 from .basic_aer import BasicAerDevice
-from .converter import load, load_qasm
+from .converter import load, load_qasm, load_qasm_from_file
 from .ibmq import IBMQDevice
 from .ops import *
--- a/pennylane_qiskit/__init__.py
+++ b/pennylane_qiskit/__init__.py
@@ -16,6 +16,6 @@
 from ._version import __version__
 from .aer import AerDevice
 from .basic_aer import BasicAerDevice
-from .converter import load, load_qasm_from_file, load_qasm
+from .converter import load, load_qasm
 from .ibmq import IBMQDevice
 from .ops import *
--- a/pennylane_qiskit/__init__.py
+++ b/pennylane_qiskit/__init__.py
@@ -20,3 +20,4 @@ from .basic_aer import BasicAerDevice
 from .ibmq import IBMQDevice
 
 from .ops import *
+from .converter import load
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -36,7 +36,7 @@ from pennylane_qiskit.qiskit_device import QISKIT_OPERATION_MAP
 inv_map = {v.__name__: k for k, v in QISKIT_OPERATION_MAP.items()}
 
 
-def check_parameter_bound(param):
+def _check_parameter_bound(param):
     """Utility function determining if a certain parameter in a QuantumCircuit has
     been bound.
 
@@ -51,7 +51,7 @@ def check_parameter_bound(param):
     return param
 
 
-def check_circuit_and_bind_parameters(quantum_circuit: QuantumCircuit, params: dict) -> QuantumCircuit:
+def _check_circuit_and_bind_parameters(quantum_circuit: QuantumCircuit, params: dict) -> QuantumCircuit:
     """Utility function for checking for a valid quantum circuit and then binding parameters.
 
     Args:
@@ -105,7 +105,7 @@ def execute_supported_operation(operation_name: str, parameters: list, wires: li
     """
     operation = getattr(pennylane_ops, operation_name)
 
-    parameters = [check_parameter_bound(param) for param in parameters]
+    parameters = [_check_parameter_bound(param) for param in parameters]
 
     if not parameters:
         operation(wires=wires)
@@ -145,7 +145,7 @@ def load(quantum_circuit: QuantumCircuit):
             function: the new PennyLane template
         """
 
-        qc = check_circuit_and_bind_parameters(quantum_circuit, params)
+        qc = _check_circuit_and_bind_parameters(quantum_circuit, params)
 
         # Wires from a qiskit circuit are unique w.r.t. a register name and a qubit index
         qc_wires = [(q.register.name, q.index) for q in quantum_circuit.qubits]
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -59,7 +59,7 @@ def check_circuit_and_bind_parameters(quantum_circuit: QuantumCircuit, params: d
         params (dict): dictionary of the parameters in the circuit
 
     Returns:
-        qc: quantum circuit with bound parameters
+        qc (QuantumCircuit): quantum circuit with bound parameters
     """
     if not isinstance(quantum_circuit, QuantumCircuit):
         raise ValueError("The circuit {} is not a valid Qiskit QuantumCircuit.".format(quantum_circuit))
@@ -79,12 +79,11 @@ def map_wires(wires: list, qc_wires: list) -> dict:
     specified by the user for the template.
 
     Args:
-        wires: wires specified for the template
-        qc_wires: wires from the converted quantum circuit
+        wires (list): wires specified for the template
+        qc_wires (list): wires from the converted quantum circuit
 
     Returns:
-        wire_map: a dictionary that contains maps from quantum circuit wires to the user
-            defined wires
+        wire_map (dict): map from quantum circuit wires to the user defined wires
     """
     if wires is None:
         return dict(zip(qc_wires, range(len(qc_wires))))
@@ -100,9 +99,9 @@ def execute_supported_operation(operation_name: str, parameters: list, wires: li
     """Utility function that executes an operation that is natively supported by PennyLane.
 
     Args:
-        operation_name: wires specified for the template
-        parameters: parameters of the operation that will be executed
-        wires: wires of the operation
+        operation_name (str): wires specified for the template
+        parameters (str): parameters of the operation that will be executed
+        wires (list): wires of the operation
     """
     operation = getattr(pennylane_ops, operation_name)
 
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -28,13 +28,16 @@ Classes
 Code details
 ~~~~~~~~~~~~
 """
-from qiskit import QuantumCircuit
-from qiskit.circuit import Parameter, ParameterExpression
-from pennylane_qiskit.qiskit_device import QISKIT_OPERATION_MAP
-import pennylane.ops.qubit as pennylane_ops
-import pennylane_qiskit.ops as plugin_ops
 import warnings
 import numpy as np
+
+from qiskit import QuantumCircuit
+from qiskit.circuit import Parameter
+from qiskit.exceptions import QiskitError
+
+import pennylane.ops.qubit as pennylane_ops
+import pennylane_qiskit.ops as plugin_ops
+from pennylane_qiskit.qiskit_device import QISKIT_OPERATION_MAP
 
 # pylint: disable=too-many-instance-attributes
 
@@ -43,6 +46,15 @@ inv_map = {v.__name__: k for k, v in _operation_map.items()}
 
 
 def check_parameter_bound(param):
+    """Utility function determining if a certain parameter in a QuantumCircuit has
+    been bound.
+
+    Args:
+        param: the parameter to be checked
+
+    Returns:
+        param: the parameter after the check
+    """
     if isinstance(param, Parameter):
         raise ValueError("The parameter {} was not bound correctly.".format(param))
     return param
@@ -96,17 +108,11 @@ def load(quantum_circuit: QuantumCircuit):
             wires = [wire_shift + qubit.index for qubit in op[1]]
             instruction_name = op[0].__class__.__name__
 
-            print(plugin_ops)
-
-            if instruction_name in inv_map and\
-                    (inv_map[instruction_name] in pennylane_ops.ops or inv_map[instruction_name] in plugin_ops._ops):
+            if instruction_name in inv_map and inv_map[instruction_name] in pennylane_ops.ops:
 
                 operation_name = inv_map[instruction_name]
 
-                if operation_name in pennylane_ops.ops:
-                    operation = getattr(pennylane_ops, operation_name)
-                else:
-                    operation = getattr(plugin_ops, operation_name)
+                operation = getattr(pennylane_ops, operation_name)
 
                 parameters = [check_parameter_bound(param) for param in op[0].params]
 
@@ -123,15 +129,35 @@ def load(quantum_circuit: QuantumCircuit):
                     operation(*float_params, wires=wires)
 
             else:
-                warnings.warn(__name__ + " The {} instruction is not supported by PennyLane.".
-                              format(instruction_name),
-                              UserWarning)
+                try:
+                    operation_matrix = op[0].to_matrix()
+                    pennylane_ops.QubitUnitary(operation_matrix, wires=wires)
+                except (AttributeError, QiskitError):
+                    warnings.warn(__name__ + " The {} instruction is not supported by PennyLane.".
+                                  format(instruction_name),
+                                  UserWarning)
     return _function
 
 
 def load_qasm_from_file(file: str):
+    """Returns a PennyLane template created based on the input qasm file.
+
+        Args:
+            file (str): the name of the file
+
+        Returns:
+            function: the new PennyLane template
+    """
     return load(QuantumCircuit.from_qasm_file(file))
 
 
-def load_qasm(string: str):
-    return load(QuantumCircuit.from_qasm_str(string))
+def load_qasm(qasm_string: str):
+    """Returns a PennyLane template created based on the input qasm string.
+
+        Args:
+            qasm_string (str): the name of the qasm string
+
+        Returns:
+            function: the new PennyLane template
+        """
+    return load(QuantumCircuit.from_qasm_str(qasm_string))
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -136,7 +136,7 @@ def load(quantum_circuit: QuantumCircuit):
         not incorporated in the PennyLane template.
 
         Args:
-            params (dict): specifies the parameters that need to bound in the QuantumCircuit
+            params (dict): specifies the parameters that need to be bound in the QuantumCircuit
             wires (Sequence[int] or int): The wires the converted template acts on.
                 Note that if the original QuantumCircuit acted on :math:`N` qubits,
                 then this must be a list of length :math:`N`.
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -17,29 +17,23 @@ QuanctumCircuit converter module
 
 .. currentmodule:: pennylane_qiskit.load
 
-This module contains functions for converting Qiskit QuantumCirducit objects
-PennyLane templates.
+This module contains functions for converting Qiskit QuantumCircuit objects
+into PennyLane circuit templates.
 """
 import warnings
 
 import numpy as np
-
 from qiskit import QuantumCircuit
 from qiskit.circuit import Parameter
 from qiskit.exceptions import QiskitError
 
+import pennylane as qml
 import pennylane.ops.qubit as pennylane_ops
 from pennylane_qiskit.qiskit_device import QISKIT_OPERATION_MAP
 
 # pylint: disable=too-many-instance-attributes
 
 inv_map = {v.__name__: k for k, v in QISKIT_OPERATION_MAP.items()}
-
-
-class WiresError(Exception):
-    """Exception raised by  the pennylane_qiskit.converter when it encounters an illegal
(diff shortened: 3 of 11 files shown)


tests/test_converter.py
```diff
@@ -0,0 +1,587 @@
...
+
+        assert len(recorder.queue) == 2
+
+        assert recorder.queue[0].name == 'PauliX'
+        assert recorder.queue[0].params == []
+        assert recorder.queue[0].wires == [0]
+
+        assert recorder.queue[1].name == 'PauliX'
+        assert recorder.queue[1].params == []
+        assert recorder.queue[1].wires == [2]
```
josh146:
```suggestion
        assert recorder.queue[1].wires == [2]

```

josh146:
By the way, it looks like the tests are failing on Python 3.5 - could be a dictionary ordering issue

pennylane_qiskit/__init__.py
```diff
@@ -20,3 +20,4 @@
 from .ibmq import IBMQDevice
 
 from .ops import *
+from .converter import load
```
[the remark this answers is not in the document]
antalszava: Great, thanks for the tip on the tool!

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,154 @@
...
+
+    Args:
+        quantum_circuit (qiskit.QuantumCircuit): the QuantumCircuit to be converted
+
+    Returns:
+        function: the new PennyLane template
+    """
+
+    if isinstance(quantum_circuit, str):
+        quantum_circuit = QuantumCircuit._circuit_from_qasm(quantum_circuit)
```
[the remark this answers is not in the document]
antalszava: Yeah, quite nice to be able to use that one!

tests/test_converter.py
```diff
@@ -0,0 +1,587 @@
...
+
+        quantum_circuit = load(qc)
+
+        with recorder:
+            quantum_circuit()
+
+        assert len(recorder.queue) == 1
+        assert recorder.queue[0].name == 'RZ'
+        assert recorder.queue[0].params == [angle]
+        assert recorder.queue[0].wires == [0]
```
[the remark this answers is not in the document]
antalszava: That's true, it was a great job coming up with that! ;)

tests/test_converter.py
```diff
@@ -0,0 +1,587 @@
...
+
+        with recorder:
+            quantum_circuit()
+
+        assert len(recorder.queue) == 1
+        assert recorder.queue[0].name == 'RZ'
+        assert recorder.queue[0].params == [0.5]
+        assert recorder.queue[0].wires == [0]
+
+    def test_converter_pass_parameters_to_bind(self, recorder):
```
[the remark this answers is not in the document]
antalszava: That's a great one, thanks! Modified it

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,154 @@
...
+                else:
+                    float_params = [float(param) for param in parameters]
+                    operation(*float_params, wires=wires)
+
+            else:
+                try:
+                    operation_matrix = op[0].to_matrix()
+                    pennylane_ops.QubitUnitary(operation_matrix, wires=wires)
+                except (AttributeError, QiskitError):
+                    warnings.warn(__name__ + " The {} instruction is not supported by PennyLane.".
```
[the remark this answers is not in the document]
antalszava: Sounds good, modified it!

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,154 @@
...
+    def _function(params: dict = None, wire_shift: int = 0):
+        """Returns a PennyLane template created based on the input QuantumCircuit.
+            Warnings are created for each of the QuantumCircuit instructions that were
+            not incorporated in the PennyLane template.
+
+            Args:
+                params (dict): specifies the parameters that need to bound in the
+                QuantumCircuit
+
+                wire_shift (int): the shift to be made with respect to the wires
```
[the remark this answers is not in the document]
antalszava: Yes, true! I would guess that it would be worth adding that later, as you suggested.

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,154 @@
...
+                already specifiec in the QuantumCircuit
+                e.g. if the QuantumCircuit acts on wires [0, 1] and if wire_shift == 1,
+                then the returned Pennylane template will act on wires [1, 2]
```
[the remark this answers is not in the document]
antalszava: Currently, this does not appear in the docs. However, once the syntax is fixed in PennyLane, I'll add a small snippet of how to use the load from a user perspective.

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,150 @@
...
+        not incorporated in the PennyLane template.
+
+        Args:
+            params (dict): specifies the parameters that need to bound in the
+            QuantumCircuit
+
+            wire_shift (int): the shift to be made with respect to the wires
+            already specifiec in the QuantumCircuit
+            e.g. if the QuantumCircuit acts on wires [0, 1] and if wire_shift == 1,
+            then the returned Pennylane template will act on wires [1, 2]
```
josh146: Thinking about this more, this should be relatively easy to port it to a `wires=` like approach.

For example, if it's a 3 qubit quantum circuit, and the user converts it into a template with `wires=[2,0,4]`, then:
```python
qc_wires = [q.index for q in quantum_circuit.qubits]
wire_map = {k:v for k, v in zip(qc_wires, wires)}
```

and then, further down below it becomes
```python
operation(wires=[wire_map[w] for w in wires])
```

Is this a small enough change to make here now?

antalszava: Thanks for that @josh146! I pushed a modification that includes this as well with new tests.

This made me realize, that ``qubit.index`` might not uniquely identify the wires operations act on. The following is an example for this:
```python
qr1 = QuantumRegister(2)
qr2 = QuantumRegister(1)
qc = QuantumCircuit(qr1, qr2)
wires = [q.index for q in circuit.qubits]
```
In this example, ``wires == [0, 1, 0, 1]``, therefore the ``(qubit.register.name, qubit.index)`` tuple is used to distinguish between wires.

Modified the code according to this as well.

antalszava:
The reason why there's a test failing for Python 3.5 is that the test called ``test_qasm_from_file`` which uses the following fixture:
```python
@pytest.mark.skipif(sys.version_info < (3, 6), reason="tmpdir fixture requires Python >=3.6")
```
This one is used for creating a temporary ``qasm`` file.

tests/test_converter.py
```diff
@@ -0,0 +1,721 @@
...
+        with pytest.warns(UserWarning) as record:
+            with recorder:
+                quantum_circuit(params={})
+
+        # check that only one warning was raised
+        assert len(record) == 1
+        # check that the message matches
+        assert record[0].message.args[0] == "pennylane_qiskit.converter The {} instruction is not supported by" \
+                                            " PennyLane, and has not been added to the template." \
+            .format('Barrier')
```
co9olguy: why put a string in `.format`?

tests/test_converter.py
```diff
@@ -0,0 +1,721 @@
...
+        quantum_circuit = load(qc)
+
+        with pytest.warns(UserWarning) as record:
+            with recorder:
+                quantum_circuit(params={})
+
+        # check that only one warning was raised
+        assert len(record) == 1
+        # check that the message matches
+        assert record[0].message.args[0] == "pennylane_qiskit.converter The {} instruction is not supported by" \
```
co9olguy: (throughout): Seems to be some punctuation missing in this msg

tests/test_converter.py
```diff
@@ -0,0 +1,721 @@
...
+
+        assert len(recorder.queue) == 2
+
+        assert recorder.queue[0].name == 'PauliX'
+        assert recorder.queue[0].params == []
+        assert recorder.queue[0].wires == [0]
+
+        assert recorder.queue[1].name == 'PauliX'
+        assert recorder.queue[1].params == []
+        assert recorder.queue[1].wires == [2]
```
co9olguy: Amazing work on the tests! 👍

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ### New features since last release
 
+* Added a `QuantumCircuit`/QASM to PennyLane template converter
```
co9olguy: "Added the ability to automatically convert Qiskit `QuantumCircuit`s or QASM circuits directly into PennyLane templates. The loaded operations can be used directly inside PennyLane circuits.

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+    if isinstance(param, Parameter):
+        raise ValueError("The parameter {} was not bound correctly.".format(param))
+    return param
+
+
+def load(quantum_circuit: QuantumCircuit):
+    """Returns a PennyLane template created based on the input QuantumCircuit or QASM string.
+    Warnings are created for each of the QuantumCircuit instructions that were
+    not incorporated in the PennyLane template.
```
co9olguy: In a future PR, maybe we can think about adding a keyword argument to suppress warnings?

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+
+    def _function(params: dict = None, wires: list = None):
+        """Returns a PennyLane template created based on the input QuantumCircuit.
+        Warnings are created for each of the QuantumCircuit instructions that were
+        not incorporated in the PennyLane template.
+
+        Args:
+            params (dict): specifies the parameters that need to bound in the
+            QuantumCircuit
```
co9olguy: (throughout): indentation?

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,167 @@
...
+
+            wire_shift (list): the shift to be made with respect to the wires
+            already specifiec in the QuantumCircuit
```
co9olguy:
```suggestion
            already specified in the QuantumCircuit
```

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,167 @@
...
+    def _function(params: dict = None, wires: list = None):
+        """Returns a PennyLane template created based on the input QuantumCircuit.
+        Warnings are created for each of the QuantumCircuit instructions that were
+        not incorporated in the PennyLane template.
+
+        Args:
+            params (dict): specifies the parameters that need to bound in the
+            QuantumCircuit
+
+            wire_shift (list): the shift to be made with respect to the wires
```
co9olguy: I think it is better to allow a permutation of wires, rather than just a shift. Makes things more flexible

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,167 @@
...
+                else:
+                    float_params = [float(param) for param in parameters]
+                    operation(*float_params, wires=operation_wires)
+
+            else:
+                try:
+                    operation_matrix = op[0].to_matrix()
+                    pennylane_ops.QubitUnitary(operation_matrix, wires=operation_wires)
+                except (AttributeError, QiskitError):
+                    warnings.warn(__name__ + " The {} instruction is not supported by PennyLane,"
```
co9olguy: Add some punctuation here, e.g.,
```
__name__ + ": The ..."
```

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,167 @@
...
+def load_qasm_from_file(file: str):
+    """Returns a PennyLane template created based on the input qasm file.
+
+        Args:
+            file (str): the name of the file
+
+        Returns:
+            function: the new PennyLane template
+    """
+    return load(QuantumCircuit.from_qasm_file(file))
```
co9olguy: clever 😺

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,167 @@
...
+
+            if instruction_name in inv_map and inv_map[instruction_name] in pennylane_ops.ops:
+
+                operation_name = inv_map[instruction_name]
+
+                operation = getattr(pennylane_ops, operation_name)
+
+                parameters = [check_parameter_bound(param) for param in op[0].params]
+
+                if not parameters:
```
co9olguy: Nothing to do with this PR, but this op-specific logic kind of bothers me.

tests/test_converter.py
co9olguy: Can we add a couple integration tests? I'm thinking that:
- the Recorder is not used (instead, we use a standard PL Qnode)
- the `assert` checks that the correct output is returned
- also, be good to check when the `load` function is used inline in the QNode (from what I could see, all tests above used `load` before creating the QNode

pennylane_qiskit/ops.py
```diff
@@ -52,6 +52,18 @@
 # pylint: disable=too-few-public-methods
 from pennylane.operation import Operation
 
+_ops = {
```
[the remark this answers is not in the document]
josh146: We should move all remaining operations over in a new PR, and then delete these.

The ones we still need to move over:
* `U2`
* `U3`
* `Toffoli`

Note that `U1` is already present in PennyLane as the `PhaseShift` gate.

pennylane_qiskit/converter.py
josh146: This has been added, but this docstring is out of date and needs to be supported 🙂 

You can now do something like this:
```python
qc = QuantumCircuit(3, 1)
theta = Parameter('theta')
qc.rz(theta, 0)
qc.cx(0, 1)
qiskit_template = load(qc)

@qml.qnode(dev)
def circuit(x):
    # apply template to wires (2, 0), i.e., CNOT_{2, 0} .  RZ(x)_2 |psi>
    qiskit_template(params={theta: x}, wires=[2, 0])
    return qml.expval(qml.PauliZ(2))
```
that is, it acts exactly like a PennyLane operation.

pennylane_qiskit/converter.py
josh146: We played around with it, but there was ambiguity because PennyLane and Qiskit treat operations with array input (i.e., `QubitUnitary` and `QubitStateVector`) differently.

Since all plugins currently do this as well (use conditional branches for these two operations), I'm okay with this here --- it doesn't look like there is a standard clean solution to this yet.

josh146:
By the way @antalszava, it looks like Codecov is failing because the lines 130-131 aren't being covered by the tests:
```python
else:
    float_params = [float(param) for param in parameters]
    operation(*float_params, wires=operation_wires)
```

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,167 @@
...
+        not incorporated in the PennyLane template.
+
+        Args:
+            params (dict): specifies the parameters that need to bound in the
+            QuantumCircuit
+
+            wire_shift (list): the shift to be made with respect to the wires
+            already specifiec in the QuantumCircuit
+            e.g. if the QuantumCircuit acts on wires [0, 1] and if wire_shift == 1,
+            then the returned Pennylane template will act on wires [1, 2]
```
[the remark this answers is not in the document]
antalszava: Great one, thanks for that!

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,167 @@
...
+
+# pylint: disable=too-many-instance-attributes
+
+inv_map = {v.__name__: k for k, v in QISKIT_OPERATION_MAP.items()}
+
+
+class WiresError(Exception):
+    """Exception raised by  the pennylane_qiskit.converter when it encounters an illegal
+    operation in the quantum circuit.
+    """
```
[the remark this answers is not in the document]
antalszava: The intention here was to raise a custom error specifically related to the number of wires. I do agree though, that ``qml.QuantumFunctionError`` can also be raised with a message describing the situation. It might be even better, because then each plugin can just raise this type of error, and they do not have to define a custom error type.

Modifying these, thanks!

pennylane_qiskit/ops.py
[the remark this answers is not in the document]
antalszava: Great, I'll open a new PR for this

tests/test_converter.py
```diff
@@ -0,0 +1,721 @@
...
+    def test_quantum_circuit_init_by_specifying_rotation_in_circuit(self, recorder):
+        """Tests the load method for a QuantumCircuit initialized using separately defined
+         quantum and classical registers."""
+
+        angle = 0.5
+
+        qr = QuantumRegister(1)
+        cr = ClassicalRegister(1)
+
+        # quantum circuit to make an entangled bell state
```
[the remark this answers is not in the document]
antalszava: Oh, thanks for that! The comment remained from a Bell state example

tests/test_converter.py
```diff
@@ -0,0 +1,721 @@
...
+        quantum_circuit = load(qc)
+        with recorder:
+            quantum_circuit(wires=three_wires)
+
+        assert recorder.queue[0].name == 'CSWAP'
+        assert recorder.queue[0].params == []
+        assert recorder.queue[0].wires == three_wires
+
+    def test_wires_pass_different_wires_than_for_circuit(self, recorder):
+        """Tests that the wires is correct even if the quantum circuit was initiliazed with two
```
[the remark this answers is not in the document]
antalszava: Yes, correcting it, thanks!

tests/test_converter.py
antalszava: Good point, modified it!

pennylane_qiskit/converter.py
antalszava: The `load` is a wrapper around `_function` here, would you suggest to have another indentation style in this case?

josh146: I think this is a reference to the `Args:`; the line continuation should be indented, and there is no need to have a line break; i.e.
```
        Args:
            params (dict): specifies the parameters that need to bound in the
                QuantumCircuit
            wire_shift (list): the shift to be made with respect to the wires
```

tests/test_converter.py
antalszava: @co9olguy this was a great idea, thank you!

I've added some integration tests and turned out that further checks were needed in `_function` for the case when parameters are of `Variable` type.

pennylane_qiskit/converter.py
antalszava: That would be reasonable, yes. The users could then toggle to see warnings while converting.

pennylane_qiskit/converter.py
antalszava: True, thanks!

antalszava:
Thank you for all the feedback, it was really valuable!

1. As far as the changes in PennyLane are concerned, I would envision something like the following in ``__init__.py`` (similar to how the device entrypoint is treated):
```python

def load(name: str, quantum_circuit_object):
    """load(name, quantum_circuit_object)
    Load a plugin :func:`~.load` which can then be used to convert objects of
    quantum circuits from other frameworks.

    Args:
        name (str): the name of the plugin to convert from
        quantum_circuit_object: the quantum circuit that will be converted
            to a PennyLane template

    Returns:
        _function: the PennyLane template created from the quantum circuit
            object
    """

    if name in plugin_converters:

        # loads the plugin load function
        plugin_converter = plugin_converters[name].load()

        # calls the load function of the converter on the quantum circuit object
        return plugin_converter(quantum_circuit_object)

    raise ValueError(
        "Converter does not exist. Make sure the required plugin is installed "
        "and supports conversion."
    )
```
Once this is defined, the following ``from_qasm`` can also be added

```python
def from_qasm(quantum_circuit: str):
    """load(name, quantum_circuit)
    Loads quantum circuits defined in QASM by using the converter in the
    PennyLane-Qiskit plugin.

    Args:
        quantum_circuit (str): the name for the QASM file or string

    Returns:
        _function: the PennyLane template created from the quantum circuit
            object
    """
    return load('qasm', quantum_circuit)
```
In the latest version, the entrypoints were extended with:
``'qasm = pennylane_qiskit:load_qasm',``

Therefore, each plugin will be available through entrypoints and by passing their string as the ``name`` argument to ``load``.

In order to make this seamless, I have created one single function for QASM and added a simple check.

Furthermore, the load function has been refactored, it has been split into smaller functions such that it is more readable.

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,194 @@
...
+        operation(np.array(parameters), wires=wires)
+    elif operation_name == 'QubitUnitary':
+        operation(*parameters, wires=wires)
+    else:
+        float_params = [float(param) for param in parameters]
+        operation(*float_params, wires=wires)
+
+
+def load(quantum_circuit: QuantumCircuit):
+    """Returns a PennyLane template created based on the input QuantumCircuit or QASM string.
```
josh146:
```suggestion
    """Loads a PennyLane template from a Qiskit QuantumCircuit.
```

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,194 @@
...
+    Warnings are created for each of the QuantumCircuit instructions that were
+    not incorporated in the PennyLane template.
+
+    Args:
+        quantum_circuit (qiskit.QuantumCircuit): the QuantumCircuit to be converted
+
+    Returns:
+        function: the new PennyLane template
```
josh146:
```suggestion
        function: the resulting PennyLane template
```

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,194 @@
...
+    """
+
+    def _function(params: dict = None, wires: list = None):
+        """Returns a PennyLane template created based on the input QuantumCircuit.
+        Warnings are created for each of the QuantumCircuit instructions that were
+        not incorporated in the PennyLane template.
+
+        Args:
+            params (dict): specifies the parameters that need to bound in the QuantumCircuit
+
```
josh146:
```suggestion
```

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,194 @@
...
+                    warnings.warn(__name__ + ": The {} instruction is not supported by PennyLane,"
+                                             " and has not been added to the template.".
+                                  format(instruction_name),
+                                  UserWarning)
+
+    return _function
+
+
+def load_qasm(qasm_string: str):
+    """Returns a PennyLane template created based on the input QASM string or file.
```
josh146:
```suggestion
    """Loads a PennyLane template from a QASM string.
```

pennylane_qiskit/converter.py
```diff
@@ -0,0 +1,194 @@
...
+
+        Returns:
+            function: the new PennyLane template
+    """
+
+    # Checks whether a qasm string or a file name was provided
+    if isinstance(qasm_string, str) and not qasm_string.endswith('.qasm'):
+        return load(QuantumCircuit.from_qasm_str(qasm_string))
+
+    return load(QuantumCircuit.from_qasm_file(qasm_string))
```
josh146: It might be best to separate this into two functions; `load_qasm_string` and `load_qasm_file`, just to be explicit. This avoids implicitly checking `qasm_string.endswith('.qasm')`, in case a user has saved the file with a different extension.

antalszava: That's true, there is actually no explicit requirement for the file to end with the string ".qasm". Thanks!

tests/test_converter.py
```diff
@@ -0,0 +1,876 @@
...
+            load(qc)({theta: angle, phi: rotation_angle2})
+            return qml.expval(qml.PauliZ(0))
+
+        @qml.qnode(qubit_device_2_wires)
+        def circuit_native_pennylane(angle):
+            qml.RZ(angle, wires=0)
+            qml.RX(rotation_angle2, wires=0)
+            return qml.expval(qml.PauliZ(0))
+
+        assert circuit_loaded_qiskit_circuit(rotation_angle1) ==\
```
co9olguy:
```suggestion
        assert circuit_loaded_qiskit_circuit(rotation_angle1) == \
```

pennylane_qiskit/converter.py
```diff
@@ -57,16 +51,83 @@ def check_parameter_bound(param):
     return param
 
 
+def check_circuit_and_bind_parameters(quantum_circuit: QuantumCircuit, params: dict) -> QuantumCircuit:
```
co9olguy: make private?

pennylane_qiskit/converter.py
```diff
@@ -57,16 +51,83 @@ def check_parameter_bound(param):
...
+        return quantum_circuit
+
+    for k, v in params.items():
+        if isinstance(v, qml.variable.Variable):
+            params.update({k: v.val})
+
+    return quantum_circuit.bind_parameters(params)
+
+
+def map_wires(wires: list, qc_wires: list) -> dict:
```
co9olguy: Is this tested anywhere?

pennylane_qiskit/ops.py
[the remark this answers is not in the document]
antalszava: Yep, thanks for the reminder! This change was done in PL core by Josh, but it's still a todo here.

tests/test_converter.py
```diff
@@ -716,6 +753,124 @@ def test_qasm_(self, recorder):
...
+            return qml.expval(qml.PauliZ(0))
+
+        @qml.qnode(qubit_device_2_wires)
+        def circuit_native_pennylane():
+            qml.RZ(angle, wires=0)
+            return qml.expval(qml.PauliZ(0))
+
+        assert circuit_loaded_qiskit_circuit() == circuit_native_pennylane()
+
+    def test_circuit_loaded_in_qnode(self, qubit_device_2_wires):
```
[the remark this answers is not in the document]
antalszava: Changed them, thanks!

pennylane_qiskit/converter.py
antalszava: Done!

pennylane_qiskit/converter.py
antalszava: Added some tests for this, thanks!

Issue #56, opened 2019-11-11
Update the list of custom operations


antalszava:
The list of supported operations has been updated in Pennylane core. Now gates like the U2, U3 and Toffoli gates are among the operations.

Previously, in this plugin, these were added separately in the ``ops.py`` file. It would be nice to update the list of operations in PennyLane-Qiskit as well.

antalszava:
Done with #58, closing this for now

PR #58, opened 2019-11-19
Update operations list and add the inverses of operations

antalszava:
Update the operations list and add the inverses of operations. There were three main points to this PR:

- Remove the ``ops.py`` file as all the contained operations are now in PennyLane core
- Add and implement the capability for PennyLane-Qiskit to be able to process inverse operations and include tests to the inverse
- Change the operations list such that specific inverses (e.g. ``Sdg``) are no longer members

It is worth noting that for the inverse to work, a separate dictionary was created and later merged with ``QISKIT_OPERATION_MAP`` to initialize the ``device._operation_map``. This approach was needed such the the ``QISKIT_OPERATION_MAP`` dictionary remains invertable (a characteristic used by the ``load`` function for conversion).

This PR follows up on issue #56.


pennylane_qiskit/qiskit_device.py
```diff
@@ -82,29 +82,26 @@ def pauli_eigs(n):
     "RZ": ex.RZGate,
     "S": ex.SGate,
     "T": ex.TGate,
+
+    # Adding the following for conversion compatibility
+    "Sdg": ex.SdgGate,
+    "Tdg": ex.TdgGate,
```
josh146: How come these are kept?

pennylane_qiskit/qiskit_device.py
```diff
@@ -82,29 +82,26 @@ def pauli_eigs(n):
...
     "CSWAP": ex.FredkinGate,
     "CRZ": ex.CrzGate,
     "PhaseShift": ex.U1Gate,
     "QubitStateVector": ex.Initialize,
-
-    # TODO: add the CRY gate, once the U3Gate is a part of PennyLane
-    # "CRY": U3Gate,
```
josh146: Did you work out what this TODO meant?

josh146: The confusing thing is that `CRY` is a one qubit gate, but `U3` is a 2 qubit gate...

tests/test_apply.py
```diff
@@ -48,9 +48,7 @@
     ("PauliZ", Z),
     ("Hadamard", H),
     ("S", S),
-    ("Sdg", S.conj().T),
     ("T", T),
```
josh146: You could add the following here,
```suggestion
    ("T", T),
    ("S.inv", S.conj().T),
    ("T.inv", T.conj().T),
```
and this should work right? Might be good to leave here as a uni ttest.

tests/test_integration.py
```diff
@@ -172,3 +172,30 @@ def quantum_circuit():
...
+            return qml.expval(qml.PauliZ(0))
+
+        @qml.qnode(dev2)
+        def circuit_with_inverses_default_qubit(angle):
+            qml.Hadamard(0).inv()
+            qml.RX(angle, wires=0).inv()
+            return qml.expval(qml.PauliZ(0))
+
+        for x in angles:
+            assert np.allclose(circuit_with_inverses(x), circuit_with_inverses_default_qubit(x))
```
josh146: Nice test!

tests/test_integration.py
```diff
@@ -172,3 +172,30 @@ def quantum_circuit():
             return qml.expval(qml.PauliZ(0))
 
         assert np.allclose(loaded_quantum_circuit(), quantum_circuit())
+
+
+class TestInverses:
+    """Integration tests checking that the inverse of the operations are applied."""
+
+    def test_inverse_of_operation(self):
+        """Test that the inverse of operations works as expected."""
```
josh146:
```suggestion
        """Test that the inverse of operations works as expected
        by comparing a simple circuit with default.qubit."""
```

tests/test_inverses.py
```diff
@@ -0,0 +1,180 @@
...
+        supported on the default gaussian device"""
+
+        dev = qml.device('qiskit.aer', backend='statevector_simulator', wires=2, analytic=True)
+
+        @qml.qnode(dev)
+        def mean_photon_gaussian(phi):
+            qml.Rotation(phi, wires=0).inv()
+            return qml.expval(qml.NumberOperator(0))
+
+        with pytest.raises(qml.DeviceError, match="Gate Rotation.inv not supported on device {}"
```
josh146: awesome!

tests/test_inverses.py
```diff
@@ -0,0 +1,180 @@
...
+
+        @qml.qnode(dev)
+        def circuit():
+            qml.S(wires=[0]).inv()
+            return qml.expval(qml.PauliZ(0))
+
+        assert np.allclose(circuit(), expected_output, atol=tol, rtol=0)
+
+    def test_inverse_operations_not_supported(self):
+        """Test that the inverse of operations is not currently
```
josh146:
```suggestion
        """Test that unsupported inverse operations raise the correct exception"""
```

tests/test_inverses.py
```diff
@@ -0,0 +1,180 @@
...
+        supported on the default gaussian device"""
```
josh146:
```suggestion
```

pennylane_qiskit/qiskit_device.py
antalszava: Great one, thanks for that! Updated it in the latest commit

tests/test_apply.py
antalszava: I believe these were included since there were separate ``Sdg`` and ``Tdg`` gates defined, now I'd say the unit tests in ``test_inverses.py`` should suffice. What do you think?

josh146: The tests in `test_inverses.py` are more like integration tests, however, in that they make use of the PennyLane QNode and PennyLane operation. This test is much more of a unit test, avoiding most external code and simply calling `device.apply()` directly.

I think it's good to have both, as they serve different purposes;

* If something breaks in `test_inverses.py` but not `test_apply`, we know the issue is with the PL-Core/PL-qiskit integration.

* If something breaks in `test_apply` _and_ `test_inverses.py`, we know that the issue is likely with the plugins' use of the Qiskit API.

PR #59, opened 2019-11-19
Analytic true hw simulator warning

antalszava:
Adds a warning for the expectation and variance computations in the case of when the ``analytic`` attribute is set to ``True`` on hardware simulators.


pennylane_qiskit/qiskit_device.py
```diff
@@ -29,26 +29,26 @@
...
-from qiskit.converters import dag_to_circuit, circuit_to_dag
+from qiskit.compiler import assemble, transpile
+from qiskit.converters import circuit_to_dag, dag_to_circuit
 
 from pennylane import Device, DeviceError
 from pennylane.operation import Sample
 
-from .gates import BasisState, Rot
 from ._version import __version__
+from .gates import BasisState, Rot
```
josh146: :+1: glad the ordering here has been fixed

pennylane_qiskit/qiskit_device.py
```diff
@@ -350,6 +350,14 @@ def expval(self, observable, wires, par):
             prob = np.fromiter(self.probabilities(wires=wires).values(), dtype=np.float64)
             return (eigvals @ prob).real
 
+        if self.analytic:
+            # Raise a warning if backend is a hardware simulator
+            warnings.warn("The analytic calculation of expectations and variances "
+                          "is only supported on statevector backends, not on the {}. "
+                          "The obtained result is based on sampling.".
+                          format(self.backend),
+                          UserWarning)
```
josh146: Should we move the warning to `__init__`? That way,

* It doesn't have to be repeated for `expval` and `var`,
* The user will know straight away and can change it, rather than waiting for the qnode to be evaluated

tests/test_qiskit_device.py
```diff
@@ -1,39 +1,118 @@
...
-        None if no job has yet been run."""
-        dev = AerDevice(backend="statevector_simulator", wires=1, analytic=True)
-        assert dev.probabilities() is None
+import pytest
+
+import numpy as np
+
+from pennylane_qiskit.qiskit_device import pauli_eigs
+from pennylane_qiskit import AerDevice
+import pennylane as qml
```
josh146: Should this go below the `import numpy` line, since it is a third party import

antalszava: Ran ``isort`` on it now and got the following:
```python
import numpy as np
import pytest

import pennylane as qml
from pennylane_qiskit import AerDevice
from pennylane_qiskit.qiskit_device import pauli_eigs
```

josh146: :+1:

pennylane_qiskit/qiskit_device.py
antalszava: I first went for it to be in the ``__init__.py``. In that case, however, as by default ``analytic=True`` is used and there will be warnings raised for sampling on HW simulators

pennylane_qiskit/qiskit_device.py
```diff
@@ -134,6 +134,10 @@ class QiskitDevice(Device, abc.ABC):
     operations = set(_operation_map.keys())
     observables = {"PauliX", "PauliY", "PauliZ", "Identity", "Hadamard", "Hermitian"}
 
+    hw_analytic_warning_message = "The analytic calculation of expectations and variances "\
```
antalszava: Created separate warning message for ``expval`` and ``var``

Issue #52, opened 2019-10-15
QiskitError on the data parity function of the pennylane tutorial


RicardoGaGu:
I'm trying to reproduce the data parity variational classifier of pennylane's tutorial collection on the qiskit aer simulator backend,  the following line of code specifies the backend:

```python
dev = qml.device('qiskit.aer', wires=4)

```

This is the variational architecture I'm using, plus the basic data encoding technique _BasisState_ for the bitstring data:

```python
def layer(W):
    qml.Rot(W[0, 0], W[0, 1], W[0, 2], wires=0)
    qml.Rot(W[1, 0], W[1, 1], W[1, 2], wires=1)
    qml.Rot(W[2, 0], W[2, 1], W[2, 2], wires=2)
    qml.Rot(W[3, 0], W[3, 1], W[3, 2], wires=3)

    qml.CNOT(wires=[0, 1])
    qml.CNOT(wires=[1, 2])
    qml.CNOT(wires=[2, 3])
    qml.CNOT(wires=[3, 0])
    
def statepreparation(x):
    qml.BasisState(x, wires=[0, 1, 2, 3])

```
I get the following the error when I execute the training loop part:

`QiskitError: "Cannot unroll the circuit to the given basis, ['u1', 'u2', 'u3', 'cx', 'cz', 'cu1', 'id', 'x', 'y', 'z', 'h', 's', 'sdg', 't', 'tdg', 'ccx', 'swap', 'multiplexer', 'snapshot', 'unitary', 'reset', 'initialize', 'kraus', 'roerror']. No rule to expand instruction basis_state."`

I sense that the qml.Rot, i.e. the  single arbitrary qubit rotation, is not supported by the Qiskit plugin.

**Desktop**
 - OS: Ubuntu 18.04
 - Version 0.5.1 (pennylane-qiskit)

josh146:
Hi @RicardoGaGu! Can I ask what version of Qiskit you have installed, and what version of PennyLane? The PennyLane-Qiskit plugin defines a [custom `Rot` gate](https://github.com/XanaduAI/pennylane-qiskit/blob/master/pennylane_qiskit/gates.py#L59) that should be qiskit-compatible, so this is a strange error.

RicardoGaGu:
Sorry for the delay Josh!

Here are the versions of the libraries you asked:

- Qiskit version  0.9.0

- Pennylane version 0.5.0

josh146:
[...]

PR #60, opened 2019-11-19
Remove rotation and basisstate

antalszava:
Remove ``gates.py`` such that no PL-Qiskit ``Rot`` and ``Basisstate`` are used anymore. These operations get decomposed based on the implementation in PennyLane core.

Added an integration test for these two operations, such that they are decomposed. In the test, ``analytic=False`` for the device, such that hw simulators are not skipped.

This PR follows up on issue #52.


tests/conftest.py
```diff
@@ -1,85 +1,102 @@
...
+
+
+@pytest.fixture(params=hw_backends)
+def hardware_backend(request):
+    return request.param
+
+
+@pytest.fixture(params=state_backends)
+def statevector_backend(request):
+    return request.param
```
antalszava: Added these for convenience in the tests

tests/conftest.py
```diff
@@ -1,85 +1,102 @@
...
+
+    return _device
+
+
+@pytest.fixture(params=[AerDevice, BasicAerDevice])
+def state_vector_device(request, statevector_backend, shots, analytic):
+    def _device(n):
+        return request.param(wires=n, backend=statevector_backend, shots=shots, analytic=analytic)
+
+    return _device
```
antalszava: This one as well

tests/test_converter.py
```diff
@@ -327,6 +333,16 @@ def test_one_qubit_parametrized_operations_supported_by_pennylane(self, recorder
...
+        assert recorder.queue[4].name == 'U2'
+        assert len(recorder.queue[4].params) == 2
+        assert recorder.queue[4].params == [0.3, 0.4]
+        assert recorder.queue[4].wires == [0]
+
+        assert recorder.queue[5].name == 'U3'
+        assert len(recorder.queue[5].params) == 3
+        assert recorder.queue[5].params == [0.3, 0.4, 0.2]
+        assert recorder.queue[5].wires == [0]
+
```
antalszava: These were previously in the ``test_operations_transformed_into_qubit_unitary``, now that PL core has them, they could be reorganized

tests/test_converter.py
```diff
@@ -495,19 +507,22 @@ def test_operations_transformed_into_qubit_unitary(self, recorder):
...
-        assert recorder.queue[3].name == 'U3'
-        assert len(recorder.queue[3].params) == 3
-        assert recorder.queue[3].params == [0.3, 0.4, 0.2]
-        assert recorder.queue[3].wires == [0]
+        qc = QuantumCircuit(3, 1)
 
-        assert recorder.queue[4].name == 'Toffoli'
-        assert len(recorder.queue[4].params) == 0
-        assert recorder.queue[4].wires == [0, 1, 2]
+        qc.ch([0], [1])
```
antalszava: Added a test case for an operation from Qiskit that has the ``to_matrix()`` method defined on them and is not in PL core

tests/test_inverses.py
```diff
@@ -136,6 +130,27 @@ def circuit():
...
 
+    @pytest.mark.parametrize("name,par,expected_output", [
+        ("Rot", [math.pi/2, 0, 0], 1),
+        ("Rot", [0, math.pi/2, 0], 0),
+        ("Rot", [0, 0, math.pi/2], 1),
+        ("Rot", [math.pi/2, -math.pi/4, -math.pi/4], 1/math.sqrt(2)),
+        ("Rot", [-math.pi/4, math.pi/2, math.pi/4], 0),
+        ("Rot", [-math.pi/4, math.pi/4, math.pi/2], 1/math.sqrt(2)),
+    ])
+    def test_unsupported_gate_inverses(self, name, par, expected_output):
```
antalszava: Reorganized this into a separate test case, as ``assert dev.supports_operation("Rot")`` evaluates to False

tests/test_converter.py
josh146: nice!

PR #67, opened 2020-02-03
Remove inverse test that should be checked in PL core

antalszava:
Remove the test case that tests the inverse of the qiskit plugin. The reason is that it checks the logic that is currently incorporated in `qml.Device` and therefore this test case belongs to PL core.

Removing this test further allows the test suite to pass with PL version 0.7.0 (and thus there is no need to increment the version number for PL in the `requirements.txt`.

### DIFF
--- a/tests/test_inverses.py
+++ b/tests/test_inverses.py
@@ -178,17 +178,3 @@ class TestInverses:
             return qml.expval(qml.PauliZ(0))
 
         assert np.allclose(circuit(), expected_output, atol=tol, rtol=0)
-
-    def test_inverse_operations_not_supported(self):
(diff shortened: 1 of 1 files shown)


tests/test_inverses.py
```diff
@@ -178,17 +178,3 @@ def circuit():
             return qml.expval(qml.PauliZ(0))
 
         assert np.allclose(circuit(), expected_output, atol=tol, rtol=0)
-
```
antalszava: This test case coincides with the following PL test:
https://github.com/XanaduAI/pennylane/blob/e180c7ed5a005277a986222879ed9b569b83bb7d/tests/test_device.py#L238

PR #71, opened 2020-02-24
Fixed bug in gradients of loaded qiskit circuits

josh146:
**Context:** By expliciting calling `params.val`, the `VariableRef` classes, which allow us to keep track of which operations are using which parameters so that the parameter shift rule can be applied, are being lost in the conversion.

**Description of the changes:**

* The binding logic in the conversion functions has been modified --- now, a mapping is kept of the differentiable Qiskit parameters to the PennyLane variables. The PennyLane variables are then used for operation execution where needed.

* An integration test has been added to ensure that the gradient computation works correctly on converted qiskit circuits.

### DIFF
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -37,20 +37,20 @@ from pennylane_qiskit.qiskit_device import QISKIT_OPERATION_MAP
 inv_map = {v.__name__: k for k, v in QISKIT_OPERATION_MAP.items()}
 
 
-def _check_parameter_bound(param: Parameter, var_ref_map: Dict[Parameter, qml.variable.VariableRef]):
+def _check_parameter_bound(param: Parameter, var_ref_map: Dict[Parameter, qml.variable.Variable]):
     """Utility function determining if a certain parameter in a QuantumCircuit has
     been bound.
 
     Args:
         param (qiskit.circuit.Parameter): the parameter to be checked
-        var_ref_map (dict[qiskit.circuit.Parameter, pennylane.variable.VariableRef]):
+        var_ref_map (dict[qiskit.circuit.Parameter, pennylane.variable.Variable]):
             a dictionary mapping qiskit parameters to PennyLane variables
     """
     if isinstance(param, Parameter) and param not in var_ref_map:
         raise ValueError("The parameter {} was not bound correctly.".format(param))
 
 
-def _extract_variable_refs(params: Dict[Parameter, Any]) -> Dict[Parameter, qml.variable.VariableRef]:
+def _extract_variable_refs(params: Dict[Parameter, Any]) -> Dict[Parameter, qml.variable.Variable]:
     """Iterate through the parameter mapping to be bound to the circuit,
     and return a dictionary containing the differentiable parameters.
 
@@ -58,14 +58,14 @@ def _extract_variable_refs(params: Dict[Parameter, Any]) -> Dict[Parameter, qml.
         params (dict): dictionary of the parameters in the circuit to their corresponding values
 
     Returns:
-        dict[qiskit.circuit.Parameter, pennylane.variable.VariableRef]: a dictionary mapping
+        dict[qiskit.circuit.Parameter, pennylane.variable.Variable]: a dictionary mapping
             qiskit parameters to PennyLane variables
     """
-    # map qiskit parameters to PennyLane differentiable VariableRefs.
+    # map qiskit parameters to PennyLane differentiable Variables.
     if params is None:
         return {}
 
-    return {k: v for k, v in params.items() if isinstance(v, qml.variable.VariableRef)}
+    return {k: v for k, v in params.items() if isinstance(v, qml.variable.Variable)}
 
 
 def _check_circuit_and_bind_parameters(quantum_circuit: QuantumCircuit, params: dict, diff_params: dict) -> QuantumCircuit:
@@ -75,7 +75,7 @@ def _check_circuit_and_bind_parameters(quantum_circuit: QuantumCircuit, params: 
         quantum_circuit (QuantumCircuit): the quantum circuit to check and bind the parameters for
         params (dict): dictionary of the parameters in the circuit to their corresponding values
         diff_params (dict): dictionary mapping the differentiable parameters to PennyLane
-            VariableRef instances
+            Variable instances
 
     Returns:
         QuantumCircuit: quantum circuit with bound parameters
@@ -87,7 +87,7 @@ def _check_circuit_and_bind_parameters(quantum_circuit: QuantumCircuit, params: 
         return quantum_circuit
 
     for k in diff_params:
-        # Since we cannot bind VariableRefs to Qiskit circuits,
+        # Since we cannot bind Variables to Qiskit circuits,
         # we must remove them from the binding dictionary before binding.
         del params[k]
 
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -20,6 +20,7 @@ QuanctumCircuit converter module
 This module contains functions for converting Qiskit QuantumCircuit objects
 into PennyLane circuit templates.
 """
+from typing import Dict, Any
 import warnings
 
 import numpy as np
@@ -36,7 +37,7 @@ from pennylane_qiskit.qiskit_device import QISKIT_OPERATION_MAP
 inv_map = {v.__name__: k for k, v in QISKIT_OPERATION_MAP.items()}
 
 
-def _check_parameter_bound(param, var_ref_map):
+def _check_parameter_bound(param: Parameter, var_ref_map: Dict[Parameter, qml.variable.VariableRef]):
     """Utility function determining if a certain parameter in a QuantumCircuit has
     been bound.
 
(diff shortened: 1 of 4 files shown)


pennylane_qiskit/converter.py
```diff
@@ -83,7 +99,7 @@ def map_wires(wires: list, qc_wires: list) -> dict:
         qc_wires (list): wires from the converted quantum circuit
 
     Returns:
-        wire_map (dict): map from quantum circuit wires to the user defined wires
+        dict: map from quantum circuit wires to the user defined wires
```
antalszava:
```suggestion
        dict[int, int]: map from quantum circuit wires to the user defined wires
```

pennylane_qiskit/converter.py
```diff
@@ -59,19 +61,33 @@ def _check_circuit_and_bind_parameters(quantum_circuit: QuantumCircuit, params:
...
+    # map qiskit parameters to PennyLane differentiable VariableRefs.
+    var_ref_map = {}
 
     for k, v in params.items():
-        if isinstance(v, qml.variable.Variable):
-            params.update({k: v.val})
+        # iterate through the parameters
+        if isinstance(v, qml.variable.VariableRef):
+            # If a parameter is a Variable reference, then it represents
+            # a differentiable parameter.
```
antalszava: This could go as a dictionary comprehension as well (though might be too long of a one-liner):
```python
var_ref_map = {k:v for k, v in params.items() if isinstance(v, qml.variable.VariableRef)}
```

pennylane_qiskit/converter.py
```diff
@@ -59,19 +61,33 @@ def _check_circuit_and_bind_parameters(quantum_circuit: QuantumCircuit, params:
...
+            var_ref_map[k] = v
+
+    for k in var_ref_map:
+        # Since we cannot bind VariableRefs to Qiskit circuits,
+        # we must remove them from the binding dictionary before binding.
+        del params[k]
 
-    return quantum_circuit.bind_parameters(params)
+    return quantum_circuit.bind_parameters(params), var_ref_map
```
antalszava: It might be slightly unintuitive that here there's a tuple that is returned (but this is also connected to the original implementation which was my doing :sweat_smile: ). Given the change there seem to be four parts of logic here inside of this single function:
* check_circuit
* extract `VariableRefs`
* remove `VariableRefs` from params
* bind parameters

While in theory, this means four separate functions it might be worth having a separate function that solely extracts the `VariableRefs` and then later the `bind_parameters` function could delete the needed refs. To me, this latter separation would make this part easier to parse :)

pennylane_qiskit/converter.py
```diff
@@ -160,8 +170,24 @@ def _function(params: dict = None, wires: list = None):
...
+                # variable from the var_ref_map dictionary.
+
+                parameters = []
+                for p in op[0].params:
+
+                    _check_parameter_bound(p, var_ref_map)
+
+                    if isinstance(p, ParameterExpression):
+                        if p.parameters:
+                            parameters.append(var_ref_map.get(min(p.parameters)))
```
antalszava: Not too sure about this line here, in particular `min(p.parameters)`. How come such a call is made? Perhaps naming a certain part of this line (e.g. `min_param=min(p.parameters)`) and changing this line accordingly could help with providing further clues to the reader :)

tests/test_converter.py
```diff
@@ -946,3 +951,37 @@ def circuit_native_pennylane():
...
+        qc.rx(qiskit_params[0], 0)
+        qc.rx(qiskit_params[1], 1)
+        qc.rx(qiskit_params[2], 2)
+        qc.cnot(0, 1)
+        qc.cnot(1, 2)
+
+        # convert to a PennyLane circuit
+        qc_pl = qml.from_qiskit(qc)
+
+        dev = qml.device("default.qubit", wires=3)
```
antalszava:
```suggestion
        dev = qml.device("default.qubit", wires=3, analytic=analytic)
```

tests/test_converter.py
```diff
@@ -946,3 +951,37 @@ def circuit_native_pennylane():
             return qml.expval(qml.PauliZ(0))
 
         assert circuit_loaded_qiskit_circuit() == circuit_native_pennylane()
+
+    @pytest.mark.parametrize("analytic", [True])
+    @pytest.mark.parametrize("theta,phi,varphi", list(zip(THETA, PHI, VARPHI)))
+    def test_gradient(self, theta, phi, varphi, tol):
```
antalszava:
```suggestion
    def test_gradient(self, theta, phi, varphi, analytic, tol):
```

pennylane_qiskit/converter.py
```diff
@@ -160,8 +170,24 @@ def _function(params: dict = None, wires: list = None):
...
+                        else:
+                            parameters.append(float(p))
+                    else:
+                        parameters.append(p)
```
speller26: Don't feel particularly strongly about this, but this could be pulled into a helper method, so `parameters` then gets constructed as a list comprehension.

requirements.txt
```diff
@@ -1,4 +1,4 @@
 qiskit>=0.15
-PennyLane>=0.7.0
+git+git://github.com/XanaduAI/pennylane.git#egg=pennylane
```
speller26: Curious why this is needed

pennylane_qiskit/converter.py
```diff
@@ -36,17 +36,19 @@
...
 
-def _check_parameter_bound(param):
+def _check_parameter_bound(param, var_ref_map):
     """Utility function determining if a certain parameter in a QuantumCircuit has
     been bound.
 
     Args:
-        param: the parameter to be checked
+        param (qiskit.circuit.Parameter): the parameter to be checked
+        var_ref_map (dict[qiskit.circuit.Parameter, pennylane.variable.VariableRef]):
```
speller26: At this point, it might be worth considering the `typing` module:

```python
from typing import Dict

def _check_parameter_bound(param: ..., var_ref_map: Dict[Parameter, VariableRef]) -> Parameter:
    """
    ...
    var_ref_map (Dict[Parameter, VariableRef]): ...
    """
    ...
```

pennylane_qiskit/converter.py
```diff
@@ -160,8 +170,24 @@ def _function(params: dict = None, wires: list = None):
...
 
             if instruction_name in inv_map and inv_map[instruction_name] in pennylane_ops.ops:
+                # Extract the bound parameters from the operation. If the bound parameters are a
+                # Qiskit ParameterExpression, then replace it with the corresponding PennyLane
+                # variable from the var_ref_map dictionary.
+
+                parameters = []
+                for p in op[0].params:
+
+                    _check_parameter_bound(p, var_ref_map)
```
speller26: Can't help but feel we can use the return value of `_check_parameter_bound`

pennylane_qiskit/converter.py
speller26: Does `len(p.parameters) == 1`? If not, what happens to `p`'s other parameters?

pennylane_qiskit/converter.py
josh146: Good idea, I will add that in!

pennylane_qiskit/converter.py
josh146: I've removed the return statement on _check_parameter_bound, since it is never used

pennylane_qiskit/converter.py
josh146: Oh, this is partly because

* Python doesn't provide a way to index into a set, and

* Due to a PennyLane restriction, the Qiskit ParameterExpression must always be a set containing a single parameter.

As a result, `min(single_element_set)` is a useful way of extracting the element. Another approach is `single_element_set.pop()`, but I tend not to favour `pop` due to its side effects.

Side note: probably out of scope for this PR, but we should create a PR that raises an error if `len(p.parameters) > 1`.

pennylane_qiskit/converter.py
josh146: I had originally had this as a list comprehension, but changed it to a loop for readability (it became quite long due to the ternary expressions)

requirements.txt
josh146: A recent PR in PennyLane master renamed the `Variable` class to `VariableRef` --- I wanted to double check that this PR passes against the latest master of PennyLane. However, this renaming will likely be reverted due to side effects that occurred, so this change here is temporary and should be reverted before this PR is merged.

pennylane_qiskit/converter.py
antalszava: Oh, got ya! So a [`ParameterExpression` contains](https://qiskit.org/documentation/_modules/qiskit/circuit/parameterexpression.html) a 

> Mapping of Parameter instances to the sympy.Symbol serving as their placeholder in expr.

Having seen that, I was also wondering why simply a `Parameter` could not do it here, but this now allows for `ParameterExpression` instances that contain only one `Parameter` as well.

josh146: Yep - the issue seems to be that even if you start off with a single `Parameter`, after binding Qiskit converts _everything_ into a `ParameterExpression`. So `Parameter` objects never appear at this point in the logic.

co9olguy:
Ready to merge @josh146?

josh146:
I'm holding off, as I would like to revert `VariableRef` -> `Variable` in PennyLane master first, to avoid updating all the other plugins as well.

Issue #77, opened 2020-03-26
Error for gradient with tensor observables QiskitDevice


antalszava:
**Describe the bug**
An error is raised when the gradient of a `QNode` with tensor observables is computed when using the `Qiskit` device. (Thanks @trbromley for catching this!)

**To Reproduce**
```python
import pennylane as qml
import numpy as np

n_qubits = 2
depth = 2
noise = 0.1

def ansatz(weights):
    weights = weights.reshape(depth, n_qubits, 3)
    qml.templates.StronglyEntanglingLayers(weights, wires=list(range(n_qubits)))
#     return qml.expval(qml.PauliZ(0))  # The error does NOT occur when using a single qubit observable
    return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))

from qiskit.providers.aer.noise import NoiseModel
from qiskit.providers.aer.noise.errors.standard_errors import (
    depolarizing_error,
    amplitude_damping_error,
)

noise_model = NoiseModel()
noise_model.add_all_qubit_quantum_error(depolarizing_error(noise, 1), ["u1", "u2", "u3"])

dev_qsk = qml.device(
                "qiskit.aer",
                wires=n_qubits,
                shots=10000,
                noise_model=noise_model,
                backend="qasm_simulator",
            )

weights = np.random.random((depth, n_qubits, 3)).flatten()

exp_sampled = qml.QNode(ansatz, dev_qsk, diff_method="parameter-shift") # Want to get expectation value and gradient
grad_shift = qml.grad(exp_sampled, argnum=0)
print(exp_sampled(weights))
print(grad_shift(weights))
```
**Expected behavior**
The gradient is computed without any errors and the correct result is returned.

**Screenshots**
```python
~/pennylane_qiskit/qiskit_device.py in expval(self, observable, wires, par)
    353 
    354         # estimate the ev
--> 355         return np.mean(self.sample(observable, wires, par))
    356 
    357     def var(self, observable, wires, par):

~/pennylane_qiskit/qiskit_device.py in sample(self, observable, wires, par)
    404 
    405         for w, b in zip(eigvals, itertools.product([0, 1], repeat=len(wires))):
--> 406             samples = np.where(np.all(res == b, axis=1), w, samples)
    407 
    408         return samples

<__array_function__ internals> in where(*args, **kwargs)

ValueError: operands could not be broadcast together with shapes (9999,) () (10000,) 
```
**qml.about() output**
```
Name: PennyLane
[...]

PR #83, opened 2020-04-20
Porting to QubitDevice

antalszava:
**Context:**
The [`QubitDevice` class has been added to PennyLane](https://github.com/XanaduAI/pennylane/pull/452) and each device can now use shared functionalities implemented in this base class.

Changed the inheritance of `QiskitDevice` to inherit from `QubitDevice` instead of `Device`. More specific changes of methods within the class can be found at the end of this section.

Further changes:
- The `memory` flag from Qiskit has been set to a default of `True` for hardware and hardware simulator devices. For statevector simulators it is set to `False`. This allows for a simplified logic within `generate_samples` as samples can just be extracted from the stored memory. This replaces the previous logic of using the Qiskit method `Result.get_counts` and reproducing the samples.
- Modified the test suite to use the new `apply` syntax, adding `generate_samples`
- Modified the tests for the `S` and `T` gate inverses
* Fixed `TestHardwareApply` tests such that they indeed estimate probabilities instead of using stored states, defined `hardware_simulator_device` fixture
* Added fixture to skip `expval` tests for `unitary_simulator`
* Added a test picking up the case from #77 

**Benefits:**
The `QiskitDevice` class now inherits from the `QubitDevice` class from PennyLane core. This meant that many duplicated features could be removed from the code base of this plugin.

**Possible Drawbacks:**
The ability to adjust the `memory` flag used by Qiskit has been removed for now. This might be a [useful feature for `PulseSimulator`](https://github.com/Qiskit/qiskit-aer/blob/7de864cc989b354d0ad1febbd62b891270927b6b/qiskit/providers/aer/backends/pulse_simulator.py#L99).

**Related GitHub Issues:**
#77

**Changes to the methods of `QiskitDevice`**

- The following `QubitDevice` methods were introduced:
    - [X] `generate_samples`
    - [X] `analytic_probability`

- The following `QubitDevice` methods were removed:
    - [X] `pre_measure`
    - [X] `rotate_basis`
    - [X] `eigvals`
    - [X] `sample`
    - [X] `probability`
    - [X] `expval`
    - [X] `var`


pennylane_qiskit/qiskit_device.py
```diff
@@ -180,61 +166,122 @@ def __init__(self, wires, provider, backend, shots=1024, **kwargs):
...
 
     def compile(self):
         """Compile the quantum circuit to target
         the provided compile_backend. If compile_backend is None,
         then the target is simply the backend."""
         compile_backend = self.compile_backend or self.backend
         compiled_circuits = transpile(self._circuit, backend=compile_backend)
+
+        # Specify to have a memory for hw/hw simulators
+        memory = str(compile_backend) in self._hw_backends
```
antalszava, line 281: This is so since the [`qasm_simulator` supports `memory`](https://github.com/Qiskit/qiskit-aer/blob/7de864cc989b354d0ad1febbd62b891270927b6b/qiskit/providers/aer/backends/qasm_simulator.py#L231).

pennylane_qiskit/qiskit_device.py
```diff
@@ -30,29 +28,11 @@
...
 
-from pennylane import Device, QuantumFunctionError
-from pennylane.operation import Sample
+from pennylane import QubitDevice, QuantumFunctionError
 
 from ._version import __version__
 
 
-@functools.lru_cache()
-def pauli_eigs(n):
```
josh146, line 40: Nice that this can be removed!

pennylane_qiskit/qiskit_device.py
```diff
@@ -85,7 +65,7 @@ def pauli_eigs(n):
 QISKIT_OPERATION_INVERSES_MAP = {k + ".inv": v for k, v in QISKIT_OPERATION_MAP.items()}
 
 
-class QiskitDevice(Device, abc.ABC):
+class QiskitDevice(QubitDevice, abc.ABC):
```
josh146, line 68: 🔥

pennylane_qiskit/qiskit_device.py
```diff
@@ -114,31 +94,44 @@ class QiskitDevice(Device, abc.ABC):
     _state_backends = {"statevector_simulator", "unitary_simulator"}
     """set[str]: Set of backend names that define the backends
     that support returning the underlying quantum statevector"""
+    _hw_backends = {"qasm_simulator", "ibmq_qasm_simulator"}
```
josh146, line 97: I think the reason `_hw_backends` wasn't defined previously is that the available hardware backends are server dependent, and become available independently of Qiskit releases (e.g., `ibm_rome` became available just today). By hardcoding them in here, will that cause issues?

pennylane_qiskit/qiskit_device.py
```diff
@@ -114,31 +94,44 @@ class QiskitDevice(Device, abc.ABC):
...
     operations = set(_operation_map.keys())
     observables = {"PauliX", "PauliY", "PauliZ", "Identity", "Hadamard", "Hermitian"}
 
     hw_analytic_warning_message = (
-        "The analytic calculation of expectations and variances "
-        "is only supported on statevector backends, not on the {}. "
-        "The obtained result is based on sampling."
+        "The analytic calculation of expectations, variances and "
+        "probabilities is only supported on statevector backends, not on the {}. "
+        "Such statistics obtained from this device are estimates based on samples."
```
josh146, line 107: :+1:

pennylane_qiskit/qiskit_device.py
```diff
@@ -114,31 +94,44 @@ class QiskitDevice(Device, abc.ABC):
...
     def __init__(self, wires, provider, backend, shots=1024, **kwargs):
         super().__init__(wires=wires, shots=shots)
 
+        # Keep track if the user specified analytic to be True
+        analytic_was_specified = "analytic" in kwargs and kwargs["analytic"]
         self.analytic = kwargs.pop("analytic", False)
 
+        if self.analytic and backend not in self._state_backends:
+
+            if analytic_was_specified:
```
josh146, line 121: How come this if statement is required? Shouldn't the warning be raised if `self.analytic == True`, regardless if the user specified it or not?

pennylane_qiskit/qiskit_device.py
```diff
@@ -114,31 +94,44 @@ class QiskitDevice(Device, abc.ABC):
...
+
         if "verbose" not in kwargs:
             kwargs["verbose"] = False
 
         self.provider = provider
         self.backend_name = backend
         self._capabilities["backend"] = [b.name() for b in self.provider.backends()]
 
-        # check that backend exists
+        # check that the backend exists
```
josh146, line 134: nice catch 😆

pennylane_qiskit/qiskit_device.py
```diff
@@ -152,15 +145,8 @@ def __init__(self, wires, provider, backend, shots=1024, **kwargs):
...
-        self._reg = QuantumRegister(wires, "q")
-        self._creg = ClassicalRegister(wires, "c")
-        self._circuit = None
-        self._current_job = None
-        self._state = None  # statevector of a simulator backend
-
-        # job execution options
-        self.memory = False  # do not return samples, just counts
+        # Initialize inner state
+        self.reset()
```
josh146, line 149: Much cleaner!

pennylane_qiskit/qiskit_device.py
```diff
@@ -180,61 +166,122 @@ def __init__(self, wires, provider, backend, shots=1024, **kwargs):
...
+    def apply_operations(self, operations):
+        """Apply the circuit operations.
 
-            if len(par[0]) != 2 ** len(wires):
-                raise ValueError("Unitary matrix must be of shape (2**wires, 2**wires).")
+        This method serves as an auxiliary method to :meth:`~.QiskitDevice.apply`.
+
+        Args:
+            operations (List[pennylane.Operation]): operations that rotate into the
+                measurement basis
```
josh146, line 208:
```suggestion
            operations (List[pennylane.Operation]): operations to be applied
```

pennylane_qiskit/qiskit_device.py
```diff
@@ -180,61 +166,122 @@ def __init__(self, wires, provider, backend, shots=1024, **kwargs):
...
+            wires = operation.wires
+            par = operation.parameters
+            operation = operation.name
 
-        dag = circuit_to_dag(QuantumCircuit(self._reg, self._creg, name=""))
-        gate = mapped_operation(*par)
+            mapped_operation = self._operation_map[operation]
+
+            # TODO: Once a fix is available in Qiskit-Aer, remove the
+            # following:
```
josh146, line 225: Is there a github issue related to this? Might be good to provide a link to it in the comment!

pennylane_qiskit/qiskit_device.py
josh146: Hmm, wondering if it is better to have 
```suggestion
        memory = str(compile_backend) not in self._state_backends
```
and not explicitly define `hw_backends`. This avoids issues where the plugin needs to be updated as new hardware backends come online.

pennylane_qiskit/qiskit_device.py
```diff
@@ -267,258 +314,27 @@ def _get_state(self, result):
...
-        wires = np.hstack(wires)
-        res = samples[:, np.array(wires)]
-        samples = np.zeros([self.shots])
+            # software simulator: need to sample from probabilities
+            return super().generate_samples()
 
-        for w, b in zip(eigvals, itertools.product([0, 1], repeat=len(wires))):
-            samples = np.where(np.all(res == b, axis=1), w, samples)
+        # hardware or hardware simulator
+        samples = self._current_job.result().get_memory()
```
josh146, line 325: So much cleaner and more concise!

pennylane_qiskit/qiskit_device.py
```diff
@@ -267,258 +314,27 @@ def _get_state(self, result):
...
 
-        return samples
+        # reverse qubit order to match PennyLane convention
+        return np.vstack([np.array([int(i) for i in s[::-1]]) for s in samples])
 
     @property
     def state(self):
         return self._state
```
josh146, line 332: Will there be issues with qubit ordering here, due to the difference in conventions between PL and Qiskit? If so, this might need to be fixed to produce correct analytic probability.

pennylane_qiskit/qiskit_device.py
```diff
@@ -267,258 +314,27 @@ def _get_state(self, result):
...
-        elif observable == "Identity":
-            # single wire identity observable
-            eigvals = np.ones(2 ** len(wires))
-
-        return eigvals
-
-    def reset(self):
-        self._circuit = QuantumCircuit(self._reg, self._creg, name="temp")
-        self._state = None
+        prob = self.marginal_prob(np.abs(self._state) ** 2, wires)
```
josh146, line 339: see comment above re: probability qubit ordering.

tests/test_apply.py
```diff
@@ -183,132 +161,134 @@ def test_invalid_qubit(self, init_state, device):
...
+        dev.apply([qml.QubitStateVector(state, wires=wires), applied_operation])
+        dev._samples = dev.generate_samples()
+
+        res = np.fromiter(dev.probability(), dtype=np.float64)
+        expected = np.abs(applied_operation.matrix @ state) ** 2
         assert np.allclose(res, expected, **tol)
 
     @pytest.mark.parametrize("mat", [U, U2])
-    def test_qubit_unitary(self, init_state, device, mat, tol):
+    def test_qubit_unitary(self, init_state, hardware_simulator_device, mat, tol):
```
josh146, line 243: How come these tests don't work on statevector simulator backends any more (like they used to?)

tests/test_integration.py
```diff
@@ -88,6 +88,31 @@ def circuit(x, y, z):
...
+                    backend="qasm_simulator",
+                )
+
+        weights = np.random.random((depth, n_qubits, 3)).flatten()
+
+        # Want to get expectation value and gradient
+        exp_sampled = qml.QNode(ansatz, dev_qsk, diff_method="parameter-shift")
+        grad_shift = qml.grad(exp_sampled, argnum=0)
+        exp_sampled(weights)
+        grad_shift(weights)
```
josh146, line 115: Awesome!

josh146: (However this test is missing asserts)

tests/test_inverses.py
```diff
@@ -151,30 +154,36 @@ def circuit():
...
-    def test_s_gate_inverses(self):
+
+    @pytest.mark.parametrize("par", [np.pi/i for i in range(1, 5)])
+    def test_s_gate_inverses(self, par):
         """Tests the inverse of the S gate"""
 
         dev = qml.device('qiskit.aer', backend='statevector_simulator', wires=2, analytic=True)
 
-        expected_output = 1
+        expected_output = -0.5 * 1j * cmath.exp(-1j*par)*(-1 + cmath.exp(2j*par))
```
josh146, line 164: 🔥

pennylane_qiskit/qiskit_device.py
antalszava: As a matter of fact, I've added this option first. I think that's a fair point overall, it also assumes that the new backend would support `memory` which seems fair for a hardware backend. One downside is that it makes the code less explicit and we would by default support new backends without checking them (e.g. a situation might arise when that our test suite does not pick up a case related to a specific backend).

Having said that I'm on board with the change because it does create lower engineering overhead.

pennylane_qiskit/qiskit_device.py
antalszava: Oh you're right! Thanks for this! :blush: 
```python
IBMQ.get_provider().backends()
```
Got this:
```python
[<IBMQSimulator('ibmq_qasm_simulator') from IBMQ(hub='ibm-q', group='open', project='main')>,
 <IBMQBackend('ibmqx2') from IBMQ(hub='ibm-q', group='open', project='main')>,
 <IBMQBackend('ibmq_16_melbourne') from IBMQ(hub='ibm-q', group='open', project='main')>,
 <IBMQBackend('ibmq_vigo') from IBMQ(hub='ibm-q', group='open', project='main')>,
 <IBMQBackend('ibmq_ourense') from IBMQ(hub='ibm-q', group='open', project='main')>,
 <IBMQBackend('ibmq_london') from IBMQ(hub='ibm-q', group='open', project='main')>,
 <IBMQBackend('ibmq_burlington') from IBMQ(hub='ibm-q', group='open', project='main')>,
 <IBMQBackend('ibmq_essex') from IBMQ(hub='ibm-q', group='open', project='main')>,
 <IBMQBackend('ibmq_armonk') from IBMQ(hub='ibm-q', group='open', project='main')>,
 <IBMQBackend('ibmq_rome') from IBMQ(hub='ibm-q', group='open', project='main')>]
```

pennylane_qiskit/qiskit_device.py
antalszava: Unfortunately, we are setting `analytic=True` right now in PennyLane by default: https://github.com/XanaduAI/pennylane/blob/8a5a164b6b83fd54a141bae5f604c32ecbaf9ba7/pennylane/_qubit_device.py#L72

We could fix it quickly, but I'd imagine that it would be better to overall rethink when and for which devices we define the `analytic` attribute in the first place. E.g. if we go with subclassing `QubitDevice`, then the `StateVectorDevice` device could define `analytic` whereas `HardwareDevice` would not. Using inheritance would overall allow us to sort out this part of the logic just by using Python itself.

pennylane_qiskit/qiskit_device.py
antalszava: Oh right, sorry this was coming from this recent PR:
https://github.com/XanaduAI/pennylane-qiskit/pull/81

And the specific issue is:
https://github.com/Qiskit/qiskit-aer/issues/692

Although the issue was fixed, it's not in the pip version yet (neither `0.18.0` nor `0.18.1`). Guess we'll have to wait until `0.19.0`.

antalszava: Updated the comment such that it refers to the release. Also set on watching the `Qiskit-Aer` repo for new releases :wink:

pennylane_qiskit/qiskit_device.py
antalszava: Oh right! So the logic of this part was not changed, this was part of `sample`. The state representation is reversed by having `s[::-1]`. Is this what you mean? The comment could be updated if it is misleading.

antalszava: An example:
```python
from qiskit import QuantumCircuit, execute
from qiskit import Aer, IBMQ
from qiskit.compiler import assemble, transpile
import numpy as np

num_qubits = 2
circ = QuantumCircuit(2, 2)
circ.h(1)
circ.measure([0,1], [0,1])

backend = Aer.get_backend('qasm_simulator')

compiled_circuits = transpile(circ, backend=backend)
qobj = assemble(experiments=compiled_circuits, backend=backend,
                             shots=10, memory=True)
job = backend.run(qobj)
result = job.result()

samples = result.get_memory(0)
print('Qiskit samples: ', samples)

pennylane_samples = np.vstack([np.array([int(i) for i in s[::-1]]) for s in samples])
print('PennyLane samples: \n', pennylane_samples)
```
Which outputs:
```python
Qiskit samples:  ['10', '00', '10', '10', '00', '00', '00', '10', '10', '10']
PennyLane samples: 
 [[0 1]
 [0 0]
 [0 1]
 [0 1]
 [0 0]
 [0 0]
 [0 0]
 [0 1]
 [0 1]
 [0 1]]
```

pennylane_qiskit/qiskit_device.py
antalszava: Right! Good point! The reversal happens back in `_get_state` (as was originally included):
```python
        # reverse qubit order to match PennyLane convention
        return state.reshape([2] * self.num_wires).T.flatten()
```
(This solution makes use of the fact that transposing the tensor with `(2,2,...2)` shape is equivalent to reversing the qubit ordering from the Qiskit convention.)

The function call tree is done the following way: `QiskitDevice.apply` -> `QiskitDevice.run` -> `QiskitDevice._get_state` (only for `state_backends`).

tests/test_apply.py
antalszava: Thanks for this question! I just realized that I've had a confusion here: the first two test classes checked the cases for `analytic=True` whereas the second `analytic=False` for **all backends**. The reason for this change was that I thought that they were completely based on backends.

I've reverted to using the `device` fixture everywhere and updated the class names and docstrings.

tests/test_integration.py
antalszava: Thanks! Yes this one is just meant to check that no errors arise, as this does not work now with the `master` version

tests/test_apply.py
```diff
@@ -42,132 +42,110 @@
...
+two_qubit_param = [qml.CRZ]
+three_qubit = [qml.Toffoli, qml.CSWAP]
 
 @pytest.mark.parametrize("analytic", [True])
 @pytest.mark.parametrize("shots", [8192])
 @pytest.mark.usefixtures("skip_unitary")
-class TestStateApply:
-    """Test application of PennyLane operations to state simulators."""
+class TestAnalyticApply:
+    """Test application of PennyLane operations with analytic attribute set to True."""
```
antalszava, line 63: @josh146 included these new names and updates here, let me know what you think!

antalszava:
@josh146 thank you so much for your comments, they were really helpful! I've left an answer to them, let me know your thoughts.

pennylane_qiskit/qiskit_device.py
josh146: 👍

pennylane_qiskit/qiskit_device.py
josh146: Oh, I was trying to comment on the 
```python
    def state(self):
        return self._state
```
part! GitHub tries to show the closest change though 😆 

I wasn't 100% sure about this, because I saw the old version of the device didn't do any processing here either, but I figured if nothing is being done, the user will receive the statevector in the Qiskit convention?

pennylane_qiskit/qiskit_device.py
josh146: Perfect! Thanks @antalszava, that answers my question.

pennylane_qiskit/qiskit_device.py
josh146: But you answered my question below 🙂

tests/test_integration.py
josh146: Ah, right! I didn't see the docstring properly, and thought this test was unfinished :)

tests/test_apply.py
josh146: Sounds much better now!

tests/test_apply.py
```diff
@@ -209,106 +188,117 @@ def test_invalid_qubit_state_vector(self, device):
...
+        res = np.fromiter(dev.probability(), dtype=np.float64)
+        expected = np.abs(applied_operation.matrix @ state) ** 2
         assert np.allclose(res, expected, **tol)
 
     @pytest.mark.parametrize("theta", [0.5432, -0.232])
-    @pytest.mark.parametrize("name,func", single_qubit_param)
-    def test_single_qubit_parameters(self, init_state, device, name, func, theta, tol):
-        """Test PauliX application"""
+    @pytest.mark.parametrize("operation", single_qubit_operations_param)
+    def test_single_qubit_operations_parameters(self, init_state, device, operation, theta, tol):
```
mariaschuld, line 239: Is it possible that this test appears twice? And that ``two_qubit_pars`` is missing?

tests/test_apply.py
```diff
@@ -42,132 +42,110 @@
...
-        dev.pre_measure()
-
-        res = np.abs(dev.state) ** 2
-        expected = np.abs(mat @ state) ** 2
-        assert np.allclose(res, expected, **tol)
-
-    @pytest.mark.parametrize("theta", [0.5432, -0.232])
-    @pytest.mark.parametrize("name,func", two_qubit_param)
-    def test_single_qubit_parameters(self, init_state, device, name, func, theta, tol):
-        """Test PauliX application"""
```
mariaschuld, line 160: It was probably not related to this PR, but a few tests still have the docstring   """Test PauliX application""". Maybe a good opportunity to correct?

tests/test_apply.py
```diff
@@ -183,24 +161,25 @@ def test_invalid_qubit(self, init_state, device):
         state = init_state(1)
 
         with pytest.raises(qml.QuantumFunctionError, match="The QubitStateVector operation is not supported on the unitary simulator backend"):
-            dev.apply("QubitStateVector", [0], [state])
+            dev.apply([qml.QubitStateVector(state, wires=[0])])
 
 @pytest.mark.parametrize("shots", [8192])
 @pytest.mark.parametrize("analytic", [False])
 @pytest.mark.usefixtures("skip_unitary")
-class TestHardwareApply:
```
mariaschuld, line 191: I am a bit confused...so are Hardware backends only tested with ``analytic==False`` and State backends with ``analytic==True``? What about the other two cases?

tests/test_expval.py
```diff
@@ -16,89 +16,109 @@
...
         """Test that identity expectation value (i.e. the trace) is 1"""
         dev = device(2)
-        dev.apply("RX", wires=[0], par=[theta])
-        dev.apply("RX", wires=[1], par=[phi])
-        dev.apply("CNOT", wires=[0, 1], par=[])
 
-        O = qml.Identity
-        name = "Identity"
+        O1 = qml.Identity(wires=[0])
+        O2 = qml.Identity(wires=[1])
```
mariaschuld, line 28: Curious about this? Are ``01``, ``02`` interpreted as strings?

tests/test_qiskit_device.py
```diff
@@ -45,72 +22,22 @@ def test_warning_raised_for_hardware_backend_analytic_expval(self, hardware_back
...
-
-        @qml.qnode(dev)
-        def circuit():
-            qml.Hadamard(wires=0)
-            return qml.var(qml.PauliZ(0))
-
-        circuit()
-
         # check that no warnings were raised
         assert len(recwarn) == 0
```
mariaschuld, line 43: What exactly is ``recwarn``? It's not in the import, nor in the ``conftest.py`` file? Seems to be very useful if it just collects warnings!

tests/test_integration.py
```diff
@@ -88,6 +88,31 @@ def circuit(x, y, z):
...
 
+    def test_gradient_for_tensor_product(self):
+        """Test that the gradient of a circuit containing a tensor product is
+        computed without any errors."""
+        n_qubits = 2
+        depth = 2
+
+        def ansatz(weights):
+            weights = weights.reshape(depth, n_qubits, 3)
+            qml.templates.StronglyEntanglingLayers(weights, wires=list(range(n_qubits)))
```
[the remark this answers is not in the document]
antalszava: True! I've plugged in some rotations, but let me know if it would be better to further simplify it :)

pennylane_qiskit/qiskit_device.py
```diff
@@ -119,26 +99,36 @@ class QiskitDevice(Device, abc.ABC):
...
     )
 
     _eigs = {}
 
     def __init__(self, wires, provider, backend, shots=1024, **kwargs):
         super().__init__(wires=wires, shots=shots)
 
+        # Keep track if the user specified analytic to be True
+        analytic_was_specified = "analytic" in kwargs and kwargs["analytic"]
         self.analytic = kwargs.pop("analytic", False)
```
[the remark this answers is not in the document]
antalszava: Yes, if it was not set explicitly by the user (that is not part of `kwargs`). The `pop` is needed so that this will not be carried on in the `run_args` later.

One thing to keep in mind is that with the `super().__init__(wires=wires, shots=shots)` call we are already setting `analytic=True` by default from `QubitDevice.__init__`. More details on this topic are in this comment: https://github.com/XanaduAI/pennylane-qiskit/pull/83#discussion_r413243678

tests/test_apply.py
antalszava: Thanks! Could you check if these are alright with the latest commit? I've pushed one yesterday after noticing this and that should clear these up

tests/test_apply.py
antalszava: Right! This was also something I noticed and had a commit yesterday, could you check if it should be fine now? The idea would be to have one case for each per `class`

tests/test_apply.py
antalszava: Yes, this caused a bit of confusion for me as well initially (thanks to @josh146 who helped me out with spotting some cases that were not covered!).

What we had:
* `TestStateApply`: setting `analytic=True` and using `device` fixture that checks **all** types of backends (so not just the state backends)
* `TestHardwareApply`: setting `analytic=False` and using `device` fixture that checks **all** types of backends (so not just the hardware backends)

My confusion came from the fast that the names for the classes did not reflect what was going on in the tests, so I've renamed them and adjusted the docstrings. Let me know, however, if you think something is still amiss.

pennylane_qiskit/qiskit_device.py
```diff
@@ -180,61 +163,121 @@ def __init__(self, wires, provider, backend, shots=1024, **kwargs):
...
-        if operation == "QubitStateVector":
+    def apply(self, operations, **kwargs):
+        rotations = kwargs.get("rotations", [])
 
-            if self.backend_name == "unitary_simulator":
-                raise QuantumFunctionError(
-                    "The QubitStateVector operation is not supported on the unitary simulator backend."
-                )
+        applied_operations = self.apply_operations(operations)
+        rotation_circuits = self.apply_operations(rotations)
```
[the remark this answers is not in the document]
antalszava: Valid point, thanks! Added a comment

tests/test_expval.py
antalszava: Previously the name of the observable had to be passed to `Device.expval`. With `QubitDevice.expval`, we are passing specific `Observable` instances (e.g. `Identity`). Here `O1` and `O2` are just convenience objects to organize the observables to the beginning of the test case, but they are instances of the `Identity` class.

tests/test_qiskit_device.py
antalszava: Yes! :) It is one of the [default fixtures from `pytest`](https://docs.pytest.org/en/latest/warnings.html#recwarn). It is similar to `monkeypatch` that we frequently use. Other fixture useful for capturing output with `pytest` are `capsys`, `capfd` for example.

tests/test_integration.py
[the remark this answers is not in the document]
mariaschuld: Sure that's perfect!

pennylane_qiskit/qiskit_device.py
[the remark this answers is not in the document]
mariaschuld: Thanks!

tests/test_apply.py
mariaschuld: Great, looks good now!

tests/test_apply.py
mariaschuld: Great, looks good!

tests/test_apply.py
mariaschuld: Cool, all good!

tests/test_qiskit_device.py
mariaschuld: Nice, thanks, will look them up!

antalszava:
@mariaschuld thanks so much for the comments! :blush: (I've seen that you have also approved in the meantime)

The last couple of small commits were related to execution on `IBMQ`, the tests were failing from time to time (I'd assume depending on how quickly the requests could be satisfied). Decreased the `shots` value and recorded the `test_probability` to only have one hardware circuit call.

co9olguy:
:fireworks:

PR #76, opened 2020-03-17
Incrementing the version number to 0.9.0-dev

antalszava:

PR #88, opened 2020-05-14
Incrementing the version number to 0.9.0

antalszava:
- Checked Changelog to include relevant PRs since version increase (#76): numbered 78-83, 85-86 (#77 being an issue, #84 entry to CHANGELOG, #87 requirements related change)
- Increased the required PennyLane version and PennyLane requires version number due to [analytic probability api change in QubitDevice](https://github.com/XanaduAI/pennylane/pull/573)
- Added a Python 3.8 check to Travis


setup.py
```diff
@@ -23,7 +23,7 @@
 
 requirements = [
     "qiskit>=0.19.1",
-    "pennylane>=0.8.1",
+    "pennylane>=0.9.0",
```
antalszava, line 26: Wondering about the ["unpinned" pair to this in requirements.txt](https://github.com/XanaduAI/pennylane-qiskit/blob/9e6a7f11cb88dbc43c3852fc905103085fbe849b/requirements.txt#L2) :thinking:

josh146: It should definitely be `pennylane>=0.9` in the requirements.txt as well! We can also likely remove the Networkx pinning, I think?

antalszava: Thanks!

Would there be any specific reason for Networkx? It seems like one would get the correct version with this pinning :thinking: (I can be convinced otherwise).

josh146: Oh, simply because it echoes what Qiskit is pinning for networkx anyway, so we're just duplicating a downstream dependency. Is there a reason for duplicating the networkx pinning here?

antalszava: Thanks, removed!

antalszava:
@josh146 turned out `networkx` was funky for 3.5 after all on [Travis](https://travis-ci.com/github/XanaduAI/pennylane-qiskit/jobs/333929067) :thinking: 
([see rendered picture](https://user-images.githubusercontent.com/24476053/82010531-57015580-9640-11ea-8b53-f3e616bb5ceb.png))

After a local check, reverted the pinning, but let me know if you'd suggest another solution!

josh146:
:(

The next release of PennyLane will drop support for Python 3.5, so it might be safe to just remove the pinning _and_ the 3.5 Travis job after this release of the Qiskit plugin is made. Otherwise, looks good from my end

PR #94, opened 2020-07-24
Rename op.params to op.data

thisac:
**Context:** 
This is a small fix that renames calls to `pennylane.Operator.params` to `pennylane.Operator.data` following the [recently merged change in PL](https://github.com/XanaduAI/pennylane/pull/719).


pennylane_qiskit/converter.py
```diff
@@ -175,7 +175,7 @@ def _function(params: dict = None, wires: list = None):
                 # variable from the var_ref_map dictionary.
 
                 parameters = []
-                for p in op[0].params:
+                for p in op[0].data:
```
antalszava, line 178: Will have to stay as is, since here `op[0]` is a Qiskit gate :)
```suggestion
                for p in op[0].params:
```

josh146:
@thisac: we can remove the 3.5 tests, as PL no longer supports 3.5 😄

thisac:
> 100
> 
> Out of curiosity, could these tests be changed to use `parameters` rather than `params`?

I think they actually can be `parameters` instead. The tests seem to pass! :thinking:

PR #98, opened 2020-08-05
Fix failing integration tests

josh146:
* Adds shared device integration tests
* Configures CI to install PL from github master
* Fixes bug where `QubitUnitary().inv()` resulted in both the matrix being inverted _and_ the wires being reversed.
* Updates location of `OperationRecorder` in the tests
* Changes a `QuantumFunctionError` to a `DeviceError`, since (a) this is what the device should be raising, and (b) the shared device tests will only skip a `DeviceError`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -67,7 +67,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install git+https://github.com/PennyLaneAI/pennylane.git
           pip install -r requirements.txt
-          pip install wheel pytest pytest-cov pytest-mock --upgrade
+          pip install wheel pytest pytest-cov pytest-mock flaky --upgrade
 
       - name: Install Plugin
         run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,30 +8,79 @@ on:
 jobs:
   tests:
     runs-on: ubuntu-latest
+
     strategy:
       matrix:
         python-version: [3.6, 3.7, 3.8]
+
     steps:
       - name: Cancel Previous Runs
         uses: styfle/cancel-workflow-action@0.4.1
         with:
           access_token: ${{ github.token }}
+
       - uses: actions/checkout@v2
+
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
+
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
+          pip install git+https://github.com/PennyLaneAI/pennylane.git
           pip install -r requirements.txt
           pip install wheel pytest pytest-cov pytest-mock --upgrade
+
       - name: Install Plugin
         run: |
           python setup.py bdist_wheel
           pip install dist/PennyLane*.whl
+
       - name: Run tests
         run: python -m pytest tests --cov=pennylane_qiskit --cov-report=term-missing --cov-report=xml -p no:warnings --tb=native
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1.0.7
+        with:
+          file: ./coverage.xml
+
+  integration-tests:
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -3,3 +3,4 @@ pennylane>=0.9.0
 numpy
 networkx>=2.2;python_version>'3.5'
 networkx>=2.2,<2.4;python_version=='3.5'
+pyscf<=1.7.2
```
antalszava, line 6: How come `pyscf` was made a dependency? :thinking:

pennylane_qiskit/qiskit_device.py
```diff
@@ -234,6 +234,8 @@ def apply_operations(self, operations):
 
             if operation.endswith(".inv"):
                 gate = gate.inverse()
+                if "QubitUnitary" in operation:
+                    qregs = list(reversed(qregs))
```
antalszava, line 238: Good catch! :blush: :100: Could this be merged with the logic on lines 227-230? As is now, there would be two reversals in certain cases. Also, could it be a problem for `"QubitStateVector"` too?

pennylane_qiskit/qiskit_device.py
```diff
@@ -245,7 +247,7 @@ def qubit_state_vector_check(self, operation, par, wires):
         """Input check for the the QubitStateVector operation."""
         if operation == "QubitStateVector":
             if self.backend_name == "unitary_simulator":
-                raise QuantumFunctionError(
+                raise DeviceError(
```
antalszava, line 250: Nice! :100:

.github/workflows/tests.yml
```diff
@@ -8,30 +8,79 @@ on:
...
+          file: ./coverage.xml
+
+  integration-tests:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.4.1
+        with:
+          access_token: ${{ github.token }}
```
antalszava, line 56: Would it be good practice to add this to any new workflow files? Or would it be important here specifically for the Qiskit plugin?

.github/workflows/tests.yml
```diff
@@ -8,30 +8,79 @@ on:
...
+          pip install wheel pytest pytest-cov pytest-mock flaky --upgrade
+
+      - name: Install Plugin
+        run: |
+          python setup.py bdist_wheel
+          pip install dist/PennyLane*.whl
+
+      - name: Run tests
+        run: |
+          pl-device-test --device=qiskit.basicaer --tb=short --skip-ops --analytic=False --shots=10000 --device-kwargs backend=qasm_simulator
```
antalszava, line 79: This was one where locally I had to go for having `--shots=20000`

.github/workflows/tests.yml
```diff
@@ -8,30 +8,79 @@ on:
...
+          pl-device-test --device=qiskit.aer --tb=short --skip-ops --analytic=False --shots=10000 --device-kwargs backend=qasm_simulator
+          pl-device-test --device=qiskit.aer --tb=short --skip-ops --analytic=True --device-kwargs backend=statevector_simulator
```
antalszava, line 81: Oh, missed this one locally, that's why the tests were passing! :sweat_smile:

pennylane_qiskit/qiskit_device.py
josh146: Yes I considered this! I thought the same as you as well; that `QubitStateVector` shouldn't be invertible 😆

.github/workflows/tests.yml
josh146: Maybe! I'm not actually sure how well this action works, since it's not an official action 😆

Probably only really important on repos where we know the builds are very time consuming (like lightning)

.github/workflows/tests.yml
josh146: 🤔 hopefully it was a very low chance 😆 if it keeps happening, I will increase this

PR #99, opened 2020-08-05
[WIRES] Implement wires refactor

mariaschuld:
This PR makes the qiskit device compatible with the PennyLane master after the wires refactor was merged.

PR #100, opened 2020-08-10
Some more polishing of the wires refactor

mariaschuld:
This PR adds some items missing in the wires refactor PR #99 :

- Update some device docstrings
- Update required PL version
- Delete superfluous (and in fact wrong) line in `analytic_probabilities`


mariaschuld:
I think coverage is complaining for no reason, since the changes were in docstrings, and the only code change deleted one line that does not call another function (and can therefore not change coverage).

PR #101, opened 2020-08-11
Qiskit v0.20

antalszava:
* Removes the use of the `CU1` gate in a test which tests for unsupported instructions. The [`CU1` gate](https://github.com/Qiskit/qiskit-terra/blob/33743b54912a257e0d4ae4bc9082341685e91366/qiskit/circuit/library/standard_gates/u.py#L116) is now supported with [Qiskit v0.20](https://qiskit.org/documentation/release_notes.html#qiskit-0-20-0) and is converted using its `to_matrix` method and `QubitUnitary`.
* Monkeypatches the `IBMQX_TOKEN` environment variable in the `test_account_error` case (env var might be defined making the test fail)
* Pins Qiskit to v0.20 in `requirements.txt` and `setup.py`

### DIFF
--- a/tests/test_converter.py
+++ b/tests/test_converter.py
@@ -761,14 +761,8 @@ class TestConverterQasm:
                'x q[2];' \
                'barrier q;' \
                'h q[0];' \
-               'cu1(pi/2) q[1],q[0];' \
                'h q[1];' \
-               'cu1(pi/4) q[2],q[0];' \
-               'cu1(pi/2) q[2],q[1];' \
                'h q[2];' \
-               'cu1(pi/8) q[3],q[0];' \
-               'cu1(pi/4) q[3],q[1];' \
-               'cu1(pi/2) q[3],q[2];' \
                'h q[3];' \
                'measure q -> c;'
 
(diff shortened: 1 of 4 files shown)


tests/test_converter.py
```diff
@@ -812,16 +806,13 @@ def test_qasm_from_file(self, tmpdir, recorder):
         assert recorder.queue[5].parameters == []
         assert recorder.queue[5].wires == Wires([3])
 
-        assert len(record) == 11
+        assert len(record) == 5
```
antalszava, line 809: As far as I recall that the `CU1Gate`  was added at the time because it was the only gate that lacked the `to_matrix` method. Hence it couldn't be converted.

josh146: Nice! So it turns out the test was failing because the converter now _works_ automatically for the CU1 gate 😆

tests/test_ibmq.py
```diff
@@ -114,11 +114,13 @@ def test_load_from_disk(token):
...
 
-def test_account_error():
+def test_account_error(monkeypatch):
 
     # Token is passed such that the test is skipped if no token was provided
     with pytest.raises(IBMQAccountError, match="No active IBM Q account"):
-        IBMQDevice(wires=1)
+        with monkeypatch.context() as m:
+            m.delenv("IBMQX_TOKEN", raising=False)
+            IBMQDevice(wires=1)
```
josh146, line 123: 🎉

PR #103, opened 2020-09-21
Remove PySCF from the plugin setup.py

josh146:


setup.py
```diff
@@ -27,8 +27,7 @@
     "numpy",
     "networkx>=2.2;python_version>'3.5'",
     # Networkx 2.4 is the final version with python 3.5 support.
-    "networkx>=2.2,<2.4;python_version=='3.5'",
-    "pyscf<=1.7.2"
```
co9olguy, line 31: not necessary? how come it was in there?

josh146: PySCF has been a buggy for a while; they have released a couple of versions with broken binaries.

PennyLane and Qiskit require mutually exclusive versions of PySCF, and this causes the PennyLane-qiskit plugin to fail at pip installation if Qiskit is already installed (PL-qchem required <=1.7.2, and Qiskit required >1.7.3). We have two options:

* Including this here forces pip to downgrade pyscf if previously installed by qiskit, allowing qiskit + pl-qiskit + pl-qchem to all be installed together. The downside is that pyscf setup.py file has a bug on windows

* Removing this will no longer allow qiskit + pl-qiskit + pl-qchem to all be installed together. However, it will allow pennylane-qiskit to be installed on windows.

In retrospect, this is mostly `pip`'s fault at not being able to properly resolve dependencies, and there isn't a good solution.

josh146:
Okay, this is a lot more subtle than I first thought.

* PySCF<1.7.2 can use any version of SciPy.

* PySCF>=1.7.2 requires SciPy<1.5

Upgrading the PySCF version causes a huge number of dependency conflicts.

josh146:
See https://github.com/PennyLaneAI/pennylane/pull/814

setup.py
antalszava: Would `pyscf` had to be removed from `requirements.txt` too?

antalszava (approved):
:100: Nice! Perhaps the upcoming change to `pip` in October could help us out in some way here.

setup.py
josh146: Oops! I missed that

antalszava: Made a quick patch #104 :)

PR #106, opened 2020-10-02
Fix failing Qiskit test

josh146:
The test `TestConverterIntegration.test_gradient()` is currently failing in the file `test_converter.py`. This PR fixes the test to ensure it passes.

Note: I am unsure why the test is failing, this is likely worth exploring before merging this PR, to ensure there isn't another issue at play. This could be connected to the latest changes in the `pennylane.numpy` module.

Edit: perhaps due to the new numpy indexing behaviour returning a scalar array?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,15 @@
 
 ### Bug fixes
 
+* Fixed a bug related to extracting differentiable parameters for the Qiskit
+  converter and PennyLane array indexing.
+  [(#106)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/106)
+
 ### Contributors
 
 This release contains contributions from (in alphabetical order):
+
+Josh Izaac, Antal Száva
 
 ---
 
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -69,7 +69,7 @@ def _extract_variable_refs(params: Dict[Parameter, Any]) -> Dict[Parameter, qml.
             if isinstance(v, qml.variable.Variable):
                 variable_refs[k] = v
 
-    return variable_refs   # map qiskit parameters to PennyLane differentiable Variables.
+    return variable_refs  # map qiskit parameters to PennyLane differentiable Variables.
 
 
 def _check_circuit_and_bind_parameters(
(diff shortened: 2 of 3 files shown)


antalszava:
> perhaps due to the new numpy indexing behaviour returning a scalar array?

@josh146 think this is going to be it. In specific, `test_gradient` uses parameterization, which involves indexing into an array and also uses parts of `converter.py` that is related to determining `diff_params`. When the test was parametrized previously, we'd had a single `Variable` object for each parameter (`theta`,`phi`,`varphi`), but now we have an `array` with a single `Variable` object in there. This is something that the converter in PennyLane-Qiskit cannot handle: [it assumes that parameters are of type `Variable`](https://github.com/PennyLaneAI/pennylane-qiskit/blob/2e02dff74f87c20f21176abd3e7b5a976be0ff07/pennylane_qiskit/converter.py#L63) and doesn't cover the case of array objects containing a single `Variable` object.

**Change in `converter.py`**

Covering the edge case of a size one array when extracting the differentiable parameters, the error can be covered up:

```python
def _extract_variable_refs(params: Dict[Parameter, Any]) -> Dict[Parameter, qml.variable.Variable]:
    variable_refs = {}
    # map qiskit parameters to PennyLane differentiable Variables.
    if params is not None:
        for k, v in params.items():

            # Values be arrays of size 1, need to extract the Python scalar
            if isinstance(v, np.ndarray):
                v = v.item()

            if isinstance(v, qml.variable.Variable):
                variable_refs[k] = v

    return variable_refs
```

**How does the current quick fix make a difference?**

```python
import pennylane as qml

from pennylane import numpy as np

THETA = np.linspace(0.11, 3, 5)
PHI = np.linspace(0.32, 3, 5)
VARPHI = np.linspace(0.02, 3, 5)

# This happens implicitly when parametrizing our test case
theta = THETA[0]
phi = PHI[0]
varphi = VARPHI[0]

print([theta, phi, varphi])
print(np.array([theta, phi, varphi]))
```
Outputs
```
[tensor(0.11, requires_grad=True), tensor(0.32, requires_grad=True), tensor(0.02, requires_grad=True)]
[0.11 0.32 0.02]
```

josh146:
> Overall I think it is worth changing the logic. I was not entirely sure if this is the only case where such a thing could arise 🤔

I agree @antalszava, I think changing the logic is better than changing the test!

josh146:
@antalszava I just had a look, thanks for fixing this! The new test is 💯 as well. Feel free to approve and merge.

antalszava (approved):

PR #112, opened 2020-10-19
Qiskit v0.23 updates

antalszava:
Fix failing tests:
- [X] Converter (due to a new validation in Qiskit, error is raised from there)
- [x] Noise model extraction

*Noise models:*

Due to a [change in the signature of `QasmSimulator.run`](https://github.com/Qiskit/qiskit-aer/blame/f715703c631574565669e5ebc3b8a8b8008a1d34/qiskit/providers/aer/backends/aerbackend.py#L109), `noise_model` is no longer a parameter. A solution using the `set_options` method was added as per the [examples from Qiskit](https://github.com/Qiskit/qiskit-aer/pull/486).

A Qiskit [`NoiseModel` can be defined for the `Aer` provider and the `QasmSimulator`](https://qiskit.org/documentation/stubs/qiskit.providers.aer.noise.NoiseModel.html?highlight=noisemodel). The new solution specifically uses this information.

Updates:
- [X] Change `u3` to `u`, `u1` to `p`, and `u2(φ,λ)` to `u(π/2, φ, λ)`: added the `PhaseGate` and `UGate` gates. The rest of the gates were left to use the soon to be deprecated gates (such that users would be informed about the deprecation.
- [X] Replace `backend_options` with direct `kwargs` for `Aer`

Test cases:
- [ ] Write a test for the noise model
- [ ] Write tests for the auxiliary function operations

### DIFF
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -188,11 +188,8 @@ def load(quantum_circuit: QuantumCircuit):
             # gates exist with a different name)
             # TODO: remove the following when gates have been renamed in PennyLane
             if instruction_name == "UGate":
-                instruction_name = "U3"
-            elif instruction_name == "PhaseGate":
-                instruction_name = "U1"
+                instruction_name = "U3Gate"
 
-            print(instruction_name, inv_map[instruction_name], instruction_name in inv_map and inv_map[instruction_name] in pennylane_ops.ops)
             if instruction_name in inv_map and inv_map[instruction_name] in pennylane_ops.ops:
                 # Extract the bound parameters from the operation. If the bound parameters are a
                 # Qiskit ParameterExpression, then replace it with the corresponding PennyLane
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -187,8 +187,7 @@ def load(quantum_circuit: QuantumCircuit):
             # New Qiskit gates that are not natively supported by PL (identical
             # gates exist with a different name)
             # TODO: remove the following when gates have been renamed in PennyLane
-            if instruction_name == "UGate":
-                instruction_name = "U3Gate"
+            instruction_name = "U3Gate" if instruction_name == "UGate" else instruction_name
 
             if instruction_name in inv_map and inv_map[instruction_name] in pennylane_ops.ops:
                 # Extract the bound parameters from the operation. If the bound parameters are a
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -184,6 +184,15 @@ def load(quantum_circuit: QuantumCircuit):
 
             operation_wires = [wire_map[(qubit.register.name, qubit.index)] for qubit in op[1]]
 
+            # New Qiskit gates that are not natively supported by PL (identical
+            # gates exist with a different name)
+            # TODO: remove the following when gates have been renamed in PennyLane
+            if instruction_name == "UGate":
+                instruction_name = "U3"
+            elif instruction_name == "PhaseGate":
+                instruction_name = "U1"
+
+            print(instruction_name, inv_map[instruction_name], instruction_name in inv_map and inv_map[instruction_name] in pennylane_ops.ops)
             if instruction_name in inv_map and inv_map[instruction_name] in pennylane_ops.ops:
                 # Extract the bound parameters from the operation. If the bound parameters are a
                 # Qiskit ParameterExpression, then replace it with the corresponding PennyLane
--- a/pennylane_qiskit/qiskit_device.py
+++ b/pennylane_qiskit/qiskit_device.py
@@ -156,11 +156,12 @@ class QiskitDevice(QubitDevice, abc.ABC):
         if "compile_backend" in kwargs:
             self.compile_backend = kwargs.pop("compile_backend")
 
+        self.noise_model = None
         if "noise_model" in kwargs:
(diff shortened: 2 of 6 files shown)


tests/test_converter.py
```diff
@@ -551,36 +545,60 @@ def test_operation_transformed_into_qubit_unitary(self, recorder):
...
+
+        assert recorder.queue[1].name == 'U2'
+        assert recorder.queue[1].parameters == [0.1, 0.2]
+        assert recorder.queue[1].wires == Wires(single_wire)
+
+        assert recorder.queue[2].name == 'U3'
+        assert recorder.queue[2].parameters == [0.1, 0.2, 0.3]
+        assert recorder.queue[2].wires == Wires(single_wire)
 
-    def test_quantum_circuit_error_by_calling_wrong_parameters(self, recorder):
```
antalszava, line 570: Weird diff from GitHub, this test was removed.

It doesn't make too much sense as [parameter validation has been added to Qiskit](https://github.com/Qiskit/qiskit-terra/blob/c7c110d3d11176da3166e074a63e897d0ea02ca5/qiskit/circuit/gate.py#L235) so a string input would not work. Arrays will soon be deprecated too.

pennylane_qiskit/qiskit_device.py
```diff
@@ -147,22 +170,28 @@ def __init__(self, wires, provider, backend, shots=1024, **kwargs):
...
-        if "backend_options" in s.parameters:
+            self.noise_model = kwargs.pop("noise_model")
+
+        # Get further arguments for run
+        s = inspect.signature(b.run)
+        self.run_args = {}
+
+        if aer_provider:
+            # Consider the remaining kwargs as keyword arguments to run
+            self.run_args.update(kwargs)
```
antalszava, line 191: `backend_options` seems to be deprecated for the `run` method of `Aer` backends (e.g. for [QasmSimulator](https://qiskit.org/documentation/stubs/qiskit.providers.aer.QasmSimulator.run.html#qiskit.providers.aer.QasmSimulator.run))

pennylane_qiskit/qiskit_device.py
```diff
@@ -147,22 +170,28 @@ def __init__(self, wires, provider, backend, shots=1024, **kwargs):
...
+
+        elif "backend_options" in s.parameters:
+            # BasicAer
```
antalszava, line 194: This does not seem to be the case for `BasicAer` though (e.g. [`QasmSimulatorPy`](https://qiskit.org/documentation/stubs/qiskit.providers.basicaer.QasmSimulatorPy.run.html#qiskit.providers.basicaer.QasmSimulatorPy.run)).

pennylane_qiskit/converter.py
```diff
@@ -184,6 +184,11 @@ def _function(params: dict = None, wires: list = None):
 
             operation_wires = [wire_map[(qubit.register.name, qubit.index)] for qubit in op[1]]
 
+            # New Qiskit gates that are not natively supported by PL (identical
+            # gates exist with a different name)
+            # TODO: remove the following when gates have been renamed in PennyLane
+            instruction_name = "U3Gate" if instruction_name == "UGate" else instruction_name
```
josh146, line 190: :+1: I'll convert this into an issue so that we don't forget

josh146: https://github.com/PennyLaneAI/pennylane-qiskit/issues/113

pennylane_qiskit/qiskit_device.py
```diff
@@ -33,6 +33,25 @@
...
+    """Auxiliary function for the ``U2Gate``.
+
+    Uses the equation ``u2(phi, lam) = u(pi/2, phi, lam)``.
+    """
+    return ex.U(np.pi / 2, phi, lam)
+
+
+def U3Gate(theta, phi, lam):
+    """Auxiliary function for the ``U3Gate``."""
+    return ex.U(theta, phi, lam)
```
josh146, line 52: Very nice solution!

pennylane_qiskit/qiskit_device.py
josh146: Note that the entirety of `BasicAer` is deprecated, so might be removed in future updates. This is why the new Aer features are likely not backported

tests/test_converter.py
```diff
@@ -551,36 +545,60 @@ def test_operation_transformed_into_qubit_unitary(self, recorder):
         assert np.array_equal(recorder.queue[0].parameters[0], ex.CHGate().to_matrix())
         assert recorder.queue[0].wires == Wires([0, 1])
 
-    def test_quantum_circuit_error_by_passing_wrong_parameters(self, recorder):
-        """Tests the load method for a QuantumCircuit raises a QiskitError,
-        if the wrong type of arguments were passed."""
+    def test_qiskit_gates_to_be_deprecated(self, recorder):
```
josh146, line 548: This is a good idea. This way, we will know when the gates have been removed due to this test failing :)

tests/test_converter.py
```diff
@@ -551,36 +545,60 @@ def test_operation_transformed_into_qubit_unitary(self, recorder):
...
+        """Tests the Qiskit gates taht will be deprecated in an upcoming Qiskit version.
```
josh146, line 549:
```suggestion
        """Tests the Qiskit gates that will be deprecated in an upcoming Qiskit version.
```

josh146 (approved):

PR #114, opened 2020-11-06
No noise model reset

antalszava:
**Context**
Upon device initialization, a noise model can be passed to a Qiskit Aer device by specifying the `noise_model` keyword argument. The noise model is then stored as an attribute, which is initialized to `None` and is further reset to `None` once `dev.reset` is called.

PennyLane core calls `dev.reset` in certain places (e.g., [as part of QNode evaluation](https://github.com/PennyLaneAI/pennylane/blob/0410be66bc5a549ba269ffcd1de3c799cc61a598/pennylane/qnodes/base.py#L821)) which eventually result in "losing" the user defined noise model.

Furthermore, Qiskit Aer devices persist the options that were set for them prior. This affects consecutive device creations: e.g., devices created after a noisy device creation would also have noise be set as an option.

**Changes**
* This PR removes the `noide_model` attribute and moves the `dev.backend.set_options` call directly to `QiskitDevice.__init__` from `QiskitDevice.run`.
* Adds a call to the `clear_options` method for `Aer` backends

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -4,6 +4,7 @@ import numpy as np
 import pennylane as qml
 import pytest
 import qiskit
+import qiskit.providers.aer.noise as noise
 
 from pennylane_qiskit import AerDevice, BasicAerDevice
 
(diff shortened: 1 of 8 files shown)


tests/test_integration.py
```diff
@@ -332,3 +333,23 @@ def circuit_with_inverses_default_qubit(angle):
...
+        noise_model.add_all_qubit_quantum_error(bit_flip, ["rx"])
+
+        dev = qml.device('qiskit.aer', wires=2, noise_model=noise_model)
+
+        @qml.qnode(dev)
+        def circuit():
+            qml.RX(0, wires=[0])
+            return qml.expval(qml.PauliZ(wires=0))
+
+        assert circuit() == -1
```
josh146, line 355: nice test :)

josh146 (approved):
Looks good @antalszava, however it seems that this is causing a large number of tests to fail 🤔 

Are some of the tests implicitly written assuming noise models are being reset?

glassnotes:
I get two failing tests (running 0.13.0 dev for both PennyLane and PennyLane-Qiskit):

```
=========================================================================================================================== short test summary info ===========================================================================================================================
FAILED test_ibmq.py::test_account_error - Failed: DID NOT RAISE <class 'qiskit.providers.ibmq.exceptions.IBMQAccountError'>
FAILED test_sample.py::TestTensorSample::test_hermitian[AerDevice-qasm_simulator-8192-False-1.0-1.0-1.0] - AssertionError: assert False
```

But neither of these seem to involve the noise model (at least not directly).  

In terms of the actual fix - my code that was previously not working due to the noise model being reset is now working as expected; the noise model persists between jobs.

glassnotes (approved):

josh146:
@antalszava not sure if I missed it, but don't forget to update the changelog!

antalszava:
@josh146 thanks for the reminder :) Added it in #116.

PR #116, opened 2020-11-13
Qiskit v0.23.1 fixes

antalszava:
**Context**

As per the Qiskit `v0.23.1` release, the backend object returned using the `get_backend` function is now a copy each time:

```python
import qiskit

a = qiskit.Aer.get_backend('qasm_simulator')
b = qiskit.Aer.get_backend('qasm_simulator')

print(a is b)

False
```
This used to not be the case and hence the `QiskitDevice.backend` property got the reference to the same backend object each time.

This change can be perceived by checking the latest PennyLane plugin tests run that use Qiskit v0.23.1:
https://github.com/PennyLaneAI/plugin-test-matrix/runs/1393563190?check_suite_focus=true#step:8:56

**Changes**

* Creates the `QiskitDevice._backend` private object to store a single backend object throughout the execution. This can then be used to persist backend options.
* Pins the requirement to Qiskit `v0.23.1`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 
 This release contains contributions from (in alphabetical order):
 
-Antal Száva
+Olivia Di Matteo, Josh Izaac, Antal Száva
 
 ---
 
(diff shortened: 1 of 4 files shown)


josh146 (approved):
Thanks for catching and fixing this Antal! 🏆

Issue #119, opened 2020-11-30
Showcase demos using PennyLane-Qiskit on main docs page


glassnotes:
The bottom of the main documentation page has a section for [tutorials](https://pennylaneqiskit.readthedocs.io/en/latest/#tutorials) with a few links, and instructions for filtering the demo page. It would be great to enhance this by adding some explicit links to the demos that make extensive use of PL-Qiskit.

Two ways we can do this are:
 - a row of cards, like the [basic](https://pennylane.ai/qml/demos_basics.html)/research demo pages
 - a carousel, like the [featured demos](https://pennylane.ai/qml/demonstrations.html) section of the demo page

I'm inclined to go with the cards, since there are currently only 1-2 demos that use the plugin. The carousel will look nicer in the future once are more than, e.g., one row's worth of cards.

PR #120, opened 2020-12-04
Add demo showcase

glassnotes:
As discussed in #119, this PR contains a framework for adding demo cards to the bottom of the main docs page. This should help users find relevant demos more quickly.

Currently there are only two demos using PL-Qiskit, one of which has not been formally released. Not having a full row of cards looks a little strange, so it might be better to wait to merge this until there are at least 3 demos. That said, I think this is ready for a second pair of eyes, and if people like the format it's something we can add to the docs of the other plugins as well. 

TODO:
 - [x] fix CSS for the cards 
 - [x] get the tooltips to display


doc/index.rst
```diff
@@ -74,10 +74,30 @@ and simply replace ``'default.qubit'`` with any of the available Qiskit devices,
...
-`demos <https://pennylane.ai/qml/demonstrations.html>`_.
+A few of the PennyLane demos make extensive use of the plugin:
+
+.. raw:: html
+
+    <div class="row">
+
+.. demogalleryitem::
+    :name: Ensemble classification with Forest and Qiskit devices
+    :figure: _static/ensemble_diagram.png
```
josh146, line 85: Out of curiosity, does it work if we include the URL to the figure? E.g., https://pennylane.ai/qml/_images/ensemble_diagram1.png?

This avoids needing to copy all thumbnails into the plugin docs, but I'm not sure if it will work---the `demogalleryitem` directive might need to be updated for this.

doc/index.rst
```diff
@@ -74,10 +74,30 @@ and simply replace ``'default.qubit'`` with any of the available Qiskit devices,
 You can also try to run tutorials, such as the qubit rotation tutorial, on actual quantum hardware by
 using the ``'qiskit.ibmq'`` device.
```
josh146, line 75: What do you think of the existing text here (and above)?

Part of me is wondering if we should remove it, and simply have the demo cards and a link to the demonstrations+filter?

doc/directives.py
```diff
@@ -87,3 +89,89 @@ def run(self):
         thumb = nodes.paragraph()
         self.state.nested_parse(thumbnail, self.content_offset, thumb)
         return [thumb]
+
+
+DEMO_GALLERY_TEMPLATE = """
+.. raw:: html
+
+   <a href={link} data-toggle="tooltip" title="{tooltip}">
+    <div class="card" style="width: 15rem; height: 19rem; margin: 10px;">
```
josh146, line 98: If you increase the margin to 20px, it doesn't look that bad with only two cards:

image

josh146 (approved):
Looks good from my end! Just minor suggestions regarding styling and images

doc/index.rst
glassnotes: Great idea. I adjusted the directive and it works no problem. We can't do it yet for the QV one though since it's not in the gallery.

antalszava (approved):
This looks awesome! :100:

doc/index.rst
glassnotes: I tried adjusting the text, let me know what you think - I actually removed the link to the filter since we're going to showcase those same ones here anyways.

josh146: Oh yeah looks way better!

josh146:
@glassnotes is this okay to be merged?

glassnotes:
@josh146 if you think it looks okay we can merge it! Are there other plugins for which this would be a good addition? (e.g., PL-Forest or PL-cirq?)

josh146:
Yeah it looks great to me! I would go for Cirq next (we can add the qsim demo, the noisy demo, and a few others maybe?)

glassnotes:
Sure, sounds like a plan! Will merge this in, then.

PR #124, opened 2021-01-25
update U gates

albi3ro:
In the PennyLane device tests, `test_four_qubit_random_circuit` was raising error messages.

The Qiskit `U` gate has been deprecated in favour of individual `U1Gate`, `U2Gate`, `U3Gate`, and `UGate` (which is just `U3Gate`).

This pull requests updates plugin to comply with this Qiskit change.

### DIFF
--- a/pennylane_qiskit/qiskit_device.py
+++ b/pennylane_qiskit/qiskit_device.py
@@ -33,25 +33,6 @@ from pennylane import QubitDevice, DeviceError
 from ._version import __version__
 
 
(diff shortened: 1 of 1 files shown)


antalszava (approved):
Looks good to me, thanks for catching this! 💯 😊 Oh it seems that we don't have explicitly tests for applying `U1`, `U2`, `U3` in this plugin, so that's why it didn't cause problems when the test suite was run 🤔 😕

PR #127, opened 2021-02-01
Get local tests passing

albi3ro:
This pull request aims to update the pennylane-qiskit plugin to the most recent master of pennylane, and have all local tests passing.

**Changes to the Plugin Code**
The `load` function in `converter.py` instead binds variables with `requires_grad=False`, instead of binding non-`qml.Variable` objects.

**Changes to the tests**
Variables passed to the device are now Pennylane `tensor` objects.  The plugin unwraps them before passing them into qiskit.


albi3ro:
In `converter.py`,  variables with `requires_grad=False` get bound to the circuit.  All others get entered to the circuit during the construction of the pennylane operations.

When we are testing qiskit errors that arise things get bound, we need to test on tensors with `requires_grad=False`, else they won't get bound.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 qiskit>=0.23.1
-pennylane>=0.13.0
+pennylane>=git+https://github.com/PennyLaneAI/pennylane.git
```
josh146, line 2:
```suggestion
git+https://github.com/PennyLaneAI/pennylane.git
```

pennylane_qiskit/converter.py
```diff
@@ -61,12 +61,11 @@ def _extract_variable_refs(params: Dict[Parameter, Any]) -> Dict[Parameter, qml.
...
-
-            if isinstance(v, qml.variable.Variable):
+            if getattr(v, "requires_grad", True):
+                # Values can be arrays of size 1, need to extract the Python scalar
+                # (this can happen e.g. when indexing into a PennyLane numpy array)
+                if isinstance(v, np.ndarray):
+                    v = v.item()
                 variable_refs[k] = v
 
     return variable_refs  # map qiskit parameters to PennyLane differentiable Variables.
```
antalszava, line 71: Suggest changing ` map qiskit parameters to PennyLane differentiable Variables.` -> ` map qiskit parameters to PennyLane differentiable parameters.`

CHANGELOG.md
```diff
@@ -8,10 +8,15 @@
 
 ### Bug fixes
 
+* When loading Qiskit circuits to PennyLane templates using `load` in `converter.py`, parameters with `requires_grad=False` are bound to the circuit.  The old version bound objects that were not PennyLane `Variable`'s, but that object class is now deprecated.
```
antalszava, line 11: Maybe this description could be shortened and rephrased a bit? It's worth noting that not only those parameters with `requires_grad=False` can be bound to the circuit. The `_extract_variable_refs` function actually tries to gather the values for each trainable parameter (`requires_grad=True`) such that the value of those parameters can be bound to the circuit.

pennylane_qiskit/converter.py
```diff
@@ -61,12 +61,11 @@ def _extract_variable_refs(params: Dict[Parameter, Any]) -> Dict[Parameter, qml.
     if params is not None:
         for k, v in params.items():
 
-            # Values can be arrays of size 1, need to extract the Python scalar
-            # (this can happen e.g. when indexing into a PennyLane numpy array)
-            if isinstance(v, np.ndarray):
-                v = v.item()
-
-            if isinstance(v, qml.variable.Variable):
+            if getattr(v, "requires_grad", True):
```
antalszava, line 64: It's a great appraoch to use `requires_grad`! One thing to note is that having `getattr(v, "requires_grad", True)` means that if `v` doesn't have a `requires_grad` attribute, we still perform the following operations. Is that what we'd want here? 🤔 

Also, it might be worth considering how we could support backwards compatibility so long as there's an option to use the old core.

antalszava (commented):
Looking good @albi3ro! 💯 Left a couple of suggestions.

Could we also only format the `pennylane` directory using `black`? Having many changes hides the real updates of the PR.

pennylane_qiskit/converter.py
albi3ro: I think currently the policy is "Everything not explicitly marked with `requires_grad=False` is trainable"

antalszava (approved):
Looks good to me! 💯

antalszava:
There remains a [converter test case](https://github.com/PennyLaneAI/pennylane-qiskit/blob/af673faaabacbd677ab105bb409481342224e128/tests/test_converter.py#L242) that uses `Variable` objects explicitly.

We could either update that in this PR or when we are completely deprecating `Variable` objects from PennyLane core.

pennylane_qiskit/converter.py
josh146: Yep, this is correct, but it's annoying so I'm hoping to switch it soon!

josh146 (approved):

PR #133, opened 2021-04-13
Using hash instead of tuple

albi3ro:
New qiskit is deprecating getting the register name from a qubit.  We therefore need a better way of uniquely identifying qubits and their `(register, index)` combination.   Qiskit `Bit`'s have their own hash defined on the tuple of register name and index, and so will uniquely define each bit.

Though these bits will be harder to debug and read, the hash will be sufficiently unique per qubit to allow matching up qiskit qubits and PL wires.

### DIFF
--- a/doc/devices/aer.rst
+++ b/doc/devices/aer.rst
@@ -63,7 +63,7 @@ Backend Methods and Options
 
 This ``AerSimulator`` backend has several available methods, which
 can be passed via the ``method`` keyword argument. For example
-``'automatica'``, ``'statevector'``, and ``'unitary'``.
+``'automatic'``, ``'statevector'``, and ``'unitary'``.
 
 .. code-block:: python
 
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -171,11 +171,7 @@ def load(quantum_circuit: QuantumCircuit):
         qc = _check_circuit_and_bind_parameters(quantum_circuit, params, var_ref_map)
 
         # Wires from a qiskit circuit are unique w.r.t. a register name and a qubit index
-        qc_wires = [
-            (reg.name, qubit_ind)
-            for reg in quantum_circuit.qregs
-            for qubit_ind, _ in enumerate(reg)
-        ]
+        qc_wires = [hash(q) for q in qc.qubits]
 
         wire_map = map_wires(wires, qc_wires)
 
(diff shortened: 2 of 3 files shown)


albi3ro:
I have no idea why one completely unrelated test would fail.  Seems quite strange.

albi3ro:
And now it's passing? Maybe a bad test?

doc/devices/aer.rst
```diff
@@ -63,7 +63,7 @@ Backend Methods and Options
 
 This ``AerSimulator`` backend has several available methods, which
 can be passed via the ``method`` keyword argument. For example
-``'automatica'``, ``'statevector'``, and ``'unitary'``.
+``'automatic'``, ``'statevector'``, and ``'unitary'``.
```
josh146, line 66: Oops, nice catch!

josh146 (commented):
I might defer to @antalszava for the full review here, as I am much less familiar with the conversion code :)

pennylane_qiskit/converter.py
```diff
@@ -171,11 +171,7 @@ def _function(params: dict = None, wires: list = None):
         qc = _check_circuit_and_bind_parameters(quantum_circuit, params, var_ref_map)
 
         # Wires from a qiskit circuit are unique w.r.t. a register name and a qubit index
```
antalszava, line 173:
```suggestion
        # Wires from a qiskit circuit have unique IDs, so their hash is unique too
```

tests/test_converter.py
```diff
@@ -716,39 +716,40 @@ def test_map_wires_instantiate_quantum_circuit_with_registers(self, recorder):
...
         """Tests the map_wires function for wires of non-standard order."""
 
         wires = [1, 2, 0]
         qc = QuantumCircuit(3)
-        qc_wires = [(q.register.name, q.index) for q in qc.qubits]
+        qc_wires = [hash(q) for q in qc.qubits]
 
         mapped_wires = map_wires(qc_wires, wires)
 
+        for q in qc.qubits:
```
antalszava, line 737: Any reason why these asserts were placed before lines 740-741? (they came after these lines originally)

antalszava (approved):
Looks great! :100: Awesome solution :heart_eyes: Thank you :)

tests/test_converter.py
albi3ro: just keeping single line asserts together and not interrupted. Looks nicer?

Issue #134, opened 2021-04-13
Order of inputs to `converter.py`'s `map_wires` function


albi3ro:
The [`map_wires`](https://github.com/PennyLaneAI/pennylane-qiskit/blob/b646dab624e9953fad4a29e369458c9632d42dc8/pennylane_qiskit/converter.py#L103) function takes inputs in the order `values, keys` and returns a dictionary of mapping `keys` to `values`.

My brain does not like this.  We should instead have `map_wires` take inputs of the form `keys, values`.  Then the code will be easier to read and remember.

PR #135, opened 2021-04-16
switch input order for map_wires

albi3ro:
Issue #134 

Previously in `converter.py`, the function `map_wires` took in arguments of the order `values, keys` to create a dictionary from `keys` to `values`.

This was conterintuitive, so I have switched the order.


antalszava (approved):
:100:

PR #137, opened 2021-05-27
Add a projector observable to QiskitDevice

wongwsvincent:
Context:
Adding a projector observable to QiskitDevice.

Description of the Change:
Adding ```Projector``` to the list of observables compatible with QiskitDevice in ```qiskit_device.py```.

Related GitHub PR:
Work coherently with PR [#1356](https://github.com/PennyLaneAI/pennylane/pull/1356)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ### New features since last release
 
-* Added Projector observable to QiskitDevice.
+* Added support for the new `qml.Projector` observable in
+  PennyLane v0.16 to the Qiskit devices.
   [(#137)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/137)
 
 ### Breaking changes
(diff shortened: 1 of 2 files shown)


wongwsvincent:
@josh146 , I added docstrings here because I saw CodeFactor in the github workflow was complaining about not having docstrings under the functions. I'm not sure why the docstrings weren't inherited from the parent class, and I'm not familiar enough with Sphinx to understand why. The flagged issues can be found [here](https://github.com/PennyLaneAI/pennylane-qiskit/pull/137/checks?check_run_id=2682260099).

Could you give me a bit of help here?

josh146:
> I saw CodeFactor in the github workflow was complaining about not having docstrings under the functions

Thanks for flagging this @wongwsvincent --- it appears that Codefactor is incorrect here, and these particular methods do not need docstrings.

Feel free to remove them, even if codefactor complains about the missing docstrings, I will be able to overwrite those :)

wongwsvincent:
Thanks for clarifying, @josh146 ! I have removed the docstrings.

josh146 (approved):
Thanks @wongwsvincent! Good to go now 💯

Issue #129, opened 2021-02-02
Support loading Parameter Expressions


albi3ro:
Currently, the `load` function to convert a Qiskit circuit cannot convert Qiskit `ParameterExpression`'s with more than a simple, unaltered `Parameter`.  See error raised at [converter.py#L205](https://github.com/PennyLaneAI/pennylane-qiskit/blob/059964d097e14e5c1b36d231ec5b4330a4e345f5/pennylane_qiskit/converter.py#L205)

PennyLane now, as of v0.14.0, supports classical processing of arguments within a circuit.  With this change, PennyLane-Qiskit should update their support of `ParameterExpression`'s.  

We should be able to do the following:

```python
from qiskit import QuantumCircuit
from qiskit.circuit import Parameter
from pennylane_qiskit import load

a = Parameter("a")
b = Parameter("b")

qc = QuantumCircuit(1,1)
qc.rz(a*b, [0])

circ = load(qc)(params={a: 0.1, b: 0.1})
```

## Potential Solution

`ParameterExpression`'s are based on SymPy, and thus we can use SymPy functions with them.  

```python
from sympy import lambdify

expr = a*b
f = lambdify([a,b], expr, 'numpy')
f(0.1, 0.1)
```

The resulting function is differentiable in pennylane:

```
import pennylane as qml

qml.grad(f)(1.0, 1.0)
```

antalszava:
@albi3ro has this been resolved?

lillian542:
This is supported now - for example, we can do:

```
from qiskit import QuantumCircuit
from qiskit.circuit import Parameter
from pennylane_qiskit import load

dev = qml.device("default.qubit")

a = Parameter("a")
b = Parameter("b")

qc = QuantumCircuit(1,1)
qc.rx(a*b, [0])

@qml.qnode(dev)
def circuit(x, y):
    load(qc)(params={a: x, b: y})
    return qml.expval(qml.PauliZ(0))
```
and run

```
>>> circuit(0, np.pi)
tensor(1., requires_grad=True)

>>> circuit(1, np.pi)
tensor(-1., requires_grad=True)

>>> circuit(0.5, np.pi)
tensor(2.22044605e-16, requires_grad=True)
```
    
The circuit is differentiable.

PR #139, opened 2021-06-14
Parameter expressions

albi3ro:
Closes Issue #129 

This improvement allows the plugin to load circuits with more general parameter expressions.

For example,

```python
from qiskit.circuit import Parameter
from pennylane_qiskit import load

a = Parameter("a")
b = Parameter("b")

qc = QuantumCircuit(1,1)
qc.rz(a*b, [0])

dev  = qml.device('default.qubit', wires=1)

@qml.qnode(dev)
def circ(x, y):
    load(qc)(params={a: x, b: y}, wires=(0,))
````

Currently, the above code would raise an error.

I also add `sympy` to the requirements, as it is explicitly used.  Qiskit already has sympy as a dependency, so this does not add any cost to the requirements.

### DIFF
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -22,6 +22,7 @@ import numpy as np
 from qiskit import QuantumCircuit
 from qiskit.circuit import Parameter, ParameterExpression
 from qiskit.exceptions import QiskitError
+from sympy import lambdify
 
 import pennylane as qml
 import pennylane.ops.qubit as pennylane_ops
@@ -176,11 +177,11 @@ def load(quantum_circuit: QuantumCircuit):
         wire_map = map_wires(qc_wires, wires)
 
         # Processing the dictionary of parameters passed
-        for op in qc.data:
+        for op, qargs, cargs in qc.data:
 
-            instruction_name = op[0].__class__.__name__
+            instruction_name = op.__class__.__name__
 
-            operation_wires = [wire_map[hash(qubit)] for qubit in op[1]]
+            operation_wires = [wire_map[hash(qubit)] for qubit in qargs]
 
             # New Qiskit gates that are not natively supported by PL (identical
             # gates exist with a different name)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: tests/test_converter.py [hunk 1]
```diff
@@ -227,37 +250,196 @@ def test_parameter_was_not_bound(self, recorder):
             with recorder:
                 quantum_circuit(params={})
 
-    def test_invalid_parameter_expression(self, recorder):
```
albi3ro, line 230: If you're wondering, I also moved some tests around to try and organize the file better.

tests/test_converter.py
```diff
@@ -1063,3 +1075,38 @@ def circuit(params):
         ]
 
         assert np.allclose(res, expected, **tol)
+
+    def test_parameterexpression(self):
```
albi3ro, line 1079: This is a new test.

tests/test_converter.py
```diff
@@ -144,6 +144,29 @@ def test_quantum_circuit_with_gate_requiring_multiple_parameters(self, recorder)
         assert recorder.queue[0].parameters == [0.5, 0.3, 0.1]
         assert recorder.queue[0].wires == Wires([0])
 
+    def test_longer_parameter_expression(self):
```
albi3ro, line 147: this is a new test.

requirements.txt
```diff
@@ -1,5 +1,6 @@
 qiskit>=0.25
 pennylane>=0.15
 numpy
+sympy
```
antalszava, line 4: Usually, introducing a new dependency would be good to be considered, but in this case, Qiskit already depends on `sympy`. Worth noting that we can do this "without a cost" maybe in the PR description.

pennylane_qiskit/converter.py
```diff
@@ -192,30 +193,28 @@ def _function(params: dict = None, wires: list = None):
...
-                                    "expressions containing differentiable parameters as operation "
-                                    "arguments".format(instruction_name, p)
-                                )
-
-                            param = min(p.parameters)
-                            parameters.append(var_ref_map.get(param))
+                        if p.parameters:  # non-empty set = has unbound parameters
+                            ordered_params = tuple(p.parameters)
+
+                            f = lambdify(ordered_params, p._symbol_expr, modules=qml.numpy)
```
antalszava, line 205: I wonder if this has any performance implications. :thinking: In Strawberry Fields, there was a case where using `lambdify` turned out to not be the best choice:
https://github.com/XanaduAI/strawberryfields/pull/579

tests/test_converter.py
antalszava, line 1079:
```suggestion
    def test_parameter_expression(self):
```

tests/test_converter.py
```diff
@@ -1063,3 +1075,38 @@ def circuit(params):
         ]
 
         assert np.allclose(res, expected, **tol)
+
+    def test_parameterexpression(self):
+        """Tests for functions of parameters passed to qiskit gate"""
```
antalszava, line 1080:
```suggestion
        """Tests the output and the gradient of a QNode that contains loaded Qiskit gates taking functions of parameters as argument"""
```

antalszava (commented):
Overall looks good! :tada: This is a nice improvement to have. Had some comments, would just be curious about the performance changes of using `lambdify` (if any at all).

pennylane_qiskit/converter.py
albi3ro: Acoording to [the sympy docs](https://docs.sympy.org/latest/modules/numeric-computation.html), `lambdify` is more efficient than `evalf`.  Also, it allows us to use `qml.numpy` as the namespace for conversion.  Using our wrapped numpy allows the expression to be differentiable.

I don't think this conversion is ever going to be very much of a bother compared to other bottlenecks.  Qiskit does not seem very supportive of users creating elaborate expressions.

antalszava: Could we keep the condition as before and use the solution with lambdify only `if len(p.parameters) > 1`? That way we'll make use of it fewer times.

antalszava (approved):
Looks good to me :+1: Had a minor question on limiting the use of `lambdify`, but that would probably be a minor change to add if accepted.

pennylane_qiskit/converter.py
albi3ro: What is the expression is something like `ParameterExpression(2*a)`.  Then there's only one parameter, `a`, but we can't just use `a`'s value in place of the expression.  This is a case that would have fallen through the cracks before.  If there's a viable alternative that's faster, I'm all for it, but I think this is the only way to support necessary functionality.

requirements.txt
albi3ro: do you mean in the changelog, or just in the PR?

PR #141, opened 2021-06-18
Pin to PennyLane 0.16

antalszava:
Pins the requirement to `pennylane>=0.16` before releasing the plugin.


josh146 (approved):
Looks good to me @antalszava!

PR #142, opened 2021-06-22
Incrementing the version number to `v0.17.0-dev`

rmoyard:


antalszava:
Thank you @rmoyard! It seems that in this plugin we have a separate `version` and `plugin_version` fields:

https://github.com/PennyLaneAI/pennylane-qiskit/blob/03c985c2fb07deafa88ca8b40afb977ac494b3c1/pennylane_qiskit/qiskit_device.py#L89

Maybe it's good to change to `version =  __version__` here too, so that it's dynamically set.

antalszava (approved):
Thank you! :tada:

PR #144, opened 2021-07-20
Update `qml.QubitUnitary` error message in test

antalszava:
This change was required due to an updated error message in PennyLane core from https://github.com/PennyLaneAI/pennylane/pull/1439


antalszava:
[ch7574]

pennylane_qiskit/qiskit_device.py
```diff
@@ -256,7 +256,6 @@ def apply_operations(self, operations):
 
             mapped_operation = self._operation_map[operation]
 
-            self.qubit_unitary_check(operation, par, device_wires)
```
antalszava, line 259: QubitUnitary check is in PennyLane core -> remove from device

Issue #145, opened 2021-08-03
`UserWarning: verbose is not a recognized runtime option and may be ignored by the backend`


antalszava:
When submitting circuits to IBMQ the following `UserWarning` comes up:
```
/home/pennylane-qiskit/pennylane_qiskit/qiskit_device.py:304: UserWarning: verbose is not a recognized runtime option and may be ignored by the backend.
  self._current_job = self.backend.run(qcirc, shots=self.shots, **self.run_args)
```

This can be reproduced for example when running the transfer learning QML demo. See also this comment on the [PennyLane forum](https://discuss.pennylane.ai/t/quantum-transfer-learning-code-mari-et-al-2019-ibmqdevice-endless-execution/325/21?u=antalszava).

PR #151, opened 2021-08-25
Remove adding a verbose option to the run arguments

antalszava:
**Context**

In `QiskitDevice`, a `verbose` option is being added by default to the keyword arguments to be passed to the backend. This has caused warnings when submitting jobs to IBMQ.

The lines that set the verbose to false have been added a while back in https://github.com/PennyLaneAI/pennylane-qiskit/pull/44. Likely, this used to be a valid backend option, however, at the moment there Qiskit doesn't seem to support it.

**Changes**

Removes the addition of the verbose option.

**Related issues**

#145

### DIFF
--- a/pennylane_qiskit/qiskit_device.py
+++ b/pennylane_qiskit/qiskit_device.py
@@ -125,9 +125,6 @@ class QiskitDevice(QubitDevice, abc.ABC):
 
         self._backend = None
 
-        if "verbose" not in kwargs:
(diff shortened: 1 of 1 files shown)


antalszava:
[ch7908]

albi3ro (approved):
Sounds good.

Does this need a changelog entry?

antalszava:
@albi3ro good catch! Added one. Thank you

PR #152, opened 2021-09-15
increment version to 0.18.0

albi3ro:


Jaybsoni:
Do we need to update the setup.py file to include the new version # ?

Jaybsoni (changes requested):
Looks good to me, not sure if we need to update the setup.py file, aside from that I think its good to go!

pennylane_qiskit/qiskit_device.py
```diff
@@ -85,7 +85,7 @@ class QiskitDevice(QubitDevice, abc.ABC):
             to simulate a device compliant circuit, you can specify a backend here.
     """
     name = "Qiskit PennyLane plugin"
-    pennylane_requires = ">=0.17.0"
+    pennylane_requires = ">=0.18.0"
```
antalszava, line 88:
```suggestion
    pennylane_requires = ">=0.17.0"
```

requirements.txt
```diff
@@ -1,5 +1,5 @@
 qiskit>=0.25
-pennylane>=0.17
+pennylane>=0.18
```
antalszava, line 2:
```suggestion
pennylane>=0.17
```

antalszava (changes requested):
Hi @albi3ro, @Jaybsoni, we only every increment the PennyLane version number in the `requirements.txt` and `setup.py` files and the `pennylane_requires` attribute in the device if there's a main feature in the new version of PennyLane (`v0.18.0`) that the plugin depends on.

This is not the case here, as the only change since the last release in the plugin was removing a deprecated keyword argument.

Jaybsoni (approved):
Ok, so all that was changed was the changelog and the _version.py file. 
Thanks for the review @antalszava!

antalszava (approved):

Issue #113, opened 2020-10-20
Update plugin to natively support Ugate, Pgate, and SXgate


josh146:
* The `Ugate` is the new Qiskit label for the `U3gate`

* The `Pgate` corresponds to the PennyLane `PhaseShift` operation

* The `SXgate` (sqrtX gate) is new, and currently does not have a PennyLane native equivalent.

* Qiskit has deprecated the `[u1, u2, u3]` basis in favour of the new `[p, sx]` basis.

Once the above gates are natively available in PennyLane, we should update the plugin to use the newly available gates.

charmerDark:
Hey! Is this issue still open? The `SXgate` is now [supported](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.SX.html) and cant be used in the plugin too right? I understand `Ugate` and `Pgate`  have already been updated in the v0.23 update.

josh146:
Hi @charmerDark, having a look at the Qiskit device's supported operations,
https://github.com/PennyLaneAI/pennylane-qiskit/blob/57934622ee0531212e032c6dd24271c5fd1b4d6d/pennylane_qiskit/qiskit_device.py#L36
it appears that `SX` is not yet listed. Further, `ex.PhaseGate` and `ex.U3Gate` currently in used by the plugin might be deprecated by Qiskit and could be in need of an update!

So I would say that this issue is still valid, yes.

charmerDark:
> Hey! Is this issue still open? The `SXgate` is now [supported](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.SX.html) and cant be used in the plugin too right? I understand `Ugate` and `Pgate` have already been updated in the v0.23 update.

I'm sorry..I meant now that `SX` is available on pennylane, it could be added to the list of supported operations. the 'cant' was a typo :sweat_smile:

josh146:
Ah I see, yes definitely @charmerDark!

PR #158, opened 2021-11-11
Add native support for the SX gate

antalszava:
Adds an entry to the operation map such that the SX gate is natively supported by the plugin.

Related issue: https://github.com/PennyLaneAI/pennylane/issues/1891, #113

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ### Improvements
 
+* Added support for the `qml.SX` operation to the Qiskit devices.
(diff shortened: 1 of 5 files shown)


rmoyard (commented):
@antalszava Thank you very much for adding SX, it looks great! I suggest you add a test in `test_inverses.py` on line 25 `("SX", 0)` 💯

antalszava:
Thanks @rmoyard for catching that! Now it should be updated.

rmoyard (approved):
Perfect, thank you for the changes!

PR #159, opened 2021-11-18
Update the batch execution tests to use the new QNode

antalszava:


rmoyard (approved):
It looks good to me, thanks for the change 💯

rmoyard (approved):
Thanks again !

Issue #109, opened 2020-10-07
Add support for returning job execution times


antalszava:
We would like to allow retrieving the time taken for jobs when using IBMQ. This is at the moment not available, as results are obtained from the specific `Job` objects (e.g. counts) and neither the job objects nor result objects are stored explicitly. An approach could be to store internally the information contained in [`job.time_per_step`](https://qiskit.org/documentation/stubs/qiskit.providers.ibmq.job.IBMQJob.time_per_step.html) and `result.time_taken`.

For optimization tasks, it can be desirable to store multiple job execution times and not just the most recent.

Further details can be found in [this request](https://discuss.pennylane.ai/t/how-do-i-get-execution-time-of-jobs-not-transpile-queue-python-etc-time-on-qiskit-hardware/613).

sagarpahwa:
Hi @antalszava, can I try to look into this one? Might need some help though.
Thanks in advance!

antalszava:
Hi @sagarpahwa, that's great, thanks for showing interest there!

Definitely! This could be coming up after concluding with #108. Once merged we could explore what could be a nice approach here.

sagarpahwa:
Okay, will start exploring the existing code meanwhile.

Thanks @antalszava .

antalszava:
Hi @sagarpahwa, feel free to let us know if you'd need some help in terms of design or anything here

sagarpahwa:
Oh, my bad. Forgot to provide update here. exams are going on right now, won't be able to work till 2nd November.

If possible, can I start this post that?

antalszava:
Hi @sagarpahwa, no worries! Sure, starting then works fine too.

Good luck with your exams! :slightly_smiling_face:

albi3ro:
I'd like to revive this issue, as it ties in well with the addition of the [Tracker](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.Tracker.html) in PennyLane v0.17.0 .

PR #160, opened 2021-11-19
Add job_time

KetpuntoG:


antalszava:
[sc-11508]

pennylane_qiskit/qiskit_device.py
```diff
@@ -397,6 +396,9 @@ def batch_execute(self, circuits):
         self._current_job = self.backend.run(compiled_circuits, shots=self.shots, **self.run_args)
         result = self._current_job.result()
 
+        if self.tracker.active:
```
rmoyard, line 399: Could we move this to line 416?

```
if self.tracker.active:
            self._track_run()
            self.tracker.update(batches=1, batch_len=len(circuits))
            self.tracker.record()
```

rmoyard (commented):
@KetpuntoG Thanks for the PR, it is working as expected but I don't like the empty function for `qiskit_device`. Could we delete `_track_run` in the `qiskit_device` and redefine `batch_execute` in the `IBMQDevice`. Something like

```
def batch_execute(self, circuits):
         super().(circuits)
         if self.tracker.active:
                  self._track_run()
```

should be working, let me know if it is the case !

rmoyard:
@KetpuntoG Great changes, can you just add it to the changelog 💯

KetpuntoG:
Now I'll add to the chagelog :) I find it strange that I pass the tests in python 3.7 and 3.9 but not in 3.8, do you know why it could be @rmoyard ?

KetpuntoG:
#109

rmoyard (approved):
Thanks @KetpuntoG it looks good to me 👍

PR #161, opened 2021-11-22
Change token for IBMQ tests

rmoyard:


antalszava (approved):
Thank you for catching this! :pray: The change needed for the `batch_execute` test is akin to the updates in #159: we call `batch_execute` once for the QNode evaluation (1 tape) and once to compute the gradient (4 tapes with shifted params).

rmoyard:
@KetpuntoG caught the problem 💯

PR #162, opened 2021-11-29
Add the `qml.Identity` operation to PennyLane-Qiskit

### LINKED ISSUE #153: Better support for the Identity gate
Currently a Qiskit `IGate` is converted to a `qml.ops.ops.QubitUnitary`.
The reason is that `qml.ops.Identity` is only supported in PennyLane as an observable. Likely, in the future it will be supported as a an operation too. There is already an issue for this https://github.com/PennyLaneAI/pennylane/issues/1632 .

So, as soon as https://github.com/PennyLaneAI/pennylane/issues/1632 will be fixed, it would be nice to properly support the `Identity` gate also in this plug-in. It should be enough to add a new line in this conversion map:
 https://github.com/PennyLaneAI/pennylane-qiskit/blob/043c544917da0c9b9f200d8949fb56bd76ece5fb/pennylane_qiskit/qiskit_device.py#L36

dime10:
Adds support for the `qml.Identity` operation to the Qiskit plugin, which was previously treated as a `qml.QubitUnitary`.

Support is added for both directions, that is Qiskit backends can handle PennyLane's `Identity` operation, as in this example:
```python
import pennylane as qml
from pennylane_qiskit import BasicAerDevice

dev = BasicAerDevice(1)

@qml.qnode(dev)
def circuit():
    qml.Identity(wires=0)
    return qml.probs(wires=0)

circuit()
```

And the plugin converters correctly translate a Qiskit `IGate` to PennyLane's `Idendity` operation:
```python
import pennylane as qml
from pennylane_qiskit import load

recorder = qml.tape.OperationRecorder()

circ = QuantumCircuit(1)
circ.id(0)

with recorder:
    load(c)()

assert recorder.queue[0].name == "Identity"
```

Closes #153

### DIFF
--- a/pennylane_qiskit/qiskit_device.py
+++ b/pennylane_qiskit/qiskit_device.py
@@ -48,6 +48,7 @@ QISKIT_OPERATION_MAP = {
     "S": ex.SGate,
     "T": ex.TGate,
     "SX": ex.SXGate,
+    "Identity": ex.IGate,
     # Adding the following for conversion compatibility
     "CSWAP": ex.CSwapGate,
     "CRX": ex.CRXGate,
--- a/tests/test_apply.py
+++ b/tests/test_apply.py
@@ -42,7 +42,7 @@ crz = lambda theta: np.array(
 )
 
 
-single_qubit_operations = [qml.PauliX, qml.PauliY, qml.PauliZ, qml.Hadamard, qml.S, qml.T, qml.SX]
+single_qubit_operations = [qml.Identity, qml.PauliX, qml.PauliY, qml.PauliZ, qml.Hadamard, qml.S, qml.T, qml.SX]
 
 single_qubit_operations_param = [qml.PhaseShift, qml.RX, qml.RY, qml.RZ]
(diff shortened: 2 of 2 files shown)


dime10:
[sc-9205]

rmoyard (commented):
Hi @dime10 thanks for the PR, I think you need to adapt `load` or add Identity operation to  `pennylane/ops/qubit/__init__.py`, in order to make the converter work !

dime10:
> Hi @dime10 thanks for the PR, I think you need to adapt load or add Identity operation to pennylane/ops/qubit/__init__.py, in order to make the converter work !

Yes that change is coming but since it's a different repo it's not shown here :)

rmoyard (commented):
Hi @dime10, it looks good to me 💯 Just two comments, you could add `Identity` in the test of inverses and also update the changelog.

rmoyard (approved):
Thanks @dime10 !

antalszava:
[sc-12053]

PR #163, opened 2021-12-02
Fix `batch_execute` output type for QNodes with multiple outputs

antalszava:
**Context**

[One of the demonstrations fails ](https://github.com/PennyLaneAI/qml/runs/4389087107?check_suite_focus=true#step:8:2416) due to a change of output type when using the `qiskit.aer` device.

With the latest `batch_execute` addition, this device may output results in a different format than other devices (e.g., `default.qubit`).

```python
from pennylane import numpy as np
import pennylane as qml

dev2 = qml.device("qiskit.aer", wires=4)

@qml.qnode(dev2, interface="torch")
def qnode(params):
    qml.RX(params, wires=0)
    return qml.expval(qml.PauliZ(0)), qml.expval(qml.PauliZ(1)), qml.expval(qml.PauliZ(2))


qnode(0.3)
```
```
[tensor(0.9512, dtype=torch.float64),
 tensor(1., dtype=torch.float64),
 tensor(1., dtype=torch.float64)]
```
whereas it should output
```
tensor([0.9746, 1.0000, 1.0000], dtype=torch.float64)
```
The reason why this causes is an error in the demo is that we're using `torch.stack` in the QNodeCollection class on the results of the underlying QNodes. However, with the latest output type, we may end up having a list in the output, on which `torch.stack` will fail.

Internally, the issue is a missing conversion to `np.array`, that has been done previously when using the default `execute` pipeline.

This can be showcased by comparing the outputs of executing the same tape using the `default.qubit` and the `qiskit.aer` devices:
```python
from pennylane import numpy as np
import pennylane as qml

with qml.tape.QuantumTape() as tape:
    qml.PauliX(wires=0)
    qml.expval(qml.PauliZ(wires=0)), qml.expval(qml.PauliZ(wires=1))

dev1 = qml.device("qiskit.aer", wires=4)
print(dev1.batch_execute([tape]))

dev2 = qml.device("default.qubit", wires=4)

print(dev2.batch_execute([tape]))
```
```
[[-1.0, 1.0]]
[array([-1.,  1.])]
```

**Changes**

Converts the output of `QubitDevice.statistics` to a numpy array to resolve the issue.

### DIFF
--- a/pennylane_qiskit/qiskit_device.py
+++ b/pennylane_qiskit/qiskit_device.py
@@ -409,6 +409,7 @@ class QiskitDevice(QubitDevice, abc.ABC):
                 self._samples = self.generate_samples(circuit_obj)
 
             res = self.statistics(circuit.observables)
+            res = np.asarray(res)
             results.append(res)
 
         if self.tracker.active:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ### New features since last release
 
+* Defined the `QiskitDevice.batch_execute` method, to allow
+  Qiskit backends to run multiple quantum circuits at the same time. This
+  addition allows submitting batches of circuits to IBMQ e.g., when computing
+  gradients internally.
+  [(#156)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/156)
+  [(#163)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/163)
+
 ### Breaking changes
 
 ### Improvements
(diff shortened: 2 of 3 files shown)


pennylane_qiskit/qiskit_device.py
```diff
@@ -409,6 +409,7 @@ def batch_execute(self, circuits):
                 self._samples = self.generate_samples(circuit_obj)
 
             res = self.statistics(circuit.observables)
+            res = np.asarray(res)
```
rmoyard, line 412:
```suggestion
            res = np.asarray(self.statistics(circuit.observables))
```

rmoyard (approved):
Thanks @antalszava I tried different cases and it is working! Just let one comment 💯

pennylane_qiskit/qiskit_device.py
antalszava: Personally, I prefer breaking up statements into multiple lines, such that the smaller steps can be followed easily.

antalszava:
[sc-11994]

PR #167, opened 2021-12-10
Update the tracker used with batch execution

antalszava:


tests/test_integration.py
```diff
@@ -569,3 +569,25 @@ def circuit(x, y):
         # Check that run was called twice: for the partial derivatives and for
         # running the circuit
         assert spy2.call_count == 2
+
+    def test_tracker(self):
+        """Tests the device tracker with batch execution."""
+        dev = qml.device('qiskit.aer', shots=100, wires=3)
+        x = np.array([0.1, 0.2])
```
rmoyard, line 576:
```suggestion
```

I think x is defined later.

rmoyard (commented):
Thanks @antalszava, is that correct that you execute the circuit once for evaluating it and twice for the parameter shift rules?

antalszava:
> is that correct that you execute the circuit once for evaluating it and twice for the parameter shift rules?

Yep. :slightly_smiling_face:

rmoyard (approved):
Thank you @antalszava 💯

PR #169, opened 2021-12-13
Improve token-specified login flow.

TannerRogalsky:
Creating multiple `IBMQDevice`s with credentials initialized from an environment variable tried to `enable_account` while an account already active. This PR fixes that while still allowing new environment tokens to specify new accounts to authenticate against.

### DIFF
--- a/tests/test_ibmq.py
+++ b/tests/test_ibmq.py
@@ -40,6 +40,36 @@ def test_load_from_env_multiple_device(token, monkeypatch):
     assert dev1.provider.credentials.is_ibmq()
     assert dev2.provider.credentials.is_ibmq()
 
+def test_load_from_env_multiple_device_and_token(monkeypatch):
+    mock_provider = "MockProvider"
+    mock_qiskit_device = MockQiskitDeviceInit()
+
+    with monkeypatch.context() as m:
+        m.setattr(ibmq.QiskitDevice, "__init__", mock_qiskit_device.mocked_init)
+
+        creds = []
+        def enable_account(new_creds):
+            creds.append(new_creds)
+        def active_account():
+            if len(creds) != 0:
+                return { "token": creds[-1] }
+        m.setattr(ibmq.IBMQ, "enable_account", enable_account)
+        m.setattr(ibmq.IBMQ, "disable_account", lambda: None)
+        m.setattr(ibmq.IBMQ, "active_account", active_account)
+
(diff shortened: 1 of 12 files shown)


antalszava (changes requested):
Hi @TannerRogalsky, thank you so much for this change! :slightly_smiling_face: :tada: 

It might be worth adding one more test case just to make sure that all cases are covered.

It looks that the IBMQ token used in the device test suite is not exposed when making a PR from a forked PR. We have a mechanism in place to skip tests in such cases, it required a small addition.

As a result, CodeCov will likely not pass because the related tests are being skipped (I'll create an issue related to this).

antalszava:
We could add `# pragma: no cover` to the lines that are not reported to CodeCov and remove those once we've resolved #173.

TannerRogalsky:
@antalszava I think I've got everything that you requested. Let me know if it's not how you envisioned.

tests/test_ibmq.py
```diff
@@ -32,6 +32,43 @@ def test_load_from_env(token, monkeypatch):
     dev = IBMQDevice(wires=1)
     assert dev.provider.credentials.is_ibmq()
 
+def test_load_from_env_multiple_device(token, monkeypatch):
```
antalszava, line 35:
```suggestion
def test_load_from_env_multiple_device(token, monkeypatch):
    """Test creating multiple IBMQ devices when the environment variable
    for the IBMQ token was set.""""
```

tests/test_ibmq.py
```diff
@@ -32,6 +32,43 @@ def test_load_from_env(token, monkeypatch):
...
+    monkeypatch.setenv("IBMQX_TOKEN", token)
+    dev1 = IBMQDevice(wires=1)
+    dev2 = IBMQDevice(wires=1)
+
+    assert dev1.provider.credentials.is_ibmq()
+    assert dev2.provider.credentials.is_ibmq()
+
+def test_load_from_env_multiple_device_and_token(monkeypatch):
```
antalszava, line 43:
```suggestion
def test_load_from_env_multiple_device_and_token(monkeypatch):
    """Test creating multiple devices when the different tokens are defined
    using an environment variable.""""
```

antalszava (approved):
Looks good to me! :tada: Thank you for this addition @TannerRogalsky :slightly_smiling_face:

TannerRogalsky:
Thank YOU for suggestions and guidance!

PR #174, opened 2021-12-17
Add and change some docstrings for a better quality

rmoyard:
[sc-9622]

### DIFF
--- a/pennylane_qiskit/__init__.py
+++ b/pennylane_qiskit/__init__.py
@@ -14,7 +14,8 @@
 """Top level PennyLane-qiskit module"""
 
 from ._version import __version__
+from .qiskit_device import QiskitDevice
 from .aer import AerDevice
 from .basic_aer import BasicAerDevice
-from .converter import load, load_qasm, load_qasm_from_file
 from .ibmq import IBMQDevice
+from .converter import load, load_qasm, load_qasm_from_file
--- a/pennylane_qiskit/__init__.py
+++ b/pennylane_qiskit/__init__.py
@@ -14,7 +14,6 @@
 """Top level PennyLane-qiskit module"""
 
 from ._version import __version__
-from .qiskit_device import QiskitDevice
 from .aer import AerDevice
 from .basic_aer import BasicAerDevice
 from .ibmq import IBMQDevice
--- a/pennylane_qiskit/__init__.py
+++ b/pennylane_qiskit/__init__.py
@@ -18,3 +18,6 @@ from .aer import AerDevice
 from .basic_aer import BasicAerDevice
 from .ibmq import IBMQDevice
 from .converter import load, load_qasm, load_qasm_from_file
+from .runtime_devices import IBMQCircuitRunnerDevice
+from .runtime_devices import IBMQSamplerDevice
+from .vqe_runtime_runner import vqe_runner, upload_vqe_runner, delete_vqe_runner
--- a/pennylane_qiskit/ibmq.py
+++ b/pennylane_qiskit/ibmq.py
@@ -61,51 +61,13 @@ class IBMQDevice(QiskitDevice):
     short_name = "qiskit.ibmq"
 
     def __init__(self, wires, provider=None, backend="ibmq_qasm_simulator", shots=1024, **kwargs):
-        token = kwargs.get("ibmqx_token", None) or os.getenv("IBMQX_TOKEN")
-        url = kwargs.get("ibmqx_url", None) or os.getenv("IBMQX_URL")
 
-        # Specify a single hub, group and project
+        # Connection to IBMQ
+        connect(kwargs)
+
         hub = kwargs.get("hub", "ibm-q")
         group = kwargs.get("group", "open")
         project = kwargs.get("project", "main")
-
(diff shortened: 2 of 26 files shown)


pennylane_qiskit/qiskit_device.py
```diff
@@ -181,15 +180,24 @@ def process_kwargs(self, kwargs):
         self.run_args.update(kwargs)
 
     def set_transpile_args(self, **kwargs):
-        """The transpile argument setter."""
+        """The transpile argument setter.
+
+        Keyword Args:
+            kwargs (dict): keyword arguments to be set for the Qiskit transpiler. For more details,
+            see the `Qiskit documentation <https://qiskit.org/documentation/stubs/qiskit.compiler.transpile.html>`_
```
antalszava, line 187:
```suggestion
                see the `Qiskit documentation <https://qiskit.org/documentation/stubs/qiskit.compiler.transpile.html>`_
```

pennylane_qiskit/__init__.py
```diff
@@ -14,7 +14,8 @@
 """Top level PennyLane-qiskit module"""
 
 from ._version import __version__
+from .qiskit_device import QiskitDevice
```
antalszava, line 17: How come this import was added here?

pennylane_qiskit/ibmq.py
```diff
@@ -53,6 +53,9 @@ class IBMQDevice(QiskitDevice):
             variable ``IBMQX_URL`` is used, followed by the default URL.
         noise_model (NoiseModel): NoiseModel Object from ``qiskit.providers.aer.noise``.
             Only applicable for simulator backends.
+        hub (str): Name of the provider hub.
```
antalszava, line 56: Nice catch :100:

pennylane_qiskit/qiskit_device.py
```diff
@@ -181,15 +180,24 @@ def process_kwargs(self, kwargs):
         self.run_args.update(kwargs)
 
     def set_transpile_args(self, **kwargs):
-        """The transpile argument setter."""
+        """The transpile argument setter.
+
+        Keyword Args:
+            kwargs (dict): keyword arguments to be set for the Qiskit transpiler. For more details,
+                see the `Qiskit documentation <https://qiskit.org/documentation/stubs/qiskit.compiler.transpile.html>`_.
```
antalszava, line 187: As with the example in `pennylane_qiskit/ibmq.py`, instead of a general item, we'd want some specific set of keyword arguments here, right?

pennylane_qiskit/qiskit_device.py
```diff
@@ -291,7 +299,15 @@ def apply_operations(self, operations):
         return circuits
 
     def qubit_state_vector_check(self, operation, par, wires):
-        """Input check for the the QubitStateVector operation."""
+        """Input check for the the QubitStateVector operation.
+
+        Args:
+            operation (pennylane.Operation): operations to be checked
```
antalszava, line 305:
```suggestion
            operation (pennylane.Operation): operation to be checked
```

pennylane_qiskit/qiskit_device.py
```diff
@@ -313,7 +329,11 @@ def compile(self):
         return compiled_circuits
 
     def run(self, qcirc):
-        """Run the compiled circuit, and query the result."""
+        """Run the compiled circuit, and query the result.
```
antalszava, line 332:
```suggestion
        """Run the compiled circuit and query the result.
```

pennylane_qiskit/qiskit_device.py
```diff
@@ -325,7 +345,7 @@ def _get_state(self, result, experiment=None):
 
         Args:
             result (qiskit.Result): result object
-            experiment (str): the name of the experiment to get the state for
+            experiment (str): Optional, the name of the experiment to get the state for. Default=None.
```
antalszava, line 348:
```suggestion
            experiment (str or None): the name of the experiment to get the state for. Default=None.
```

pennylane_qiskit/qiskit_device.py
```diff
@@ -350,7 +370,7 @@ def generate_samples(self, circuit=None):
         :math:`q_0` is the most significant bit.
 
         Args:
-            circuit (str): the name of the circuit to get the state for
+            circuit (str): Optional, the name of the circuit to get the state for
```
antalszava, line 373:
```suggestion
            circuit (str or None): the name of the circuit to get the state for
```

antalszava (approved):
Thanks @rmoyard! :100: Once the comments are addressed, this is good to go from my side.

pennylane_qiskit/qiskit_device.py
josh146: Also there is no need to include the default value here (unless there is complex behaviour that you want to discuss), as Sphinx will automatically extract it from the signature :)

pennylane_qiskit/qiskit_device.py
rmoyard: @antalszava Yes that would be ideal but they are so many transpiler arguments that I thought it was better just to link it, what do you think?

PR #175, opened 2021-12-24
Change IBMQ login

rmoyard:
Change type check for active_account due to the recent change in IBMQ login logic. In the CI checks of the plugin matrix the token is detected contrary to the CI checks on this page (I don't know why).

### DIFF
--- a/pennylane_qiskit/ibmq.py
+++ b/pennylane_qiskit/ibmq.py
@@ -75,7 +75,7 @@ class IBMQDevice(QiskitDevice):
                 IBMQ.enable_account(token, **ibmq_kwargs)
 
             active_account = IBMQ.active_account()
-            if active_account is None:
+            if not isinstance(active_account, dict):
                 login()
             else:
                 # There is already an active account:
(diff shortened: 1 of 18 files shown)


pennylane_qiskit/ibmq.py
```diff
@@ -75,7 +75,7 @@ def login():
                 IBMQ.enable_account(token, **ibmq_kwargs)
 
             active_account = IBMQ.active_account()
-            if active_account is None:
+            if not isinstance(active_account, dict):
```
antalszava, line 78: What was the reason for the change here, what can `active_account` now be in this case? Is it boolean, just not necessarily `None`?

antalszava:
Oh got it!

>  We cannot set it to None because the test would fail.

How come? Could we then return a dummy dictionary in the tests?

I'd think changing the tests would be preferable to changing the logic, especially if we're introducing new behaviour to how Qiskit works.

rmoyard:
@antalszava I've made changes in the tests instead of the conditional statement.

antalszava (approved):
Looks good to me, thank you! :tada: :slightly_smiling_face:

PR #176, opened 2022-01-07
Fix plugin matrix tests

rmoyard:
The CI checks do not fail but the plugin matrix is. It is probably due to how the IBMQ token is set in the environment.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -53,7 +53,7 @@ Features
 Installation
 ============
 
-This plugin requires Python version 3.5 and above, as well as PennyLane and Qiskit.
+This plugin requires Python version 3.7 and above, as well as PennyLane and Qiskit.
 Installation of this plugin, as well as all dependencies, can be done using ``pip``:
 
 .. code-block:: bash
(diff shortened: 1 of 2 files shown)


antalszava:
@rmoyard the issue seems to be that we're not setting the correct environment variable with
```python
    monkeypatch.setenv("IBMQX_TOKEN", token)
```
As done in [other test cases](https://github.com/PennyLaneAI/pennylane-qiskit/blob/55d65d31fd08509a18d5d11a8df256b92a1d5ab0/tests/test_ibmq.py#L39). Here `token` comes from the `token` fixture.

Therefore `IBMQX_TOKEN_TEST` is being set, but not `IBMQX_TOKEN`. This results in the execution evaluating `False` for [the `if token is not None` condition](https://github.com/PennyLaneAI/pennylane-qiskit/blob/55d65d31fd08509a18d5d11a8df256b92a1d5ab0/pennylane_qiskit/ibmq.py#L70) with these tests (but likely evaluating `True` with the plugin test matrix suite).

Could we cover both branches? That way we're testing more cases.

rmoyard:
@antalszava I think the solution here is to change the name of env variable in the plugin matrix, at the moment is it set to `IBMQX_TOKEN` instead of `IBMQX_TOKEN_TEST`. Almost all IBMQ tests are not run because of that, I am closing this PR:

PR #177, opened 2022-01-26
Fix bugs for the trainable arguments update

rmoyard:
A recent PR changed the behavior of arguments, their trainability and the batch_params transform. We updated the arguments for one test and the @batch_params argument for another test.


puzzleshark (approved):

antalszava:
[sc-14396]

PR #178, opened 2022-01-31
Update runtime programs location

rmoyard:
Change the runtime programs location and update two tests. One test had not enough parameters. The other tests is raising a warning because the number of parameters is updated during the computation.


tests/test_runtime.py
```diff
@@ -325,7 +324,7 @@ def test_ansatz_qiskit(self, token, tol, shots):
             backend="ibmq_qasm_simulator",
             hamiltonian=hamiltonian,
             ansatz="EfficientSU2",
-            x0=[3.97507603, 3.00854038],
+            x0=[0.0, 0.0, 0.0, 0.0],
```
antalszava, line 327: How come this was changed?

tests/test_runtime.py
```diff
@@ -551,23 +550,29 @@ def test_too_many_param(self, token, tol, shots):
 
         def vqe_circuit(params):
             qml.RX(params[0], wires=0)
-            qml.RX(params[1], wires=0)
+            qml.RY(params[1], wires=0)
```
antalszava, line 553: How come this was changed?

tests/test_runtime.py
```diff
@@ -551,23 +550,29 @@ def test_too_many_param(self, token, tol, shots):
...
-        job = vqe_runner(
-            program_id=program_id,
-            backend="ibmq_qasm_simulator",
-            hamiltonian=hamiltonian,
-            ansatz=vqe_circuit,
-            x0=[3.97507603, 3.00854038, 3.55637849],
-            shots=shots,
-            optimizer_config={"maxiter": 10},
-            kwargs={"hub": "ibm-q-startup", "group": "ibm-q-startup", "project": "reservations"},
+        with pytest.warns(UserWarning) as record:
```
antalszava, line 561: Worth adding a note on this to the PR description.

doc/requirements.txt
```diff
@@ -1,4 +1,5 @@
 docutils==0.16
+mthree
```
antalszava, line 2: Minor: would this be required to build the docs?

antalszava (commented):
Thank you @rmoyard! :slightly_smiling_face:

Could you confirm that running the suite with this PR in a fashion as the plugin test matrix does (`python -m pytest plugin_repo/tests -k 'not test_ibmq.py and not test_runtime.py' --tb=short`) works well and doesn't produce [an import error](https://github.com/PennyLaneAI/plugin-test-matrix/runs/5007267051?check_suite_focus=true#step:8:28)?

tests/test_runtime.py
rmoyard: We are using EfficientSu2 from the Qiskit library and it has 4 parameters when there are 2 qubits.

tests/test_runtime.py
rmoyard: It makes more sense to have two different operations here, anyway we are not interested in the actual result.

tests/test_runtime.py
rmoyard: We are using EfficientSu2 from the Qiskit library and it has 4 parameters when there are 2 qubits.

tests/test_runtime.py
rmoyard: Yes I am adding it!

doc/requirements.txt
rmoyard: Yes it is necessary exactly.

antalszava (approved):
Thank you! :slightly_smiling_face:

PR #182, opened 2022-02-04
Fix VQE runtime ValueError

rmoyard:
Fix a small bug displaying a ValueError on Qiskit IBMQ for VQE runtime.


antalszava (approved):
Works well, thanks for the super-fast fix! :rocket: The only suggestion is to leave a comment about when we have dictionaries and when we don't in the affected part of code.

pennylane_qiskit/vqe_runtime_runner.py
```diff
@@ -73,12 +73,13 @@ def _callback(self, *args):
             step (float): Value of the step.
             accepted (bool): True if the loss function value has improved, False otherwise.
         """
-        _, (nfev, xk, fk, step, accepted) = args
-        self.intermediate_results["nfev"].append(nfev)
-        self.intermediate_results["parameters"].append(xk)
-        self.intermediate_results["function"].append(fk)
-        self.intermediate_results["step"].append(step)
-        self.intermediate_results["accepted"].append(accepted)
+        if not isinstance(args[1], dict):
```
rmoyard, line 76:
```suggestion
        # If it is a dictionary it is the final result and does not belong to intermediate results
        if not isinstance(args[1], dict):
```

antalszava: Looks great :tada:

PR #184, opened 2022-02-10
Add workflow for running IBMQ tests on a schedule

antalszava:
Run IBMQ integration tests twice a week (Sundays and Thursdays).


rmoyard (approved):
It looks good to me, thank you for adding those tests back! We will have to wait for IBMQ to work again though 💯

PR #187, opened 2022-02-14
Change callback function for Sicpy intermediate results

rmoyard:


pennylane_qiskit/vqe_runtime_runner.py
```diff
@@ -89,7 +89,9 @@ def _scipy_callback(self, *args):
             xk (array_like): A list or NumPy array to attach.
         """
         _, xk = args
-        self.intermediate_results["parameters"].append(xk)
+        # If it is a dictionary it is the final result and does not belong to intermediate results
+        if not isinstance(args[1], dict):
```
antalszava, line 93: @rmoyard can't suggest it, but how about:

```python
        # If it is a dictionary it is the final result and does not belong to intermediate results
        if not isinstance(args[1], dict):
                _, xk = args
                self.intermediate_results["parameters"].append(xk)
```
I.e., we move the unpacking too into the conditioned block.

PR #190, opened 2022-02-22
Update `hamiltonian.terms` to `hamiltonian.terms()` due to recent change

rmoyard:
Replace `hamiltonian.terms` with `hamiltonian.terms()`


antalszava (approved):
Could you add a small changelog item?

Looks good :100:

PR #201, opened 2022-03-18
Update runtime sampler

rmoyard:
The runtime `sampler` was updated on March 17 by IBM, some changes in the arguments of the runtime program are needed. The results are returned as probabilities instead of counts, therefore the `generate_samples` function is adapted.


rmoyard:
https://github.com/PennyLaneAI/pennylane-qiskit/actions/runs/2017778007

puzzleshark (approved):

PR #204, opened 2022-03-23
Updated version bump workflows to use composite GH actions

Jaybsoni:
Using the `PennyLaneAI/automation` repo to store version bump logic. Updated the workflow files here to call the composite action.


.github/workflows/post_release_version_bump.yml
```diff
@@ -2,40 +2,29 @@ name: Automated Release Version Bumps
...
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
         with:
           commit-message: post release version bump
           title: Version Bump
           body: updated changelog and _version.py
           branch: post-release-version-bump
-          reviewers: Jaybsoni
-          base: main
+          reviewers: rmoyard
```
Jaybsoni, line 29: This will create a PR for both the pre-release and post-release version bumps with @rmoyard as the reviewer. If next release goes well, we can potentially just allow it merge directly.

rmoyard (approved):
It looks good to me :100 Thank you

PR #220, opened 2022-06-08
Define `returns_state` in `capabilities`

antalszava:
The state backend devices are capable of returning the state, yet they don't have the `returns_state=True` entry in the `capabilities` dictionary of the device (see [the explanation of `capabilities` here](https://pennylane.readthedocs.io/en/stable/development/plugins.html#device-capabilities)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,8 @@
 
 This release contains contributions from (in alphabetical order):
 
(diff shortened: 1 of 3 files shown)


rmoyard (approved):
Thank you @antalszava 💯 Don't forget to add your name in the changelog

PR #221, opened 2022-06-13
Update `requirements-ci.txt`

rmoyard:
~Unecessary `pennylane` import in the `requirements-ci.txt` file is removed.~

`requirements-ci.txt` uses master of pennylane `git+https://github.com/PennyLaneAI/pennylane.git`

### DIFF
--- a/requirements-ci.txt
+++ b/requirements-ci.txt
@@ -1,5 +1,4 @@
 qiskit
 mthree
-pennylane
 numpy
 sympy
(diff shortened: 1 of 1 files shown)


antalszava:
Is the import not required because we import PennyLane directly in the workflow file? Maybe worth leaving a note here, as PennyLane would be one import that comes to mind here.

rmoyard:
@antalszava It not important because it is installed directly when you install the plugin in a first place. If the versions are not the same then it will create some conflicts.

antalszava (approved):
:+1:

PR #229, opened 2022-09-19
add support for iswap gate

timmysilv:
the ISWAP gate exists in both Qiskit and PennyLane, so this PR adds support for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### New features since last release
 
+* Add support for the `ISWAP` operation.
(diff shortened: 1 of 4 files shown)


albi3ro (commented):
Thanks for catching this 👍 

Just needs a changelog entry, located in `./CHANGELOG.md`.  Seems like the latest version of the changelog needs to be updated too.

antalszava (approved):
Good catch! :tada: As per Christina's suggestion, needs a changelog entry. Approving on condition of that being added here.

PR #230, opened 2022-09-22
Add support for controlled ops.

AlbertMitjans:
Add mapping to several controlled operations in PennyLane that have a specific `Qiskit` class.


pennylane_qiskit/qiskit_device.py
```diff
@@ -283,8 +300,16 @@ def apply_operations(self, operations):
...
-            device_wires = self.map_wires(operation.wires)
-            par = operation.parameters
+            if isinstance(operation, Controlled):
+                device_wires = self.map_wires(operation.base.wires + operation.control_wires)
+            else:
+                device_wires = self.map_wires(operation.wires)
+
+            if isinstance(operation, Barrier):
+                op_num_wires = operation.num_wires
+                par = [self.num_wires] if op_num_wires == -1 else [op_num_wires]
```
timmysilv, line 310: does anything else in PennyLane ever have `num_wires = -1`? If not (or if it should be treated this way as well), I'd replace all this with:
```python
par = [operation.num_wires]
if par == [-1]:
    par = [self.num_wires]
```
to avoid the `isinstance` call. Then again, maybe comparing arrays is slower... something like that :p

AlbertMitjans: The problem here is that the parameters of Qiskit's barrier operation corresponds to the `num_wires` of PL's Barrier operation

pennylane_qiskit/qiskit_device.py
```diff
@@ -283,8 +300,16 @@ def apply_operations(self, operations):
 
         for operation in operations:
             # Apply the circuit operations
-            device_wires = self.map_wires(operation.wires)
-            par = operation.parameters
+            if isinstance(operation, Controlled):
+                device_wires = self.map_wires(operation.base.wires + operation.control_wires)
```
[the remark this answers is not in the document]
AlbertMitjans: The problem is the order of the wires! 

image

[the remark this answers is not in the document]
timmysilv: aha! missed this, ty for clarifying

pennylane_qiskit/qiskit_device.py
timmysilv: I suppose that would have been too much generalization without a reason. roger that

PR #234, opened 2022-11-11
don't try to use `token` if it is falsy

timmysilv:
Today it checks specifically if `token` is `None`. It might as well check that it's not `""` either.

TODO: add test


albi3ro (approved):
Changelog? 

Nice to see that the problem has such a simple fix 👍

PR #252, opened 2022-12-20
Add quotes to python version

Jaybsoni:


.github/workflows/upload.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.10
+          python-version: "3.10"
```
rashidnhm, line 15:
```suggestion
          python-version: '3.10'
```
Since we want it as a literal string and not escaping any sequences, better to use single quotes

rashidnhm (approved):

PR #260, opened 2023-01-24
Replace inv by adjoint

rmoyard:
This PR replaces `.inv` with `qml.adjoint` and remove the recorder around QNodes in two tests.


lillian542 (approved):
Looks good to me 😃

PR #267, opened 2023-02-08
Fix #266: check if i is in `states`; avoid indexing

### LINKED ISSUE #266: IndexError: list index out of range when checking states[i] != i
This line: https://github.com/EelVex/pennylane-qiskit/blob/e84c57c96ff993894fbb0007af26dd9f4c1d0658/pennylane_qiskit/runtime_devices.py#L219
will trigger an `IndexError` if the `number_of_states-1` is not in `states`.

EelVex:


AlbertMitjans:
Hi @EelVex! I lack a bit of context here. Why do you request this change? 

Using `if i not in states` has more time complexity than `if states[i] != i` (the first option needs to go through all the list in the worst case, while the second option has to check only one value). Why do we want to avoid indexing?

EelVex:
Hi @AlbertMitjans. This fixes the bug #266 where an IndexError is triggered when the max `i` included in `states` is less than  `number_of_states-1`. For example if `states = [1, 2, 5]` and `number_of_states = 8` the check `states[6] != 6` can not be performed.

This PR tries to fix the bug using the fewest modifications possible from the original. If time complexity is of an issue then another fix could be:
 
```python
if len(states) < number_of_states:
    missing = set(range(number_of_states)) - set(states)
    for i in missing:
        probs.insert(i, 0.0)
``` 

or, perhaps even better, the loop could be replaced with a `np.insert(np.array(probs), missing, 0.0)`

AlbertMitjans:
> the loop could be replaced with something like np.insert(np.array(probs), missing, 0.0)

@EelVex this would be much better! Can you implement it?

AlbertMitjans:
Also please add a new test testing that the bug is fixed

EelVex:
> > the loop could be replaced with something like np.insert(np.array(probs), missing, 0.0)
> 
> @EelVex this would be much better! Can you implement it?

@AlbertMitjans sure. New PR or in this one?

AlbertMitjans:
> New PR or in this one?

@EelVex in this one!

EelVex:
@AlbertMitjans the last commit makes the function more readable and much more efficient. I'm not sure where (and if) testing is needed.

AlbertMitjans:
Hi @EelVex ! This looks much better, thanks!!

Please add a test in `tests/test_runtime.py` in the class `TestSampler`, testing that the bug that you reported is fixed.

Also add an entry in `CHANGELOG.md` under `Bug fixes` explaining in one line the bug that you just fixed.

fran-scala:
This is still giving problems (with version `0.32.0`) when not all the states are measured. Is this PR going to be merged in the future?

lillian542:
Fixed in #337

PR #277, opened 2023-02-24
bump pl requires version

albi3ro:
While the test suite passes with PL v0.28 and the v0.29 release candidate, the test suite fails with PL v0.27.

For example, I am getting failures may look like:
```
FAILED pennylane/devices/tests/test_gates.py::TestSupportedGates::test_supported_gates_can_be_implemented[device_kwargs0-CNOT] - AttributeError: 'CNOT' object has no attribute 'return_type'
```

Therefore I'm bumping the PL requirement to be at least v0.28.


timmysilv (approved):
nice find!

mudit2812 (approved):
Thanks for the change!

PR #281, opened 2023-04-10
Compatibility with the new return system

rmoyard:


rmoyard:
[sc-35005]

tests/test_qiskit_device.py
```diff
@@ -140,8 +141,10 @@ def test_calls_to_reset(self, n_tapes, mocker, device):
 
         assert spy.call_count == n_tapes
 
-    def test_result(self, device, tol):
+    def test_result_legacy(self, device, tol):
         """Tests that the result has the correct shape and entry types."""
+        # TODO: remove once the legacy return system is activated.
```
albi3ro, line 146:
```suggestion
        # TODO: remove once the legacy return system is removed.
```

albi3ro: Also, the failure of this test once the legacy system is removed will be a pretty clear indication this needs to be removed.

albi3ro (approved):
Left a suggestion about the TODO comment, but its not particularly important.

Looks good to me 👍

mudit2812 (commented):
Looks good Romain! Please add a changelog entry. Do we need to pin the required version of pennylane in `requirements.txt` since the private functions for the new return types were renamed?

PR #296, opened 2023-04-27
require 0.30 rc branch

timmysilv:
as title says. this is to make plugin-test-matrix pass, and will need to be updated before releasing this plugin


albi3ro (approved):

mudit2812 (approved):
Thanks! Just a reminder that we need to update this before uploading for release

PR #298, opened 2023-04-28
Update IBMQ; remove VQE program/uploader, use qiskit VQE

timmysilv:
The `vqe_runner` broke, because the IBMQ uploader system just seems to have... disappeared. Fortunately, IBM also created a VQE program so we don't need to upload one! We just need to use the new `qiskit-ibm-runtime` package to invoke it, and re-work some stuff on our end a bit.

Some notable changes:
- 🚨 breaking change 🚨 : The first argument to `vqe_runner`, `program_id`, has been removed. It used to be the ID of the program you'd upload, but now it's always the program provided by IBM (`vqe`)
- We must now give initialized optimizers to Qiskit rather than just their names (pl_qiskit users still just provide "SPSA" or some SciPy optimizer name as a string). However, if you initialize a Qiskit QNSPSA optimizer with _their_ fidelity function, it fails to serialize, so I've disabled it on our end (this needs a story to follow up on)
- Their new-style Hamiltonian observables must be in the Pauli basis, so I removed support for Hadamard being part of terms. Sounds familiar 😄 
- I didn't change any of the code below `vqe_runner`, just indents and stuff. The only exception is that I swapped the order of the coeff-term tuple-pair returned by `hamiltonian_to_list_string` because that's the order expected by the `PauliSumOp.from_list` [constructor](https://qiskit.org/documentation/stubs/qiskit.opflow.primitive_ops.PauliSumOp.from_list.html#qiskit.opflow.primitive_ops.PauliSumOp.from_list)

Stating this because it seems we don't run it on CI: I ran `test_runtime.py` locally. It all passes!

Closes #287
Closes #289 

EDIT: this PR now includes the contents from #301, related to upgrading the IBMQ runtime/provider code.

PR #304, opened 2023-05-23
Add qiskit.generic device that accepts instantiated qiskit backends.

airwoodix:
This patch introduces a new `qiskit.generic` device that accepts instantiated Qiskit backends and wraps them into a Pennylane device.

This offers a more composable mechanism for specifying the Qiskit backend, especially for providers that require more filter criteria than only a name to select a unique backend with [`get_backend`](https://qiskit.org/documentation/stubs/qiskit.providers.ProviderV1.get_backend.html#qiskit.providers.ProviderV1.get_backend).

Basic integration tests were added.

A small patch on the call to `JobV1.result()` was done: changes in #298 pass a `timeout` argument to [`JobV1.result()`](https://qiskit.org/documentation/stubs/qiskit.providers.JobV1.result.html#qiskit.providers.JobV1.result). While supported by many job handles, including that from IBMQ, it is *not* part of the ABC definition and is therefore not required to be valid.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### New features since last release
 
+* Added a `GenericDevice` (PennyLane device name: `qiskit.generic`) that accepts a backend
+  instance directly. [(#304)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/304)
+
 ### Breaking changes
 
 ### Improvements
@@ -13,6 +16,8 @@
 ### Contributors
 
 This release contains contributions from (in alphabetical order):
+
+Etienne Wodey (Alpine Quantum Technologies GmbH)
 
 ---
 # Release 0.30.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### New features since last release
 
-* Added a `GenericDevice` (PennyLane device name: `qiskit.generic`) that accepts a backend
+* Added a `RemoteDevice` (PennyLane device name: `qiskit.remote`) that accepts a backend
   instance directly. [(#304)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/304)
 
 ### Breaking changes
--- a/pennylane_qiskit/generic.py
+++ b/pennylane_qiskit/generic.py
@@ -1,0 +1,43 @@
+# Copyright 2019 Xanadu Quantum Technologies Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+This module contains the :class:`~.GenericDevice` class, a PennyLane device that allows
+evaluation and differentiation on any Qiskit backend using Pennylane.
+"""
+
+from .qiskit_device import QiskitDevice
+
+
+class GenericDevice(QiskitDevice):
+    """A PennyLane device for any Qiskit backend.
+
+    Args:
+        wires (int or Iterable[Number, str]]): Number of subsystems represented by the device,
+            or iterable that contains unique labels for the subsystems as numbers (i.e., ``[-1, 0, 2]``)
+            or strings (``['ancilla', 'q1', 'q2']``).
+        provider (Provider | None): provider to lookup the backend on (ignored if a backend instance is passed).
+        backend (str | Backend): the desired backend. Either a name to look up on a provider, or a
+            BackendV1 or BackendV2 instance.
+        shots (int or None): number of circuit evaluations/random samples used
+            to estimate expectation values and variances of observables. For statevector backends,
+            setting to ``None`` results in computing statistics like expectation values and variances analytically.
+
+    Keyword Args:
+        name (str): The name of the circuit. Default ``'circuit'``.
+    """
+
+    short_name = "qiskit.generic"
+
+    def __init__(self, wires, backend, provider=None, shots=1024, **kwargs):
+        super().__init__(wires, provider=provider, backend=backend, shots=shots, **kwargs)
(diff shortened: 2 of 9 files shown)


pennylane_qiskit/generic.py
```diff
@@ -0,0 +1,43 @@
...
+            setting to ``None`` results in computing statistics like expectation values and variances analytically.
+
+    Keyword Args:
+        name (str): The name of the circuit. Default ``'circuit'``.
+    """
+
+    short_name = "qiskit.generic"
+
+    def __init__(self, wires, backend, provider=None, shots=1024, **kwargs):
+        super().__init__(wires, provider=provider, backend=backend, shots=shots, **kwargs)
```
timmysilv, line 43: I see that you're trying to make provider optional, but is it possible to somehow match the parent class constructor? wouldn't need to copy-paste the docstring in that case

pennylane_qiskit/qiskit_device.py
```diff
@@ -154,25 +162,26 @@ def __init__(self, wires, provider, backend, shots=1024, **kwargs):
 
             self.shots = 1024
 
-        self._backend = None
-
         self.provider = provider
```
timmysilv, line 165: I don't want to remove `provider` from `self` because users might depend on it, but it's not used outside of the `__init__` method anymore, so I kinda do...

pennylane_qiskit/qiskit_device.py
```diff
@@ -154,25 +162,26 @@ def __init__(self, wires, provider, backend, shots=1024, **kwargs):
...
-        self.backend_name = backend
-
-        def _get_backend_name(name):
-            return name if isinstance(name, str) else name()
 
-        self._capabilities["backend"] = [
```
timmysilv, line 165: tbh I'm not entirely sure why this was added to the device capabilities, but again, users might depend on it. This PR could be considered "breaking" in that sense, but really this should have been private even in the old implementation

timmysilv (commented):
thanks for putting this together, it looks great! Device construction feels much more intuitive. just a few things to bring up:
- If you can test the V2 for `backend.name` and calling `job.result()` without timeout, it would be sweet. If not, we'll have to skip it somehow. Dealing with the changing IBMQ API can be tough sometimes so it's all good
- Can you give some details on the specific scenario in which you want to pass an already-existing device? It feels a bit funny making a new class just to implement `short_name` (not too bad, but I guess there are other ways around that if that's the only problem 😋 )

airwoodix:
Thanks for the positive review!

> If you can test the V2 for `backend.name` and calling `job.result()` without timeout, it would be sweet. If not, we'll have to skip it somehow. Dealing with the changing IBMQ API can be tough sometimes so it's all good

I'll have a look. It turns out to be difficult to find a concrete implementation of a `BackendV2` that doesn't also implement part of the `BackendV1` interface. Maybe a dummy would do as well?

> Can you give some details on the specific scenario in which you want to pass an already-existing device?

- if a backend needs to be configured (e.g. setting options), it's more convenient to do it before wrapping it as a PennyLane device that retrieving the backend through the private and untyped `self._capabilities["backend"]` value;
- some providers might require more than just a `name` string to uniquely identify a backend. This is allowed by [`ProviderV1.get_backend`](https://qiskit.org/documentation/stubs/qiskit.providers.ProviderV1.get_backend.html) using the extra keyword arguments. Rather than expanding the `qml.device()` signature to support this, I think it's easier and more elegant to use dependency injection.

> It feels a bit funny making a new class just to implement `short_name` (not too bad, but I guess there are other ways around that if that's the only problem yum )

I initially intended to use `QiskitDevice` directly. However, making the `provider` argument optional (null by default) would require modifying the argument order in `QiskitDevice.__init__`, which would be a major breaking change.

> 
> Don't forget to add a changelog entry!

Done. For the list of contributors, I'd be happy about adding an affiliation. Would that be fine for you?

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ### New features since last release
 
+* Added a `GenericDevice` (PennyLane device name: `qiskit.generic`) that accepts a backend
```
timmysilv, line 5: slight name preference: could we call it `RemoteDevice` instead? Same with the rest of the changes.
```suggestion
* Added a `RemoteDevice` (PennyLane device name: `qiskit.remote`) that accepts a backend
```

timmysilv (commented):
this all makes sense, and looks good! Just one concern with naming. Also, could you update [index.rst](https://github.com/PennyLaneAI/pennylane-qiskit/blob/master/doc/index.rst) with a title card for this new device, and have it point to a new file in the `devices/` subfolder? It could be small like BasicAer's file, just something to show how/why to use this new device. Thanks!

timmysilv:
regarding your questions:
- don't worry much about the testing of V2 stuff, it's nearly identical
- those scenarios make sense. In the past, we'd have preferred you make a new device class for the specific device you're talking about, and having it accept the various custom args needed. This is an overall goal of the PennyLane plugins, but I can appreciate that this case goes beyond it (given the provider requirement). Perhaps we'll update the interaction between provider and backend in a later release.
- Changing the arg/kwarg order isn't the _best_ news, but it is essentially half of the point of this PR so it'll do. Plus, it's an isolated change
- the affiliation in the changelog is fine by us!

airwoodix:
Thanks for the review!

As requested, I renamed `qiskit.generic` to `qiskit.remote` and added a (very) short documentation in `doc/devices/remote.rst`.

I also added explicit `#pragma: no cover` directives for the backend v1/v2 and timeout/no timeout switches.

@timmysilv from your last comment, I understand that the remaining open threads do not need immediate action. Please let me know if that's not the case.

pennylane_qiskit/remote.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2019 Xanadu Quantum Technologies Inc.
```
timmysilv, line 1:
```suggestion
# Copyright 2023 Xanadu Quantum Technologies Inc.
```

timmysilv (approved):
great stuff, thanks for addressing all our concerns! glad to have this in 🎉  and yes, you can disregard the previous comments 😄

airwoodix:
Sounds great, thanks a lot for considering this patch and reviewing it.

> re: docs failure

Thanks for the hint! Should be fixed now.

timmysilv:
looks great! I don't have permission to update the branch on your fork, but I fixed the codecov issue on master. If you pull in the latest master, I'll re-run CI and merge 😎

airwoodix:
Thanks! I rebased on latest `master`.

PR #305, opened 2023-05-24
Include qubit gates from `pennylane.ops.op_math` in the converter.

frederikwilde:
When converting a Qiskit circuit to a PennyLane circuit a list of supported gates in PennyLane is required. This was previously given by the set `pennylane.ops.qubit.ops`. `CZ` was recently moved (https://github.com/PennyLaneAI/pennylane/pull/4117) to `pennylane.ops.op_math`, which lead to `CZ` not being one of the supported gates.

By replacing the current list of supported gates with `pennylane.ops._qubit__ops__` this issue is fixed.

### DIFF
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -25,7 +25,7 @@ from qiskit.exceptions import QiskitError
 from sympy import lambdify
 
 import pennylane as qml
-import pennylane.ops.qubit as pennylane_ops
+import pennylane.ops as pennylane_ops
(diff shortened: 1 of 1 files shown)


timmysilv (approved):
assuming my suggestion pleases codecov, lgtm! I'd usually push back more against accessing private attributes, but it's [already a rough situation](https://github.com/PennyLaneAI/pennylane/blob/8af85140f9778e9a171e0e1bf8ebf6cbab42222d/pennylane/ops/__init__.py#L42) so it's ok

PR #310, opened 2023-06-20
fix the ensemble classification demo thumbnail

timmysilv:
the current one doesn't render because the extra `/qml`. Will also make this change in other plugins.

Also added the codecov ignore line, because that line of code is no longer needed in Qiskit >= 0.43.1 (explanation [here](https://qiskit.org/documentation/release_notes.html#upgrade-notes), tl;dr bound ParameterExpressions that represent floats now just get converted to floats, dropping the unnecessary wrapper)

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -80,7 +80,7 @@ Check out these demos to see the PennyLane-Qiskit plugin in action:
 
 .. demogalleryitem::
     :name: Ensemble classification with Forest and Qiskit devices
-    :figure: _static/thumbs/ensemble_diagram.png
+    :figure: https://pennylane.ai/_images/ensemble_diagram.png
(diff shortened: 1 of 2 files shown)


doc/index.rst
```diff
@@ -80,9 +80,9 @@ Check out these demos to see the PennyLane-Qiskit plugin in action:
 
 .. demogalleryitem::
     :name: Ensemble classification with Forest and Qiskit devices
-    :figure: https://pennylane.ai/qml/_images/ensemble_diagram.png
+    :figure: _static/thumbs/ensemble_diagram.png
```
timmysilv, line 83:
```suggestion
    :figure: https://pennylane.ai/_images/ensemble_diagram.png
```

PR #313, opened 2023-06-20
Remove retired VQE runner

timmysilv:
[sadge](https://github.com/PennyLaneAI/pennylane-qiskit/actions/runs/5300854181/jobs/9594666139#step:7:48)

the CI error: `qiskit_ibm_runtime.exceptions.IBMInputValueError: 'The vqe and qaoa programs have been retired in the Qiskit Runtime service. Please visit https://qiskit.org/ecosystem/ibm-runtime for an introduction on Sessions and Primitives, and to access tutorials on how to execute VQE and QAOA using Qiskit Runtime Primitives.'`

too much effort to do any more than this.


setup.py
```diff
@@ -22,7 +22,7 @@
     long_description = fh.read()
 
 requirements = [
-    "qiskit>=0.32",
+    "qiskit>=0.32,<=0.42.1",
```
rmoyard, line 25:
```suggestion
    "qiskit>=0.42.1",
```

Should it not be that?

trbromley: It's to prevent this issue in a test:
> qiskit.exceptions.QiskitError: "Error decomposing node of instruction 'rx': 'NoneType' object has no attribute 'global_phase'. Unable to define instruction 'u' in the basis."

Though @timmysilv, have you tried fixing the issue?

trbromley (commented):
Thanks @timmysilv! Overall looks good to me. We can come back to deciding what we want to do here during the 0.33 cycle.

setup.py
timmysilv: i have not, but I figured this week isn't the best time to do it. i'll give it a shot today

timmysilv: turns out Qiskit/qiskit-terra#9343 changed the behaviour of the "unroller" transpilation method option to account for target ops, and this broke RX badly. the intention of the test is only to show that we can pass custom transpilation options, so I've changed it to just use the "translator" option! those should pass with the latest qiskit (confirmed locally)

rmoyard (commented):
Can you run the IBMQ integration test from this PR?

timmysilv:
passed: https://github.com/PennyLaneAI/pennylane-qiskit/actions/runs/5338887808

rmoyard (approved):
Thanks :+1:

PR #317, opened 2023-07-11
remove test now that disable_return is deprecated

timmysilv:
just doing what the TODO says 😄


rmoyard (approved):
:+1:

timmysilv:
never mind.. there are code changes required for this too, so we should wait until the next PL release when it's actually removed

PR #320, opened 2023-07-21
Update rtd config to v2

rashidnhm:
This PR updates the RTD configuration to use the new v2 syntax.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -8,9 +8,8 @@ python:
     - requirements: doc/requirements.txt
     - method: pip
       path: .
-  system_packages: true
 
 build:
   os: ubuntu-22.04
   tools:
-    python: "3.8"
+    python: "3.9"
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -13,3 +13,5 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.9"
+  apt_packages:
+    - graphviz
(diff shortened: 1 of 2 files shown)


.readthedocs.yml
```diff
@@ -1,11 +1,16 @@
...
+      path: .
+  system_packages: true
 
-# Don't build any extra formats
-formats:
-    - none
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.8"
```
albi3ro, line 16: Should we also bump the python version? I think we'll be dropping python 3.8 support in the next release.

.readthedocs.yml
```diff
@@ -1,11 +1,16 @@
...
+sphinx:
+  configuration: doc/conf.py
 
 python:
-    version: 3.8
+  install:
+    - requirements: doc/requirements.txt
+    - method: pip
+      path: .
+  system_packages: true
```
timmysilv, line 11: is this necessary?

timmysilv (approved):
+1 to Christina's comment re: 3.9, otherwise lgtm!

.readthedocs.yml
rashidnhm: Originally saw it being used in the lightning repo. But removed as it's being deprecated!

.readthedocs.yml
rashidnhm: Bumped!

albi3ro (approved):
👍 thanks :)

PR #331, opened 2023-08-30
Removed old return type

mudit2812:
As title says


albi3ro (commented):
looks like a couple more tests need to be removed.

albi3ro (approved):

PR #334, opened 2023-09-21
remove mthree since it is not needed

timmysilv:
mthree was originally added for the VQE runner, which isn't around anymore. I want to remove it because it's messing with my local env. It requires `qiskit-ibmq-provider` which requires `numpy<1.24`, and those days are behind us.

note: numpy is still pinned to <1.24 on PL 0.32 (which this requires, but after the next release this might all change) so CI ran with that.


rmoyard (approved):
:+1:

Issue #332, opened 2023-09-05
Support `b.num_qubits` for `BackendV2` Qiskit objects


JMuff22:
My feature request is related to a problem of using Qiskit backends which only support `BackendV2` objects and don't support backwards compatibility for `BackendV1`. [This is described in the Qiskit documentation. ](https://qiskit.org/documentation/stubs/qiskit.providers.BackendV2.html)

The current issue is that when running with such a backend fails at this [line](https://github.com/PennyLaneAI/pennylane-qiskit/blob/f77addaad382c687cc82d199b0ee0a72db485999/pennylane_qiskit/qiskit_device.py#L188) due to calling `b.configuration().n_qubits` , for which the Backend object has no `configuration()` method. The solution would be to use `backend.num_qubits`. 

The solution is not simple however, because many of the tests and backends support by pennylane-qiskit have the opposite functionality in that they don't support `BackendV2` and `backend.num_qubits` cannot be called.  The solution should support both `BackendV1` and `BackendV2` via `backend.configuration().n_qubits` and `backend.num_qubits.`

I tried this solution but I considered it hacky.

```python

    def validate_wires_against_backend(self):
        b = self.backend
        
        try:
            num_qubits = int(b.configuration().n_qubits)
        except AttributeError:
            try:
                num_qubits = int(b.num_qubits)
            except AttributeError:
                raise AttributeError("Neither 'b.configuration().n_qubits' nor 'b.num_qubits' exist.")
                
        if len(self.wires) > num_qubits:
            raise ValueError(f"Backend '{b}' supports a maximum of {num_qubits} wires.")
```

Thanks!

The particular backend I am trying to use is [`IQMBackend`](https://github.com/iqm-finland/qiskit-on-iqm/blob/eca5bd032ff97d861943c0c58a473b4952a23eeb/src/qiskit_iqm/iqm_provider.py#L41).

CatalinaAlbornoz:
Hi @JMuff22, thank you very much for opening this feature request! It's very helpful for us to improve the plugin. 

By the way, we have a new [PennyLane survey](http://bit.ly/PennyLane-Survey). We would love to know your thoughts about PennyLane in order to keep bringing you amazing features :sparkles:.

JMuff22:
Whilst some checks were added in https://github.com/PennyLaneAI/pennylane-qiskit/pull/493 . To support all backends we have been using 

```
	def validate_wires_against_backend(self):
		b = self.backend

		try:
[...]

PR #335, opened 2023-09-29
Add validate_wires_against_backend function

JMuff22:
Closes: https://github.com/PennyLaneAI/pennylane-qiskit/issues/332 

Add a function that allows for both `backend.num_qubits` to be used in addition to `backend.configuration().n_qubits` which allows usage on both BackendV1 objects and BackendV2 objects which don't support backwards compatibility.


trbromley:
Hi @JMuff22! Apologies we didn't provide feedback sooner. Do you no longer thing these proposed changes are needed?

JMuff22:
Hi @trbromley  thanks for the response. The changes are still needed however I was willing to try to figure out why the CI tests were not working. I am happy and will continue with the current solution however it would be great if pennylane-qiskit could be compatible with BackendV1 objects and BackendV2 objects.

PR #343, opened 2023-10-27
fix qiskit_device to support Aer v2 conventions

timmysilv:
Tests suddenly [began to fail](https://github.com/PennyLaneAI/plugin-test-matrix/actions/runs/6661145908/job/18103535073) with the release of Aer 0.13.0 yesterday. The main change is that it got upgraded to a V2 backend. The consequence is that we depended on the device name to check its behaviours.

Some code to highlight the change:
**Qiskit-Aer v0.12.2**
```pycon
>>> dev = qml.device("qiskit.aer", wires=2, backend="aer_simulator_unitary")
>>> dev.backend, dev.backend.options.method
(AerSimulator('aer_simulator_unitary'), 'unitary')
```

**Qiskit-Aer v0.13.0**
```pycon
>>> dev = qml.device("qiskit.aer", wires=2, backend="aer_simulator_unitary")
>>> dev.backend, dev.backend.options.method
(AerSimulator('aer_simulator'), 'unitary')
```

We often check things like `if some_method in dev.backend_name`, so I had to update those checks to be more general and support the new (and undocumented!) V2 backend naming convention used with Aer.

Test change: for some reason, transpilation of the circuit in that one test turned a PauliZ into an RZ(-pi), and we were only applying the noise to Z gates. I just added RZ to the list of gates we apply noise on to fix it.

### DIFF
--- a/pennylane_qiskit/qiskit_device.py
+++ b/pennylane_qiskit/qiskit_device.py
@@ -155,13 +155,6 @@ class QiskitDevice(QubitDevice, abc.ABC):
     def __init__(self, wires, provider, backend, shots=1024, **kwargs):
         super().__init__(wires=wires, shots=shots)
 
-        # Keep track if the user specified analytic to be True
-        if shots is None and backend not in self._state_backends:
-            # Raise a warning if no shots were specified for a hardware device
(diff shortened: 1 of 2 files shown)


mudit2812 (approved):
Thanks for the quick fix!

PR #347, opened 2023-11-07
Small fixes for compatibility with most recent Qiskit version

lillian542:


pennylane_qiskit/converter.py
```diff
@@ -179,7 +179,8 @@ def _function(params: dict = None, wires: list = None):
 
         # Processing the dictionary of parameters passed
         for op, qargs, cargs in qc.data:
-            instruction_name = op.__class__.__name__
+            # the new Singlton classes have different names than the objects they represnet, but base_class.__name__ still matches
```
lillian542, line 182:
```suggestion
            # the new Singlton classes have different names than the objects they represent, but base_class.__name__ still matches
```

pennylane_qiskit/converter.py
```diff
@@ -179,7 +179,8 @@ def _function(params: dict = None, wires: list = None):
 
         # Processing the dictionary of parameters passed
         for op, qargs, cargs in qc.data:
-            instruction_name = op.__class__.__name__
+            # the new Singlton classes have different names than the objects they represent, but base_class.__name__ still matches
+            instruction_name = op.base_class.__name__
```
timmysilv, line 183: was `op.base_class` introduced in a recent version? in other words, do we need to increase the qiskit lower-bound with this change?

tests/test_converter.py
```diff
@@ -267,21 +267,6 @@ def test_extra_parameters_were_passed(self, recorder):
             with recorder:
                 quantum_circuit(params={theta: x, phi: y})
 
-    def test_quantum_circuit_error_by_passing_wrong_parameters(self, recorder):
```
timmysilv, line 270: what happens with this test now?

albi3ro: Different package now raises an error.

Given it's not our responsibility to make sure that the correct external package is handling the errors at the right place, I'm okay with removing this test.

This wasn't as much a "PennyLane" test as a qiskit test, and qiskit decided to do something different.

lillian542: This test checked if an error is raised _in Qiskit_ if the wrong parameters are passed - I guess we weren't raising an error but wanted to make sure one was raised? But its not raised in the `load` method.

Qiskit previously raised a Qiskit `CircuitError`. Now Qiskit also doesn't raise the error themselves, they let `sympy` raise a `SympfiyError`.

You use to get this:

```
CircuitError: 'Expression cannot bind non-numeric values ({Parameter(θ): tensor("string", requires_grad=False)})'
```

Now you get this:

```
SympifyError: sympy2symengine: Cannot convert 'tensor("string", requires_grad=False)' (of type <class 'pennylane.numpy.tensor.tensor'>) to a symengine type.
```

It happens the same place, when trying to bind parameters. Its just a different error.

I also don't think this test was testing what it was supposed to - you get the same result whether you use a string or a float, i.e. it will also tell you:
```
CircuitError: 'Expression cannot bind non-numeric values ({Parameter(θ): tensor(0.5, requires_grad=False)})'
```
on the previous version of Qiskit. If `requires_grad=True`, both the string and the float are accepted (also in the previous version). So it wasn't really testing if `"some_string_instead_of_an_angle"` was a bad input to begin with. (Maybe it was many Qiskit versions ago. I compared to 0.43.0.). It was testing what happens when `requires_grad` is `True` vs `False`. 

I'm not sure why we are testing whether _Qiskit_ (not the plugin) raises errors correctly if inputs are the wrong type, and it seemed messy and unclear what was even being tested. So I just deleted the test. But I can put it back and catch the `SympifyError` instead if we're hesitant to do that.

pennylane_qiskit/converter.py
lillian542: Oooh, good point. I'll fix it so it works both ways.

lillian542:
[sc-49576]

pennylane_qiskit/converter.py
```diff
@@ -179,7 +179,9 @@ def _function(params: dict = None, wires: list = None):
 
         # Processing the dictionary of parameters passed
         for op, qargs, cargs in qc.data:
-            instruction_name = op.__class__.__name__
+            # the new Singlton classes have different names than the objects they represent, but base_class.__name__ still matches
```
trbromley, line 182:
```suggestion
            # the new Singleton classes have different names than the objects they represent, but base_class.__name__ still matches
```

tests/test_converter.py
timmysilv: that's fine by me. we have lots of tests in our plugins that don't really test the plugin itself, so I'm quite happy with removing the test. thanks for the details 😄

pennylane_qiskit/converter.py
```diff
@@ -179,7 +179,9 @@ def _function(params: dict = None, wires: list = None):
...
+            op_class = getattr(op, "base_class", None) or op.__class__
+            instruction_name = op_class.__name__
```
timmysilv, line 184:
```suggestion
            instruction_name = getattr(op, "base_class", op.__class__).__name__
```

timmysilv (approved):

PR #348, opened 2023-11-10
avoid deprecated pl features

timmysilv:
as title says. found in PennyLaneAI/plugin-test-matrix#57


albi3ro (approved):
Would be nice to add the warnings to errors filter to the `pytest.ini`, but that's not blocking.

PR #350, opened 2023-11-21
New Qiskit device prototype

lillian542:
The initial PR creating a new remote Qiskit device based on the new PennyLane device API. 

Does not currently work with state based simulators, and there will be a series of followup PRs to address missing features and gaps. 

This device interacts with a remote backend (sample-based simulator or hardware) and can either use the old Qiskit API or the new `Primitives` API. The `Primitives` API is used only for compatible measurement types (`Estimator` for `expval` and `var`, `Sampler` for `probs`); other PennyLane measurements will continue to use the old raw samples measurement procedure for the time being. The Qiskit Primitives currently don't support returning raw samples, but might in 1.0.

The preferred way of specifying execution details for the `Primitives` interface is with a Qiskit `Options` object, but the device also supports the old PennyLane format of kwargs, at least for the time being. There are some details surrounding handling potential overlap in how things are specified.

The device can use a Qiskit `Session` that is attached to all measurements (regardless of measurement type), to avoid queuing repeatedly for sequential measurements.


pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,484 @@
...
+    Keyword Args:
+        shots (int or None): number of circuit evaluations/random samples used
+            to estimate expectation values and variances of observables.
+        use_primitives(bool): whether or not to use Qiskit Primitives and the Qiskit Runtime Session. Defaults to False.
+    """
+    name = "Qiskit PennyLane plugin"
+    pennylane_requires = ">=0.30.0"
+    version = __version__
+    plugin_version = __version__
+    author = "Xanadu"
```
trbromley, line 187: are all of these needed with the new dev API?

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,484 @@
...
+        self._backend = backend
+
+        self.runtime_service = QiskitRuntimeService(channel="ibm_quantum")
+        self._use_primitives = use_primitives
+
+        # Keep track if the user specified analytic to be True
+        if shots is None:
+            # Raise a warning if no shots were specified for a hardware device
+            warnings.warn(f"The analytic calculation of expectations, variances and probabilities "
+                          f"is only supported on statevector backends, not on the {backend}. Such "
```
trbromley, line 216: Will this work now that `backend` is not a string?

trbromley (commented):
Awesome!

pennylane_qiskit/qiskit_device2.py
lillian542: Ah, good point, this needs to be `backend.name`

pennylane_qiskit/qiskit_device2.py
lillian542: I believe no

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,604 @@
...
+        # so diagonalizing gates and measurements are not included in the circuit
+        qcirc = circuit_to_qiskit(circuit, self.num_wires, diagonalize=False, measure=False)
+
+        estimator = Estimator(session=session, options=self.options)
+
+        # split into one call per measurement
+        # could technically be more efficient if there are some observables where we ask
+        # for expectation value and variance on the same observable, but spending time on
+        # that right now feels excessive
+        pauli_observables = [mp_to_pauli(mp, self.num_wires) for mp in circuit.measurements]
```
lillian542, line 555: need to sort differently for cases where the observable is not compatible with  a SparsePauliOp representation

pennylane_qiskit/converter.py
```diff
@@ -256,3 +296,120 @@ def load_qasm_from_file(file: str):
...
+        return qc
+
+    creg = ClassicalRegister(register_size)
+    qc = QuantumCircuit(reg, creg, name="temp")
+
+    for op in circuit.operations:
+        qc &= operation_to_qiskit(op, reg, creg)
+
+    # rotate the state for measurement in the computational basis
+    if diagonalize:
```
lillian542, line 341: follow up: check this in cases with multiple different bases

lillian542:
[sc-54383]

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,606 @@
...
+        # missing: split non-commuting, sum_expand, etc.
+
+        if self._use_primitives:
+            transform_program.add_transform(split_measurement_types)
+
+        return transform_program, config
+
+    def _update_kwargs(self):
+        """Combine the settings defined in options and the settings passed as kwargs, with
+        the definition in options taking precedence if there there is conflicting information"""
```
astralcai, line 386: > if there there is ...

Seems to be a typo here.

astralcai (commented):
Looks good. Here are some suggestions, mostly stylistic.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,606 @@
...
+            # if this was not a session set on the device, but just one created for this execution, close
+            session.close()
+
+        return results
+
+    def _execute_runtime_service(self, circuits, session):
+        """Execution using old runtime_service (can't use runtime sessions)"""
+
+        # update kwargs in case Options has been modified since last execution
+        self._update_kwargs()
```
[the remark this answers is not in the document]
timmysilv: I second this question. and if it is needed, can we move the `shots` warning from `__init__` into that helper function? would rather not have it tracked twice, especially because there is a magic number in there

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,606 @@
...
+
+        # in case a single circuit is passed
+        if isinstance(circuits, (QuantumTape, QuantumScript)):
```
timmysilv, line 486:
```suggestion
        if isinstance(circuits, QuantumScript):
```

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,606 @@
...
+        qcirc = [
+            circuit_to_qiskit(circ, self.num_wires, diagonalize=True, measure=True)
+            for circ in circuits
+        ]
+        compiled_circuits = self.compile_circuits(qcirc)
+
+        program_inputs = {"circuits": compiled_circuits, "shots": self.shots.total_shots}
+
+        for kwarg in self._kwargs:
+            program_inputs[kwarg] = self._kwargs.get(kwarg)
```
timmysilv, line 498: any reason why you're using `get`? I figure it's already in there
```suggestion
        for kwarg, value in self._kwargs.items():
            program_inputs[kwarg] = value
```

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,606 @@
...
+            session_id=session.session_id,
+        )
+        self._current_job = job.result(decoder=RunnerResult)
+
+        results = []
+
+        for index, circuit in enumerate(circuits):
+            self._samples = self.generate_samples(index)
+            res = [
+                mp.process_samples(self._samples, wire_order=self.wires)
```
timmysilv, line 521: do we need to save it to the device?
```suggestion
            samples = self.generate_samples(index)
            res = [
                mp.process_samples(samples, wire_order=self.wires)
```

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,606 @@
...
+        # that right now feels excessive
+        pauli_observables = [mp_to_pauli(mp, self.num_wires) for mp in circuit.measurements]
+        result = estimator.run([qcirc] * len(pauli_observables), pauli_observables).result()
+        self._current_job = result
+        result = self._process_estimator_job(circuit.measurements, result)
+
+        return result
+
+    @staticmethod
+    def _process_estimator_job(measurements, job_result):
```
timmysilv, line 565: there can't be other measurements in these tapes because of the preprocessing, right?

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,606 @@
...
+        """
+        counts = self._current_job.get_counts()
+        # Batch of circuits
+        if not isinstance(counts, dict):
+            counts = self._current_job.get_counts()[circuit]
+
+        samples = []
+        for key, value in counts.items():
+            for _ in range(0, value):
+                samples.append(key)
```
timmysilv, line 605: might be more performant since there can be a lot of samples, but i'm not sure.
```suggestion
            samples.extend([key] * value)
```

tests/test_base_device.py
```diff
@@ -0,0 +1,947 @@
...
+        self.backend = backend
+        self.max_time = max_time
+        self.session_id = "123"
+
+
+try:
+    service = QiskitRuntimeService(channel="ibm_quantum")
+    backend = service.backend("ibmq_qasm_simulator")
+except:
+    backend = MockedBackend()
```
timmysilv, line 82: is it too messy to make this a pytest fixture? not really sure what the benefit is anyway

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,606 @@
...
+            qiskit.qiskit_ibm_runtime.Session
+        """
+        return self._session
+
+    @property
+    def num_wires(self):
+        return len(self.wires)
+
+    def update_session(self, session):
+        self._session = session
```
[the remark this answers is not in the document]
timmysilv: better yet, do we need this setter at all? we might be giving users too many ways to provide sessions

tests/test_base_device.py
```diff
@@ -0,0 +1,947 @@
...
+
+        operations = [qml.PauliX(0), qml.PauliY(1), qml.Hadamard(2), qml.CNOT([2, 1])]
+        qs = QuantumScript(operations, measurements=measurements)
+        tapes, reorder_fn = split_measurement_types(qs)
+
+        # operations not modified
+        assert np.all([tape.operations == operations for tape in tapes])
+
+        # measurements split as expected
+        [tape.measurements for tape in tapes] == expectation
```
timmysilv, line 294:
```suggestion
        assert [tape.measurements for tape in tapes] == expectation
```

tests/test_base_device.py
```diff
@@ -0,0 +1,947 @@
...
+            measurements=[
+                qml.expval(qml.PauliZ(0)),
+                qml.probs(wires=[0, 1]),
+                qml.counts(),
+                qml.sample(),
+            ],
+        )
+        tapes, reorder_fn = split_measurement_types(qs)
+
+        with patch.object(dev, "_execute_runtime_service", return_value="runtime_execute_res"):
```
timmysilv, line 744: I think you can get this `as obj_name` to avoid the `mocker.spy` below and keep it simple, but idk :p

timmysilv (commented):
looks really great, most comments are not such a big deal but alas i left many.

My big concern: this device is rather mutable, but we're trying to avoid that in PL. how much of this needs to be customizable on-the-fly (as opposed to at construction time), and can we remove whatever else remains? I don't want to block users in the name of ideology (I don't want to do anything in that name) but there are definite drawbacks we introduce when we allow it. At a minimum, I'd like warnings highlighting possible changes you can make to existing devices.

pennylane_qiskit/converter.py
```diff
@@ -256,3 +296,120 @@ def load_qasm_from_file(file: str):
...
+
+# diagonalize is currently only used if measuring
+# maybe always diagonalize when measuring, and never when not?
+# will this be used for a user-facing function to convert from PL to Qiskit as well?
+def circuit_to_qiskit(circuit, register_size, diagonalize=True, measure=True):
+    """Builds the circuit objects based on the operations and measurements
+    specified to apply.
+
+    Args:
+        circuit (Union[~.QuantumScript, ~.QuantumTape]): the circuit applied
```
lillian542, line 309:
```suggestion
        circuit (QuantumTape): the circuit applied
```

tests/test_base_device.py
lillian542: 😅

pennylane_qiskit/converter.py
```diff
@@ -12,25 +12,65 @@
...
+    "U3": ex.U3Gate,
+    "IsingZZ": ex.RZZGate,
+    "IsingYY": ex.RYYGate,
+    "IsingXX": ex.RXXGate,
+    "S": ex.SGate,
+    "T": ex.TGate,
+    "SX": ex.SXGate,
+    "Adjoint(S)": ex.SdgGate,
+    "Adjoint(T)": ex.TdgGate,
+    "Adjoint(SX)": ex.SXdgGate,
```
lillian542, line 70:
```suggestion
    "Adjoint(SX)": ex.SXdgGate,
    "Barrier": ex.Barrier,
```

pennylane_qiskit/qiskit_device2.py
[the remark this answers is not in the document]
lillian542: > Is there ever a need for changing device configurations, or are all options locked in at device creation? If everything is locked in, why do we still allow users to edit configurations through the options object?

Kwargs are locked in at initialization and always have been. The `Options` object is mutable and can be updated between executions, which is why is here. Long term I guess the idea is we move away from the kwargs and only have the mutable options, but we won't remove this functionality until Qiskit removes the functionality it accesses (so possibly never).

> if it is needed, can we move the shots warning from __init__ into that helper function?
Yes, good point, there's no need to do this twice

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,606 @@
...
+        if isinstance(mp, (ExpectationMP, VarianceMP)):
+            estimator.append((mp, i))
+        elif isinstance(mp, ProbabilityMP):
+            sampler.append((mp, i))
+        else:
+            no_prim.append((mp, i))
+
+    order_indices = [[i for mp, i in group] for group in [estimator, sampler, no_prim]]
+
+    tapes = []
```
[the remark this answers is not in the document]
lillian542: Currently it doesn't use tape shots at all, it uses the device shots, which is why I didn't notice this transform was losing `shots` information. Discussed offline. There are some complications with differences in how PL vs Qiskit pass and track shots. 

Current plan: Fix this so it doesn't accidentally delete shots. Then if they all the tapes have the same number of shots, treat that as the source of truth - update Options and use them. If they don't, raise a warning (shots have been distributed unevenly) and use device shots.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,606 @@
...
+            validate_measurements, sample_measurements=accepted_sample_measurement, name=self.name
+        )
+        transform_program.add_transform(
+            validate_observables,
+            stopping_condition=self.observable_stopping_condition,
+            name=self.name,
+        )
+
+        transform_program.add_transform(broadcast_expand)
+        # missing: split non-commuting, sum_expand, etc.
```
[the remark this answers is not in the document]
lillian542: yes

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,606 @@
...
+                execute_fn = self._execute_estimator
+            elif isinstance(circ.measurements[0], ProbabilityMP):
+                execute_fn = self._execute_sampler
+            else:
+                execute_fn = self._execute_runtime_service
+            results.append(execute_fn(circ, session))
+
+        if self._session is None:
+            # if this was not a session set on the device, but just one created for this execution, close
+            session.close()
```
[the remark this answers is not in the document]
lillian542: a context manager sounds like a good idea

pennylane_qiskit/qiskit_device2.py
lillian542: correct

tests/test_base_device.py
```diff
@@ -0,0 +1,947 @@
...
+class MockSession:
+    def __init__(self, backend, max_time=None):
+        self.backend = backend
+        self.max_time = max_time
+        self.session_id = "123"
+
+
+try:
+    service = QiskitRuntimeService(channel="ibm_quantum")
+    backend = service.backend("ibmq_qasm_simulator")
```
lillian542, line 80: deprecated, use different simulator

pennylane_qiskit/qiskit_device2.py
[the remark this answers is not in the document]
austingmhuang: Here I fixed this with some assumptions.

The assumption is that whenever the options.executions.shots does not equal to the dev.shots we will always assume that this was because the User either 1) Edited the options.executions.shots or 2) Set the options.execution.shots at initialization. The error is raised when update_kwargs is run and it simply raises the error if the dev.shots and the options.shots are different (and then sets the options to match the dev). 

The second assumption is that the user did not purposefully choose 4000 aka write options.executions.shots = 4000 when they were defining options. If they did do so, they would not get a warning that says that shots cannot be defined through Options. This is a bit challenging to fix, but the alternative would be to warn the user that they set a common default value in Qiskit in the Options() object.

E.g. dev = QiskitDevice2(wires=2, backend=backend, options=Options())
Warning: Default number of shots (4000) detected in Options. If this was intentional, note that Pennylane does not support setting shots through Options. Device shots are set to 1024 {dev.shots}

In fact, we wouldn't need this warning if it was set this way:
E.g. dev = QiskitDevice2(wires = 2, backend=backend), since in the initialization we would see that options = None and know that the 4000 shots is a default value.

pennylane_qiskit/qiskit_device2.py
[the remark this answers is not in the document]
austingmhuang: Fixed but for now it does not check whether or not if all the tapes have the same number of shots.  Simply assumes that they all have the same number of shots for now. I think that edge case can be fixed in a separate PR personally. (How easy is it to have a mash up of tapes with different shots? What would the user have to do to actually achieve this?)

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,606 @@
...
+
+    def update_session(self, session):
+        self._session = session
+
+    def reset(self):
+        self._current_job = None
+
+    def stopping_condition(self, op: qml.operation.Operator) -> bool:
+        """Specifies whether or not an observable is accepted by DefaultQubit."""
+        return op.name in self.operations
```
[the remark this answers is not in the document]
austingmhuang: From what I can tell, DefaultQubitDevice uses name comparison instead of type checking because the "accepted" observables are not limited to those of type qml.operation.Observable. The observables dict in DefaultQubitDevice also contains operators like "Sum", which is not of type qml.operation.Observable. To have the same/similar functionality, the type check solution would look something like this: (might be wrong) instance(op, (qml.operation.Observable, qml.ops.op_math.composite.Compositeop)). I guess it's just easier to keep a dictionary of allowed observables/operations and compare names than to track all the classes? Anyway, I edited the docstrings but left the original code as-is.

pennylane_qiskit/qiskit_device2.py
austingmhuang: Will address in a separate PR in the future

pennylane_qiskit/converter.py
austingmhuang: Will address in a separate PR in the future

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,606 @@
...
+
+        else:
+            for k, v in vars(val).items():
+                if v is not None:
+                    options_dict[k] = v
+
+    return options_dict
+
+
+class QiskitDevice2(Device):
```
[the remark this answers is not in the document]
austingmhuang: Will address in a separate PR in the future

lillian542 (commented):
Looking good, @austingmhuang! Left a few small comments, but I think the main thing is that the `shots` behaviour could use a little bit of a tweak, and then a few tests to check its behaviour.

Once you unpin Qiskit, you may see some failures that need to be addressed, but fingers crossed everything goes smoothly🤞And it should resolve most of the failures you are currently seeing.

requirements-ci.txt
```diff
@@ -1,4 +1,4 @@
 pennylane>=0.32
-qiskit
+qiskit<0.46
```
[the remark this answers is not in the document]
austingmhuang: Ok, that makes sense. I figured I needed to unpin something somewhere. Thanks!

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,625 @@
...
+
+        if isinstance(circuits, QuantumScript):
+            circuits = [circuits]
+
+        @contextmanager
+        def execute_circuits(session):
+            try:
+                for circ in circuits:
+                    if circ.shots:
+                        self.options.execution.shots = circ.shots.total_shots
```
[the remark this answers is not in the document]
austingmhuang: > Also a valid input in Pennylane would be `circuit(shots=[2, 10, 5])`, currently this would give 3 results with differing shot numbers in, for example, `default.qubit`.
> 
> That information is lost by using `.total_shots`, since those would be the sum of all the shots, in this case 17. I think that's a fine way to handle this for the Qiskit device, but maybe we should raise a warning in the case where `len(shots.shot_vector > 1)`, since we aren't giving people what they ask for in that case.

Ah, I totally didn't realize that that was a possible case in Pennylane. I'll raise a warning in this case and let the user know that that that really isn't supported by Qiskit and what was actually run.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,625 @@
...
+                "to 1024 shots. The analytic calculation of results is not supported on "
+                "this device. All statistics obtained from this device are estimates based "
+                "on samples.",
+                UserWarning,
+            )
+
+            shots = 1024
+
+        self.options = options or Options()
+        if self.options.execution.shots == 4000:  ## 4000 is default value in Qiskit.
```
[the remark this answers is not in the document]
austingmhuang: I thought we were gonna leave the default at 1024 for consistency across the devices, which makes sense to me. But changing to 4000 also makes sense to avoid ambiguity. Either way I think this can be addressed in a separate PR. Maybe just changing to 4000 here makes sense though.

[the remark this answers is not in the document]
lillian542: I'm not set on it either way, they're both a little inconvenient 🤷‍♀️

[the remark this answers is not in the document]
austingmhuang: I'll leave it as is for now.

pennylane_qiskit/qiskit_device2.py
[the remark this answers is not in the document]
austingmhuang: Addressed the shot tracking behaviour, added tests and also warning for shot vector. For testing the shots I did a bit of a hack by saving the result that comes from the IBM devices directly to device._current_job and then pulling the shots information from there to verify. 

e.g.
`assert dev._current_job.metadata[0]["shots"] == 5`

Could also make a getter function that gets us the shot information instead, not sure if that's worth the effort though. If you have any preference, let me know and I'll write a getter function and call that instead.

astralcai (commented):
Looks good overall. Just a few questions.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,628 @@
...
+        @contextmanager
+        def execute_circuits(session):
+            try:
+                for circ in circuits:
+                    if circ.shots and len(circ.shots.shot_vector) > 1:
+                        warnings.warn(
+                            "Setting multiple shots in circuit initialization is not supported for the Qiskit-Pennylane Plugin."
+                            f"The circuit will be run once with {circ.shots.total_shots} instead."
+                        )
+                    if isinstance(circ.measurements[0], (ExpectationMP, VarianceMP)):
```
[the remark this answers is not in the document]
austingmhuang: My understanding (and @lillian542 correct me if I'm wrong) is that `execute_runtime_service` cannot use sessions and we're using IBM's `circuit-runner` instead, which takes in batched circuits. For the other two, the user is expected to get the same functionality by creating a session via `qiskit_session`, which is what IBM encourages.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -0,0 +1,628 @@
...
+
+        qcirc = circuit_to_qiskit(circuit, self.num_wires, diagonalize=True, measure=True)
+        if circuit.shots:
+            self.options.execution.shots = circuit.shots.total_shots
+        sampler = Sampler(session=session, options=self.options)
+
+        result = sampler.run(qcirc).result()
+        self._current_job = result
+
+        # needs processing function to convert to the correct format for states, and
```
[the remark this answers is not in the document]
austingmhuang: I think [this one tracks it. ](https://app.shortcut.com/xanaduai/story/54389/a-translation-function-is-added-for-converting-the-qiskit-probability-output-dictionary-keys-integers-indicating-state-into)

tests/test_base_device.py
```diff
@@ -0,0 +1,1056 @@
...
+        exp_res1 = np.zeros(num_wires)
+        exp_res1[0] = 1
+
+        # the two expected results are in samples
+        assert exp_res1 in samples
+        assert exp_res0 in samples
+
+        # nothing else is in samples
+        assert [s for s in samples if not s in np.array([exp_res0, exp_res1])] == []
+
```
lillian542, line 999: Not blocking, but for good measure, it might be nice to add a few more lines in these tape shots tests, so that after each call with `circuit(shots=[5])`, you call it again with no shots and confirm it returns to the default rather than being permanently modified.

Also let's add a short docstring to each - it makes things much nicer for people coming back to troubleshoot if we break something a year from now 😅 Just something like "Tests that if shots are passed to the circuit at execution, they override the default device shots for that execution for the [insert which type of execute function here] function."

lillian542 (commented):
Nice work @austingmhuang! I left a few optional suggestions, see what you think - but at this point I would approve if I could 🚀

tests/test_base_device.py
austingmhuang: > Not blocking, but for good measure, it might be nice to add a few more lines in these tape shots tests, so that after each call with `circuit(shots=[5])`, you call it again with no shots and confirm it returns to the default rather than being permanently modified.
I missed that. I think I did this while manually testing it myself but then just didn't do it for the actual tests for some reason. Should be a quick fix though.
 
> Also let's add a short docstring to each - it makes things much nicer for people coming back to troubleshoot if we break something a year from now 😅 Just something like "Tests that if shots are passed to the circuit at execution, they override the default device shots for that execution for the [insert which type of execute function here] function."
Will add docstrings! Forgot 😅

austingmhuang (approved):
Approving in place of @lillian542.

astralcai (approved):
No more questions from me! Looks good 🎸

Issue #337, opened 2023-10-03
Quasi distributions can have negative probabilities


fran-scala:
Hi! This is my first issue for Pennylane, I hope that my explanation will be clear enough. I was using the real IBM processor `ibmq_guadalupe` with the `pennylane-qiskit` plugin and discovered that a procedure to manage negative probabilities in quasi-probability distributions is missing. In particular, as stated on the [official Qiskit page](https://qiskit.org/documentation/how_to/use_sampler.html#get-the-probability-distribution) "A quasi-probability distribution differs from a probability distribution in that negative values are also allowed. However the quasi-probabilities must sum up to 1 like probabilities. Negative quasi-probabilities may appear when using error mitigation techniques.". 

Without a way to take care of negative values it would be impossible to generate samples with the method `generate_samples` which calls `sample_basis_states` since `np.random.choice` will complain about the negative probabilities.

I think that this can be fixed by using the [`nearest_probability_distribution`](https://qiskit.org/documentation/stubs/qiskit.result.QuasiDistribution.html#qiskit.result.QuasiDistribution.nearest_probability_distribution) method of the Qiskit class [`QuasiDistribution`](https://qiskit.org/documentation/stubs/qiskit.result.QuasiDistribution.html#qiskit.result.QuasiDistribution).

isaacdevlugt:
Hey @fran-scala!

This sounds like something that could be implemented directly in pennylane. Could you use pennylane directly for your needs? If so, I suggest making a [feature request](https://github.com/PennyLaneAI/pennylane/issues/new?assignees=&labels=enhancement+%3Asparkles%3A&projects=&template=feature_request.yml) on the pennylane repo. We have to allocate more resources to pennylane, which makes adding new features to plugins take more time.

If you could also make sure to share a code example that demonstrates what you'd like the feature to do, that would be extremely helpful 🙌

fran-scala:
This can be implemented in pennylane but it's already implemented in Qiskit, is it really necessary to build the feature again from scratch? 

One way to solve the problem is modifying `generate_samples` function in `runtime_devices.py`:

```
def generate_samples(self, circuit_id=None):
        r"""Returns the computational basis samples generated for all wires.

        Note that PennyLane uses the convention :math:`|q_0,q_1,\dots,q_{N-1}\rangle` where
        :math:`q_0` is the most significant bit.

        Args:
            circuit_id (int): position of the circuit in the batch.

        Returns:
[...]

PR #357, opened 2023-12-06
Solved issue #337 (this fix also #267)

fran-scala:
This change to `runtime_devices.py` avoid negative probabilities in quasi probability distributions (issue #337). It also includes the changes proposed in issue #267 which are still not implemented in the plugin.

### DIFF
--- a/pennylane_qiskit/runtime_devices.py
+++ b/pennylane_qiskit/runtime_devices.py
@@ -208,6 +208,7 @@ class IBMQSamplerDevice(IBMQDevice):
         probs = [0] * number_of_states
         # Fill in probabilities from counts: (state, prob) (e.g. ('010', 0.5))
         for state, prob in counts.items():
+            # Formatting all strings to the same lenght
             while len(state) < self.num_wires:
                 state = '0' + state[:]
             probs[int(state[::-1], 2)] = prob
--- a/pennylane_qiskit/runtime_devices.py
+++ b/pennylane_qiskit/runtime_devices.py
@@ -201,16 +201,20 @@ class IBMQSamplerDevice(IBMQDevice):
              array[complex]: array of samples in the shape ``(dev.shots, dev.num_wires)``
         """
         # We get nearest probability distribution because the quasi-distribution may contain negative probabilities
-        counts = self._current_job.quasi_dists[circuit_id].nearest_probability_distribution().binary_probabilities()
+        counts = (
+            self._current_job.quasi_dists[circuit_id]
+            .nearest_probability_distribution()
+            .binary_probabilities()
+        )
         # Since qiskit does not return padded string we need to recover the number of qubits with self.num_wires
-        number_of_states = 2 ** self.num_wires
+        number_of_states = 2**self.num_wires
         # Initialize probabilities to 0
         probs = [0] * number_of_states
         # Fill in probabilities from counts: (state, prob) (e.g. ('010', 0.5))
         for state, prob in counts.items():
             # Formatting all strings to the same lenght
             while len(state) < self.num_wires:
-                state = '0' + state[:]
+                state = "0" + state[:]
             # Inverting the order to recover Pennylane convention
             probs[int(state[::-1], 2)] = prob
         return self.states_to_binary(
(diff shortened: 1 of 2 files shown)


pennylane_qiskit/runtime_devices.py
```diff
@@ -200,27 +200,21 @@ def generate_samples(self, circuit_id=None):
...
-            for i in range(0, number_of_states):
-                if states[i] != i:
-                    states.insert(i, i)
-                    probs.insert(i, 0.0)
+        
+        # Initialize probabilities to 0
+        probs = [0] * number_of_states
+        # Fill in probabilities from counts: (state, prob) (e.g. ('010', 0.5))
+        for state, prob in counts.items():
+            probs[int(state, 2)] = prob
```
lillian542, line 217: It doesn't seem like this is caused by any of your changes, but let's go ahead and knock out a third issue with this function as well: it looks to me like the order of the returned wires is wrong. Qiskit has a different ordering convention for wires than PennyLane, and currently this is returning with the Qiskit convention, whereas all the other plugin devices use the PennyLane convention. If we run with `default.qubit`:

```
dev = qml.device("default.qubit", wires=3, shots=1024)

@qml.qnode(dev)
def circuit(x):
    qml.RX(x, wires=[0])
    return qml.counts()

circuit(np.pi)
```
we get
```
{'100': 1024}
```
whereas with 
```
dev = qml.device('qiskit.ibmq.sampler', wires=3, backend=backend)
```
we get
```
{'001': 1024}
```

One option to fix this is to just use the solution from `generate_samples` in the `circuit_runner` device above, i.e. instead of returning the results of `self.states_to_binary`, we save those as `samples` and return

```
np.vstack([np.array([int(i) for i in s[::-1]]) for s in samples])
```

lillian542 (commented):
Thanks, this looks great! I left a couple of comments, and you'll need to remove the whitespaces to make the CodeFactor bot happy, but I think this is nearly done 🎉 

Do add this as a bug-fix (or 2 or 3...) to the change-log, and don't forget to include your name as a contributor there as well!

pennylane_qiskit/runtime_devices.py
fran-scala: I think we can solve this by simply reversing the order of bits in the counts (minimal working change):
```
for state, prob in counts.items():
    probs[int(state[::-1], 2)] = prob
```
In this way, we switch to Pennylane convention and we can leave all the rest as it is.

lillian542: Unfortunately, the Qiskit states aren't padded with 0's up to the full state size, so the current solution fails unless the qubit that furthest left in the Qiskit standard is excited. That is, the `state` `'001'`, which we aim to convert to `'100'`, is at this point in the code represented as '1', `'010'` as `'10'`, etc.

Whatever solution you prefer for this, I would recommend running something like:

```
from qiskit_ibm_runtime import QiskitRuntimeService
service = QiskitRuntimeService()
backend = service.backend("ibmq_qasm_simulator")

dev_sampler = qml.device('qiskit.ibmq.sampler', wires=3, backend=backend)

@qml.qnode(dev_sampler)
def circuit(x, wire):
    qml.RX(x, wires=[wire])
    return qml.counts()
    
circuit(np.pi, 0), circuit(np.pi, 1), circuit(np.pi, 2)
```

and confirming the output matches running the same on `default.qubit` to ensure its working as expected. When I run this is doesn't cost me any of my Qiskit runtime, since I'm running on the QASM simulator.

fran-scala: Ok, these last commits should solve also the convention issue. Let me know if it is fine.

lillian542 (approved):
Thanks for working on this @fran-scala! Everything looks good to me, just missing a blank line at the end of the file to appease the bot 🤖

lillian542:
It looks like you also need to run `black -l 100 pennylane_qiskit` before the formatting check will pass

lillian542:
[sc-52032]

fran-scala:
Now everything should be ok!

PR #358, opened 2023-12-13
pass job_tags and session_id kwargs correctly

lillian542:
A user on the forum was trying to pass `job_tags` and `session_id` to the `circuit_runner` device, unsuccessfully. 

They were being included in generic runtime kwargs and passed to the runtime service in the `program_inputs` argument, which doesn't work; `job_tags` belongs in the `options` dictionary, and `session_id` is its own kwarg on `runtime_service.run`. Now they will be passed to the correct places.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,16 @@
 
 ### Bug fixes 🐛
 
+* The kwargs `job_tags` and `session_id` are passed to the correct arguments in the 
+  `circuit_runner` device so that they will be used in the Qiskit backend; these 
+  were previously ignored.
+  [(#358)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/358)
+
 ### Contributors ✍️
 
 This release contains contributions from (in alphabetical order):
+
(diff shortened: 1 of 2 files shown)


lillian542:
[sc-52032]

mudit2812 (commented):
Doesn't matter too much but is it possible to add testing for this?

lillian542:
Maybe if I try hard enough, but it looks like currently we don't test that kwargs **work** in the tests, just that passing them to the device doesn't cause the circuit to **fail**, so it would be a bit unprecedented.

I tried getting the information from `dev._current_job`, because that would be a quick and easy addition to the tests, but it doesn't include this in the metadata, so I'm not sure where/if I can access it, or if its just stored in their database for task retrieval.

Here's a screenshot where you can see that my tag ("now with a session") and session id were registered when the task ran.

image

mudit2812 (approved):
Thanks for checking! Looks good. Here's a 🚀

timmysilv (approved):
nice!

Issue #396, opened 2024-01-22
Qiskit operation conversion ignores `Adjoint`


lillian542:
**Describe the bug**
When we removed the `.inv` property, we removed the line that inverted the operation, and didn't replace it with anything - so now `Adjoint` means nothing on the Qiskit devices. 

This is in the base class, so it is true across all devices, and the logic is currently also inherited by the new device prototype.

**Expected behavior**
I expect the same result (ground state) when running this circuit:

```
@qml.qnode(dev)
def circuit(x):
    qml.RX(x, 0)
    qml.adjoint(qml.RX(x, 0))
    return qml.expval(qml.PauliZ(0))    
```

regardless of whether the device is `default.qubit` or a qiskit device running on the IBMQ QASM simulator. 

**Instead, the qiskit device just applies RX twice.** I used this device vs default qubit for a minimal working example:

```
from pennylane_qiskit import IBMQCircuitRunnerDevice
from qiskit_ibm_runtime import QiskitRuntimeService, Session

service = QiskitRuntimeService(channel="ibm_quantum")
backend = service.backend("ibmq_qasm_simulator")

dev = IBMQCircuitRunnerDevice(5, provider=None, backend=backend)
```

lillian542:
Tracked in SC as story #55173

lillian542:
Will be fixed for the prototype in #350 
Will be fixed for master in #397

PR #397, opened 2024-01-22
remove support for Adjoint so it decomposes

lillian542:
The Qiskit devices don't handle `Adjoint` since we removed `.inv` (with the exception of 3 gates with special implementations), but many `Adjoint` operators remained on the allowed operators list. Currently, `RX(1.23, 0)` and `Adjoint(RX(1.23, 0))` both apply `RX(1.23, 0)` on the device. 

This PR removes those, so they decompose in PL before being sent to the device, such that e.g. `Adjoint(RX(1.23, 0))` is applied as `RX(-1.23, 0)` . 

Fixes issue #396

### DIFF
--- a/pennylane_qiskit/qiskit_device.py
+++ b/pennylane_qiskit/qiskit_device.py
@@ -36,7 +36,7 @@ from ._version import __version__
 SAMPLE_TYPES = (SampleMP, CountsMP, ClassicalShadowMP, ShadowExpvalMP)
 
 
-QISKIT_OPERATION_MAP_SELF_ADJOINT = {
+QISKIT_OPERATION_MAP = {
     # native PennyLane operations also native to qiskit
     "PauliX": ex.XGate,
     "PauliY": ex.YGate,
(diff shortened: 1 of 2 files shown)


pennylane_qiskit/qiskit_device.py
```diff
@@ -65,30 +65,14 @@
     "IsingZZ": ex.RZZGate,
     "IsingYY": ex.RYYGate,
     "IsingXX": ex.RXXGate,
-}
-
-QISKIT_OPERATION_INVERSES_MAP_SELF_ADJOINT = {
-    "Adjoint(" + k + ")": v for k, v in QISKIT_OPERATION_MAP_SELF_ADJOINT.items()
```
lillian542, line 71: Removing this line fixes it, everything else is pretty much just tidying up

mudit2812 (approved):

timmysilv (approved):
elegant 🏆

PR #398, opened 2024-01-31
Add compile_backend kwarg

lillian542:
Add the `compile_backend` kwarg, to be used for circuit transpilation when using the old Qiskit API (i.e. when `use_primitives=False` or for raw-sample based measurements).

Also snuck in a little bit of testing for `transpilation_kwargs`, since it came up while updating the `compile_circuits` test to check that the `compile_backend` was used.


lillian542:
[sc-54387]

austingmhuang (approved):
Good work @lillian542. Just curious, when would a user want to use a different backend for transpilation compared to the actual backend they are running it on? Is it common for a User to transpile based on some real hardware (e.g. ibm_sherbrooke) and then run that circuit on a simulator/different hardware?

pennylane_qiskit/qiskit_device2.py
```diff
@@ -223,6 +223,9 @@ class QiskitDevice2(Device):
             for more details.
         session (Session): a Qiskit Session to use for device execution. If none is provided, a session will
             be created at each device execution.
+        compile_backend (Union[Backend, None]): the backend to be used for compiling the circuit that will be
+            sent to the backend device, to be set if the backend desired for compliation differs from the
+            backend used for execution. Defaults to None. If None, the primary backend will be used.
```
obliviateandsurrender, line 228:
```suggestion
            backend used for execution. Defaults to ``None``, which means the primary backend will be used.
```

pennylane_qiskit/qiskit_device2.py
```diff
@@ -445,7 +458,7 @@ def compile_circuits(self, circuits):
         transpile_args = self.get_transpile_args(self._kwargs)
 
         for i, circuit in enumerate(circuits):
-            compiled_circ = transpile(circuit, backend=self.backend, **transpile_args)
+            compiled_circ = transpile(circuit, backend=self.compile_backend, **transpile_args)
```
obliviateandsurrender, line 461: Nice!

pennylane_qiskit/qiskit_device2.py
```diff
@@ -302,6 +307,14 @@ def backend(self):
         """
         return self._backend
 
+    @property
+    def compile_backend(self):
+        """The compile_backend is a Qiskit backend object to be used for transpilation.
```
obliviateandsurrender, line 312:
```suggestion
        """The ``compile_backend`` is a Qiskit backend object to be used for transpilation.
```

obliviateandsurrender (approved):
Thanks @lillian542! Looking good, just a couple of suggestions related to docs.

lillian542:
> Good work @lillian542. Just curious, when would a user want to use a different backend for transpilation compared to the actual backend they are running it on? Is it common for a User to transpile based on some real hardware (e.g. ibm_sherbrooke) and then run that circuit on a simulator/different hardware?

In so far as possible, I think most people want to test some version of their circuits in simulation before switching to hardware, to debug any issues that can be resolved in simulator, since hardware runs cost money. My understanding is the use case would be, for example, that you want to run on `ibm_sherbrooke`, and you are testing out a toy model on simulator first, and you want to be as realistic as possible, and examine what circuits you will actually be running on the hardware.

That being said, I think it was more useful a few years ago when most of the hardware had fewer qubits than the simulators - now, for example, if you take a 127 qubit device (like Sherbrooke) and try to compile for it on a 32 qubit simulator (like ibm_qasm_simulator), you get an error saying that you don't have enough qubits on your simulator for a 127 qubit compilation. In the past, when (especially the free) hardware had a handful of qubits, and you could easily compile and test a circuit for the hardware on a simulator. So I think it's a bit outdated in terms of its usefulness.

PR #399, opened 2024-01-31
Support both V1 and V2

lillian542:
We rarely query the backends directly, and generally just pass them through to Qiskit. Current exceptions to this are checking the number of available qubits to verify against the number of wires, and getting the name of the backend for the Runtime Service execution method. 

This PR updates these two spots to work for both the new and the legacy Qiskit Backend API.


lillian542:
[sc-54390]

tests/test_base_device.py
```diff
@@ -100,9 +130,46 @@ def options_for_testing():
...
+        
+        @qml.qnode(dev)
+        def circuit(x):
+            qml.RX(x, wires=[0])
+            qml.CNOT(wires=[0, 1])
+            return qml.sample(qml.PauliZ(0))
+        
+        res = circuit(np.pi/2)
+        assert(isinstance(res, np.ndarray))
+        assert(np.shape(res) == (1024,))
```
obliviateandsurrender, line 172:
```suggestion
        assert isinstance(res, np.ndarray)
        assert np.shape(res) == (1024,)
```

tests/test_qiskit_device.py
```diff
@@ -13,6 +74,45 @@
...
+        
+        @qml.qnode(dev)
+        def circuit(x):
+            qml.RX(x, wires=[0])
+            qml.CNOT(wires=[0, 1])
+            return qml.sample(qml.PauliZ(0))
+        
+        res = circuit(np.pi/2)
+        assert(isinstance(res, np.ndarray))
+        assert(np.shape(res) == (1024,))
```
obliviateandsurrender, line 115:
```suggestion
        assert isinstance(res, np.ndarray)
        assert np.shape(res) == (1024,)
```

obliviateandsurrender (commented):
Thanks, looks great and should be good to go after a couple of minor suggested tweaks.

pennylane_qiskit/converter.py
```diff
@@ -34,9 +35,6 @@
 
 import pennylane as qml
 import pennylane.ops as pennylane_ops
-from pennylane_qiskit.qiskit_device import QISKIT_OPERATION_MAP
-
-# pylint: disable=too-many-instance-attributes
```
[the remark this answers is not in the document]
austingmhuang: Don't think we need this. I think I missed this when resolving merge conflicts (there were a lot).

austingmhuang (approved):
Fixed some of the minor issues @obliviateandsurrender caught. Thanks Utkarsh!

PR #405, opened 2024-02-06
Support for measurements operations

obliviateandsurrender:
This adds functionality to support measurements present in a qiskit's `QuantumCircuit` object or allows to override it using a keyword argument to the converter `load` function.

Example - 

``` py
# Building a Qiskit Circuit
qc = qk.QuantumCircuit(2, 2)
qc.h(0)
#qc.measure(0, 0)
qc.rz(0.543, [0])
qc.cx(0, 1)
qc.measure_all()

# Using the `load` functionality in PL
with qml.tape.OperationRecorder() as recorder:
    load(qc, measurements=None)()

>>> recorder.items()
((Hadamard(wires=[0]), {}),
 (measure(wires=[0]), {}),
 (RZ(0.543, wires=[0]), {}),
 (CNOT(wires=[0, 1]), {}),
 (Barrier(wires=[]), {}),
 (measure(wires=[0]), {}),
 (measure(wires=[1]), {}))

# Override it using the `keyword` argument
measurements = [qml.expval(qml.PauliZ(0)), qml.vn_entropy([0])]
with qml.tape.operation_recorder.AnnotatedQueue() as queue:
    load(qc, measurements=measurements)()
>>> queue
AnnotatedQueue([(Hadamard(wires=[0]), {}),
                (measure(wires=[0]), {}),
                (RZ(0.543, wires=[0]), {}),
                (CNOT(wires=[0, 1]), {}),
                (Barrier(wires=[]), {}),
                (expval(PauliZ(wires=[0])), {}),
                (vnentropy(wires=[0]), {})])


```

### DIFF
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -20,7 +20,8 @@ import warnings
 
 import numpy as np
 from qiskit import QuantumCircuit
-from qiskit.circuit import Parameter, ParameterExpression, Measure
+from qiskit.circuit import Parameter, ParameterExpression, Measure, Barrier
+from qiskit.circuit.library import GlobalPhaseGate
 from qiskit.exceptions import QiskitError
 from sympy import lambdify
 
@@ -148,6 +149,9 @@ def load(quantum_circuit: QuantumCircuit, measurements=None):
 
     Args:
         quantum_circuit (qiskit.QuantumCircuit): the QuantumCircuit to be converted
+        measurements (list[pennylane.measurements.MeasurementProcess]): the list of PennyLane
+            `measurements <https://docs.pennylane.ai/en/stable/introduction/measurements.html>`_
+            that overrides the terminal measurements that may be present in the input circuit.
 
     Returns:
         function: the resulting PennyLane template
@@ -176,7 +180,8 @@ def load(quantum_circuit: QuantumCircuit, measurements=None):
 
         wire_map = map_wires(qc_wires, wires)
 
-        num_inst, terminal_measurements = len(qc.data), []
+        # num_inst = len(qc.data)
+        mid_circuit_measurements, terminal_measurements = [], []
 
         # Processing the dictionary of parameters passed
         for idx, (op, qargs, _) in enumerate(qc.data):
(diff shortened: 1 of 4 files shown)


obliviateandsurrender:
[sc-56072]

pennylane_qiskit/converter.py
```diff
@@ -224,16 +225,43 @@ def _function(params: dict = None, wires: list = None):
                 gate = dagger_map[instruction_name]
                 qml.adjoint(gate)(wires=operation_wires)
 
+            elif isinstance(op, Measure):
+                meas_terminal = True
+                # Look-ahead for more gate(s) on its wire(s)
+                for meas_ix in range(idx, num_inst - 1):
```
lillian542, line 231: Just personal preference but I think this would be easier to read if we define a couple of lists, i.e.

```
instructions =[q.operation for q in qc.data]
qargs = [q.qubits for q in qc.data]
                
remaining_wires = [getattr(op, "base_class", op.__class__).__name__ for op in instructions[idx+1:]]
remaining_names = [set([wire_map[hash(q)] for q in qubits]) for qubits in qargs[idx+1]]
```

and then have the loop be over `name, wires in zip(remaining_names, remaining_wires)`. We could get rid of the `num_inst` variable, and the text below this would be much easier to read with just `name` and `wires` as variables.

I kind of want to re-organize the entire function to be easier to read though, so 🤷‍♀️

pennylane_qiskit/converter.py
```diff
@@ -224,16 +225,43 @@ def _function(params: dict = None, wires: list = None):
...
+                    (next_op, next_qargs, _), op_wires = qc.data[meas_ix + 1], set(operation_wires)
+                    next_op_name = getattr(next_op, "base_class", next_op.__class__).__name__
+                    if next_op_name not in ["Barrier", "GlobalPhaseGate"]:
+                        next_op_wires = set(wire_map[hash(qubit)] for qubit in next_qargs)
+                        if next_op_wires.intersection(op_wires):
```
lillian542, line 236:
```suggestion
                        if next_op_wires.intersection(set(operation_wires)):
```
I was momentarily confused trying to figure out how the contents of `op_wires` differ from `operation_wires`. Seems cleaner to just convert it to a set here, rather than making a new variable to represent the same information in set form. Then the `op_wires` definition can be removed from 232.

tests/test_converter.py
```diff
@@ -1110,3 +1114,49 @@ def cost(x, y):
         ]
 
         assert np.allclose(jac, jac_expected)
+
+    def test_meas_circuit_in_qnode(self, qubit_device_2_wires):
```
lillian542, line 1112: Can we add a test that specifically targets the behaviour of distinguishing between a mid-circuit and a terminal measurement? For example this circuit:

```
qc = QuantumCircuit(3, 3)
qc.h(0)
qc.measure(0, 0) 
qc.rx(0.2, [0])
qc.cx(0, 1)
qc.measure(1, 1)
```

versus this one:

```
qc = QuantumCircuit(3, 3)
qc.h(0)
qc.measure(3, 3) 
qc.rx(0.2, [0])
qc.cx(0, 1)
qc.measure(1, 1)
```

lillian542: Also, it seems for Qiskit that `qc.measure` goes to a classical register, where `qc.measure_all()` goes to a separate readout register unless you tell it to read out to the primary classical register. Does that influence how we interpret whether they are mid-circuit or terminal measurements? Just wondering if terminal measurements should be measurements read out to the `meas` register rather than being based on whether additional operations are performed. Not sure if this is for this PR or a followup PR regarding mid-circuit measurements, or if it even makes sense to consider. @trbromley

lillian542 (commented):
Thanks @obliviateandsurrender ! A couple of comments/suggestions, see what you think :)

pennylane_qiskit/converter.py
obliviateandsurrender: I have made it a bit easier to read and removed the `num_inst` variable. 
However, I prefer the current method instead of looping over individual lists, as the latter would make us re-iterate Qiskit's quantum circuit at least three times. This would add a computational cost if the circuit would be huge.

pennylane_qiskit/converter.py
obliviateandsurrender: Looks like I misplaced the `op_wires` last time. I wanted to keep it on the top so that I don't recompute it every time for hitting the `if` condition.

lillian542 (approved):
Thanks @obliviateandsurrender! Looks good to me :)

pennylane_qiskit/converter.py
```diff
@@ -224,16 +230,43 @@ def _function(params: dict = None, wires: list = None):
...
+                            meas_terminal = False
+                            break
+
+                # Allows for queing the mid-circuit measurements
+                if not meas_terminal:
+                    mid_circ_meas.append(qml.measure(wires=operation_wires))
+                else:
+                    terminal_meas.extend(operation_wires)
+
```
albi3ro, line 252:
```suggestion
                if meas_terminal:
                    terminal_meas.extend(operation_wires)
                else:
                    mid_circ_meas.append(qml.measure(wires=operation_wires))

```

Tends to be a bit more readable to control on the positive, since it's one less thing the brain needs to keep in active memory.

albi3ro (commented):
changelog?

albi3ro (approved):

PR #406, opened 2024-02-06
New UI for passing Parameter values

lillian542:
Old UI continues to be supported:

```
# Qiskit circuit
from qiskit.circuit import Parameter

theta = Parameter("theta")
phi = Parameter("phi")
psi = Parameter("psi")

qc = QuantumCircuit(1,1)
qc.rx(theta*phi, [0])
qc.rz(psi, [0])

# Conversion to PennyLane
dev = qml.device("default.qubit")

@qml.qnode(dev)
def circuit(x, y, z):
    qml.from_qiskit(qc)(params={theta: x, phi: y, psi:z})
```
or
```
qml.from_qiskit(qc)({theta: x, phi: y, psi:z})
```

but the new reccommended UI is also implemented, with options to pass as args, kwargs, or a combination:
```
qml.from_qiskit(qc)(y, z, x)  # order is alphabetical wrt parameter names
qml.from_qiskit(qc)(theta=x, phi=y, psi=z)
qml.from_qiskit(qc)(z, theta=x, phi=y)
```

If params are passed that don't match, we get clear errors for too many arguments:
```
>>> qml.from_qiskit(qc)(1, 2, 3, 4)
TypeError: Expected 3 positional arguments but 4 were given
```

```
>>> qml.from_qiskit(qc)(1, 2, 3, theta=4)
TypeError: Expected 2 positional arguments but 3 were given
```

for not enough arguments:
```
>>> qml.from_qiskit(qc)()
TypeError: Missing required arguments to define Parameter values for: phi, psi, theta
```

```
>>> qml.from_qiskit(qc)(phi=0.2, psi=0.3)
TypeError: Missing 1 required argument to define Parameter value for: theta
```

```
>>> qml.from_qiskit(qc)(0.2)
TypeError: Missing 2 required arguments to define Parameter values for: theta, psi
```

and unrecognized key word arguments:

```
>>> qml.from_qiskit(qc)(phi=0.2, psi=0.3, misspelled_kwarg=0.4)
TypeError: Got unexpected parameter keyword argument 'misspelled_kwarg'. Circuit contains parameters: phi, psi, theta
```


lillian542:
[sc-56075]

trbromley:
> qml.from_qiskit(z, theta=x, phi=y)

To confirm, is this part a typo that should be `qml.from_qiskit(qc)(z, theta=x, phi=y)`?

lillian542:
> To confirm, is this part a typo that should be qml.from_qiskit(qc)(z, theta=x, phi=y)?

Yes!

pennylane_qiskit/converter.py
```diff
@@ -35,17 +35,80 @@
...
+def _format_params_dict(quantum_circuit, params, *args, **kwargs):
+    """Processes the inputs for calling the quantum function and returns
+    a dictionary of the format {Parameter("name"): value} for all the parameters.
+
+    For a ``quantum_circuit`` with parameters ``[Parameter("phi"), Parameter("psi"), Parameter("theta")]``,
+    inputs can be one of the following:
+        1. the kwargs passed to when calling the qfunc, other than ``params`` and ``wires``.
+            The keys in kwargs are expected to correspond to the names of the Parameters on
+            the QuantumCircuit, i.e. (qc, None, phi=0.35, theta=0.2, psi=1.7)
+        2. the args passed when calling the qfunc, assigned to the Parameters in alphabetical
```
obliviateandsurrender, line 60: What do we mean by alphabetical order? Are we sorting them?

pennylane_qiskit/converter.py
```diff
@@ -35,17 +35,80 @@
...
+            if not qc_param:
+                raise TypeError(
+                    f"Got unexpected parameter keyword argument '{k}'. Circuit contains parameters {quantum_circuit.parameters}"
+                )
+            params[qc_param[0]] = v
+
+        # get any parameters not defined in kwargs (may be all of them) and match to args in order
+        arg_parameters = [p for p in quantum_circuit.parameters if p.name not in kwargs.keys()]
+        # if too many args were passed to the function call, raise an error
+        if len(args) > len(arg_parameters):
```
obliviateandsurrender, line 103: What happens in `<` case? Do we not need to worry about that?

pennylane_qiskit/converter.py
```diff
@@ -91,15 +154,38 @@ def _check_circuit_and_bind_parameters(
...
+        if name in [p.name for p in quantum_circuit.parameters]:
+            raise RuntimeError(
+                f"Cannot interpret QuantumCircuit with parameter '{name}' as a PennyLane "
+                f"quantum function, as this argument is reserved"
+            )
+
     if params is None:
+        if quantum_circuit.parameters:
+            raise TypeError(
+                f"Missing required argument{'s' if len(quantum_circuit.parameters) > 1 else ''} "
```
obliviateandsurrender, line 168: Maybe we can store `{'s' if len(quantum_circuit.parameters) > 1 else ''}` above so that we don't have to write it twice or simply use `argument(s)` lol.

pennylane_qiskit/converter.py
```diff
@@ -169,8 +256,11 @@ def _function(params: dict = None, wires: list = None):
         Returns:
             function: the new PennyLane template
         """
-        var_ref_map = _extract_variable_refs(params)
-        qc = _check_circuit_and_bind_parameters(quantum_circuit, params, var_ref_map)
+
+        # organize parameters, format trainable parameter values correctly, and then bind the parameters to the circuit
```
obliviateandsurrender, line 260: Nice! 🍪

obliviateandsurrender (changes requested):
Thanks @lillian542! Looks great, just leaving some initial feedback.

pennylane_qiskit/converter.py
lillian542: Unfortunately they are sorting them, and we mean alphabetical by the name of the parameter (designated when the user initialized the Qiskit parameter as `Parameter("name")`. I'll add a sentence expanding on this to make it clearer.

pennylane_qiskit/converter.py
```diff
@@ -35,17 +35,80 @@
...
+            to already be a dictionary of the format {Parameter("name"): value}, i.e.
+            (qc, None, {Parameter("phi"): 0.35, Parameter("psi"): 1.7, Parameter("theta"): 0.2})
+
+    Returns:
+        params (dict): A dictionary mapping ``quantum_circuit.parameters`` to values
+    """
+
+    # if nothing passed to params, and a dictionary has been passed as a single argument, then assume it is params
+    if params is None and (len(args) == 1 and isinstance(args[0], dict)):
+        params = args[0]
```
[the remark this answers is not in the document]
lillian542: Yes, we absolutely can - in theory the next condition should never be fulfilled if this condition is fulfilled, and it will just go to the return at the end of the function. Just adding the return here seems safer and easier to read.

pennylane_qiskit/converter.py
lillian542: It's checked further down in `_check_circuit_and_assign_parameters` on the dictionary this function returns, because it needs to be checked for the full `params`, regardless of which return from this function we end up exiting via. 

Ideally I would put all of the argument checks there so they are together, but this information is lost in line 107 when we zip `arg_parameters` and `args` together and drop anything at the end of args, so the specific one was moved here.

Its quite unintuitive though, I can add a comment in the code for future reference

pennylane_qiskit/converter.py
```diff
@@ -91,15 +154,38 @@ def _check_circuit_and_bind_parameters(
             "The circuit {} is not a valid Qiskit QuantumCircuit.".format(quantum_circuit)
         )
 
+    # confirm parameter names are valid for conversion to PennyLane
+    for name in ["wires", "params"]:
+        if name in [p.name for p in quantum_circuit.parameters]:
+            raise RuntimeError(
```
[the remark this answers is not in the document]
lillian542: No, it's not mentioned anywhere currently. This will only come up if users do something horrible like:

```
p = Parameter("wires")

qc = QuantumCircuit(2, 2)
qc.rx(p, 0)
qc.rx(p, 1)

qml.from_qiskit(qc)(wires=0.23)
```

The docstring of `_function` below definitely needs to be updated to reflect the new UI, and that would be a reasonable place to note it.

tests/test_converter.py
```diff
@@ -789,6 +829,98 @@ def test_barrier_not_supported(self, recorder):
...
+
+        with pytest.raises(
+            TypeError, match="Missing 1 required argument to define Parameter values"
+        ):
+            load(qc)(0.2, c=0.4)
+
+        with pytest.raises(
+            TypeError, match="Missing 2 required arguments to define Parameter values"
+        ):
+            load(qc)(b=0.3)
```
lillian542, line 908: note: should also add the old UI with too few arguments to this test (and the other similar ones)

pennylane_qiskit/converter.py
```diff
@@ -155,7 +241,8 @@ def load(quantum_circuit: QuantumCircuit):
         function: the resulting PennyLane template
     """
 
-    def _function(params: dict = None, wires: list = None):
+    # def _function(*args, params: dict = None, wires: list = None, **kwargs):
+    def _function(*args, params: dict = None, wires: list = None, **kwargs):
```
lillian542, line 245: remove commented line and update docstring

trbromley:
Thanks @lillian542 @obliviateandsurrender!

```python
qml.from_qiskit(qc)(y, z, x)  # order is alphabetical wrt parameter names
```

Could we get the order to depend on the order specified in the Qiskit circuit? I worry that alphabetical could be less intuitive.

```pycon
>>> qml.from_qiskit(qc)(phi=0.2, psi=0.3)
TypeError: Missing 1 required argument to define Parameter values for: {Parameter(theta)}
```
Would it make sense if it were instead `TypeError: Missing 1 required argument to define Parameter values for: theta`? Is there a motivation to wrap it in a set?

```pycon
>>> qml.from_qiskit(qc)(phi=0.2, psi=0.3, misspelled_kwarg=0.4)
TypeError: Got unexpected parameter keyword argument 'misspelled_kwarg'. Circuit contains parameters ParameterView([Parameter(phi), Parameter(psi), Parameter(theta)])
```
Not a huge blocker, but I'm not so keen on the repr for `ParameterView([Parameter(phi), Parameter(psi), Parameter(theta)])`.

lillian542:
> Could we get the order to depend on the order specified in the Qiskit circuit? I worry that alphabetical could be less intuitive.

We could. Here's why I didn't: The qiskit circuit stores parameters as `qc.parameters` alphabetically. If the user queries the Qiskit circuit for parameters, they will be sorted like this. If we take them in a different order:

1. our order won't match `qc.parameters`
2. we will have to generate the list of parameters in a less reliable, more hacky way, like using protected methods (`qc._parameter_table`), or iterating over the list of instructions, pulling the parameters out, and looping over to break down `ParameterExpression` elements until there are only `Parameter` and `ParameterVectorElement`s left.

Let me know what you think. The order they are used is an option.

> Would it make sense if it were instead TypeError: Missing 1 required argument to define Parameter values for: theta? Is there a motivation to wrap it in a set?

There's no motivation, it's just already a set when it gets to this point in the code, and I used it as is. I'll change them all to use just the names of the parameters. This also covers the final comment about the `ParameterView` repr - its just used as-is right now, I'll clean it up!

trbromley:
Ah I did mean to use the order specified by `qc.parameters`, I just didn't realize that that would be alphabetical! So how it's currently done sounds great.

tests/test_converter.py
```diff
@@ -768,19 +755,73 @@ def test_map_wires_exception_mismatch_in_number_of_wires(self, recorder):
         ):
             map_wires(qc_wires, wires)
 
+    def test_format_params_dict_old_interface(self):
+        """Test the old interface for setting the value of Qiskit Parameters -
+        passing a dictionary of the form {Parameter("name"): value, ...} as
```
obliviateandsurrender, line 760:
```suggestion
        passing a dictionary of the form ``{Parameter("name"): value, ...}`` as
```

tests/test_converter.py
```diff
@@ -768,19 +755,73 @@ def test_map_wires_exception_mismatch_in_number_of_wires(self, recorder):
...
+            ((0.3,), {"a": 0.5, "c": 0.4}),
+        ],
+    )
+    def test_format_params_dict_new_interface(self, args, kwargs):
+        """Test the new interface for setting the value of Qiskit Parameters -
+        passing either ordered args, or keyword arguments where the argument
+        matches the Parameter name, or some combination of the two.
+
+        The kwargs are passed as a dictionary to this function, and the args
+        as a tuple. This tests the new case where params=None"""
```
obliviateandsurrender, line 797:
```suggestion
        as a tuple. This tests the new case where `params=None`"""
```

pennylane_qiskit/converter.py
```diff
@@ -157,23 +246,80 @@ def load(quantum_circuit: QuantumCircuit, measurements=None):
...
+            b = Parameter("beta")
+            c = Parameter("gamma")
+
+            qc = QuantumCircuit(2, 2)
+            qc.rx(a, 0)
+            qc.rx(b * c, 1)
+
+            this_function = qml.from_qiskit(qc, measurements=[qml.PauliZ(0), qml.PauliZ(1)])
+
+        For the circuit above, based on Parameters with names alpha, beta and gamma, all of the following are
```
obliviateandsurrender, line 300:
```suggestion
        For the circuit above, based on ``Parameters`` with names 'alpha', 'beta' and 'gamma', all of the following are
```

tests/test_converter.py
```diff
@@ -789,6 +830,103 @@ def test_template_not_supported(self, recorder):
...
+
+        # works with correct name
+        load(qc)(name1=0.3)
+
+        # raises error with incorrect name
+        with pytest.raises(TypeError, match="Got unexpected parameter keyword argument 'name2'"):
+            load(qc)(name2=0.3)
+
+    def test_too_many_args(self):
+        """Test that if too many positional arguments are passed to define Parameter values,
```
obliviateandsurrender, line 867:
```suggestion
        """Test that if too many positional arguments are passed to define ``Parameter`` values,
```

pennylane_qiskit/converter.py
```diff
@@ -36,17 +36,83 @@
...
     """
-    if isinstance(param, Parameter) and param not in var_ref_map:
-        raise ValueError(f"The parameter {param} was not bound correctly.")
+    if isinstance(param, Parameter) and param not in trainable_params:
+        raise ValueError(f"The parameter {param} was not bound correctly.".format(param))
+
+
+def _format_params_dict(quantum_circuit, params, *args, **kwargs):
+    """Processes the inputs for calling the quantum function and returns
+    a dictionary of the format {Parameter("name"): value} for all the parameters.
```
obliviateandsurrender, line 54:
```suggestion
    a dictionary of the format ``{Parameter("name"): value}`` for all the parameters.
```

obliviateandsurrender (commented):
Just some docs suggestions, the rest looks good to me!

pennylane_qiskit/converter.py
```diff
@@ -36,17 +36,83 @@
...
+        # get any parameters not defined in kwargs (may be all of them) and match to args in order
+        arg_parameters = [p for p in quantum_circuit.parameters if p.name not in kwargs.keys()]
+        # if too many args were passed to the function call, raise an error
+        # all other checks regarding correct arguments will be processed in _check_circuit_and_assign_parameters
+        # (based on the full params dict generated by this function), but this information can only be captured here
+        if len(args) > len(arg_parameters):
+            raise TypeError(
+                f"Expected {len(arg_parameters)} positional argument{'s' if len(arg_parameters) > 1 else ''} but {len(args)} were given"
+            )
+        params.update(dict(zip(arg_parameters, args)))
```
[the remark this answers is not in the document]
albi3ro: Running pylint on this file gives:
```
pennylane_qiskit/converter.py:103:81: C0201: Consider iterating the dictionary directly instead of calling .keys() (consider-iterating-dictionary)
pennylane_qiskit/converter.py:333:8: R1702: Too many nested blocks (6/5) (too-many-nested-blocks)
pennylane_qiskit/converter.py:249:4: R0912: Too many branches (19/12) (too-many-branches)
```

So basically this is becoming too nested and branched. Basically it's time to start pulling out helper functions.

pennylane_qiskit/converter.py
```diff
@@ -36,17 +36,83 @@
...
 
-def _check_parameter_bound(param: Parameter, var_ref_map: Dict[Parameter, Any]):
+def _check_parameter_bound(param: Parameter, trainable_params: Dict[Parameter, Any]):
     """Utility function determining if a certain parameter in a QuantumCircuit has
     been bound.
 
     Args:
         param (qiskit.circuit.Parameter): the parameter to be checked
-        var_ref_map (dict[qiskit.circuit.Parameter, Any]):
+        trainable_params (dict[qiskit.circuit.Parameter, Any]):
```
albi3ro, line 45: Seems like a pre-existing naming choice, but `trainable_params` does already have a connotation of `QuantumScript.trainable_params`, which is a different kind of thing.

pennylane_qiskit/converter.py
```diff
@@ -36,17 +36,83 @@
...
+            qc_param = [p for p in quantum_circuit.parameters if p.name == k]
+            if not qc_param:
+                param_names = ", ".join([p.name for p in quantum_circuit.parameters])
+                raise TypeError(
+                    f"Got unexpected parameter keyword argument '{k}'. Circuit contains parameters: {param_names}"
+                )
+            params[qc_param[0]] = v
+
+        # get any parameters not defined in kwargs (may be all of them) and match to args in order
+        arg_parameters = [p for p in quantum_circuit.parameters if p.name not in kwargs.keys()]
```
albi3ro, line 103:
```suggestion
        arg_parameters = [p for p in quantum_circuit.parameters if p.name not in kwargs]
```

something pylint is locally complaining about.

pennylane_qiskit/converter.py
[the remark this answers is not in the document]
albi3ro: So pylint and sourcery are definitely complaining about this file, but they don't seem to be concerned about this function. And it doesn't seem particularly nested or branchy to me... 🤔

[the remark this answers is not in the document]
albi3ro: I wonder if its placing the warning in the wrong place, since `_function` is overly nested, branchy, and overall low code quality.

pennylane_qiskit/converter.py
```diff
@@ -157,23 +246,80 @@ def load(quantum_circuit: QuantumCircuit, measurements=None):
         function: the resulting PennyLane template
     """
 
-    # pylint:disable=fixme, too-many-branches
-    def _function(params: dict = None, wires: list = None):
+    def _function(*args, params: dict = None, wires: list = None, **kwargs):
         """Returns a PennyLane template created based on the input QuantumCircuit.
```
albi3ro, line 250: Should we continue saying "template"? I think this term might have a slightly different connotation now, that this function doesn't really match.

pennylane_qiskit/converter.py
```diff
@@ -36,17 +36,83 @@
...
+    Returns:
+        params (dict): A dictionary mapping ``quantum_circuit.parameters`` to values
+    """
+
+    # if no kwargs are passed, and a dictionary has been passed as a single argument, then assume it is params
+    if params is None and not kwargs and (len(args) == 1 and isinstance(args[0], dict)):
+        return args[0]
+
+    # make params dict if using args and/or kwargs
+    if args or kwargs:
```
albi3ro, line 81: Could potentially use a guard-clause here and reduce the indentation of this next block:
```
if not args and not kwargs:
    return params
if params is not None:
```

pennylane_qiskit/converter.py
```diff
@@ -157,23 +246,80 @@ def load(quantum_circuit: QuantumCircuit, measurements=None):
...
+            params (dict): a dictionary with Parameter keys, specifying the parameters that need
+                to be bound in the QuantumCircuit
             wires (Sequence[int] or int): The wires the converted template acts on.
                 Note that if the original QuantumCircuit acted on :math:`N` qubits,
                 then this must be a list of length :math:`N`.
 
         Returns:
             function: the new PennyLane template
+
+        .. warning::
```
albi3ro, line 275: I wonder if its worth moving these warnings and examples into the docstring for `load`, as that is the docstring that will render in the website.

pennylane_qiskit/converter.py
```diff
@@ -157,23 +246,80 @@ def load(quantum_circuit: QuantumCircuit, measurements=None):
...
+        .. code-block:: python
+
+            # positional arguments passed in alphabetical order (alpha, beta, gamma)
+            circuit = this_function(np.pi, 0.5, np.pi)
+
+            # kwargs are parameter names
+            circuit = this_function(alpha=np.pi, beta=0.5, gamma=np.pi)
+
+            # kwargs and args can be combined
+            circuit = this_function(np.pi, np.pi, beta=0.5)
```
albi3ro, line 312: If we have a function `f(a, b, c)`, we can't do `f(1.0, 1.0, b=2.0)`, since `b` is the second positional argument.  But I'm fine with leaving this in as is.

albi3ro (approved):
Potentially may need to silence code factor, but my comments are purely optional and non-blocking.  Looks like a much cleaner syntax :)

CHANGELOG.md
```diff
@@ -23,6 +29,8 @@
 This release contains contributions from (in alphabetical order):
 
 Utkarsh Azad
+Lillian Frederiksen
+
```
trbromley, line 33:
```suggestion
```

CHANGELOG.md
```diff
@@ -4,12 +4,18 @@
...
+  a dictionary. The old dictionary UI continues to be supported.
+  [(#406)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/406)
+
 * Measurement operations are now added to the PennyLane template when a `QuantumCircuit`
   is converted using `load`. Additionally, one can override any existing terminal
   measurements by providing a list of PennyLane 
   `measurements <https://docs.pennylane.ai/en/stable/introduction/measurements.html>`_ themselves.
   [(#405)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/405)
 
+
```
lillian542, line 18:
```suggestion
```

pennylane_qiskit/converter.py
lillian542, line 250:
```suggestion
        """Returns a PennyLane quantum function created based on the input QuantumCircuit.
```

pennylane_qiskit/converter.py
lillian542: I think we definitely want them here, because this is what you get if you "ask" the function what it does in Jupyter, i.e.

```
qfunc = qml.from_qiskit(qc)
qfunc?
```

I think we want something more extensive in `qml.from_qiskit`, which will be the documentation on the PennyLane website. @trbromley , do we also want to expand documentation on the `load` function in the plugin, or is that more internal?

pennylane_qiskit/converter.py
lillian542: renamed to `unbound_params`

pennylane_qiskit/converter.py
lillian542: changed to "quantum function"

pennylane_qiskit/converter.py
```diff
@@ -157,23 +246,80 @@ def load(quantum_circuit: QuantumCircuit, measurements=None):
         function: the resulting PennyLane template
     """
 
-    # pylint:disable=fixme, too-many-branches
-    def _function(params: dict = None, wires: list = None):
-        """Returns a PennyLane template created based on the input QuantumCircuit.
+    def _function(*args, params: dict = None, wires: list = None, **kwargs):
```
lillian542, line 249:
```suggestion
    # pylint:disable=too-many-branches
    def _function(*args, params: dict = None, wires: list = None, **kwargs):
```

pennylane_qiskit/converter.py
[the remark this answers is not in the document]
lillian542: I'm going to ignore it for now, looks like the CodeFactor check still passes, and there are quite a few changes to this function in the next PR - I don't really want to create a bunch of merge conflicts by moving things around here.

obliviateandsurrender (approved):

pennylane_qiskit/converter.py
trbromley: I like that you can query `qfunc` in Jupyter and get that info! And yes we do want the `from_qiskit` docs to be on point - though we can treat it as a follow up to revisit those docs. I think the `load` function in the plugin can be made a copy of whatever is in `from_qiskit`.

PR #417, opened 2024-02-10
Improve mid-circuit measurement conversion abilities

obliviateandsurrender:
- **Context** - Qiskit circuits with mid-circuit measurements are supported
- **Changes** - Restructured the conversion iteration to add support for conditional operations based on mid-circuit measurements.
    1. Supported - `c_if` (legacy), `IfElseOp`, and `SwitchCaseOp`
    2. Not supported - `ForLoopOp` and `WhileLoopOp`, mainly because doing them with `qml.cond` doesn't seem possible at the moment.


obliviateandsurrender:
[sc-56131]

pennylane_qiskit/converter.py
```diff
@@ -49,6 +51,26 @@ def _check_parameter_bound(param: Parameter, unbound_params: Dict[Parameter, Any
         raise ValueError(f"The parameter {param} was not bound correctly.".format(param))
 
 
+def _process_trivial_param_args(params, *args, **kwargs):
```
albi3ro, line 54: I'm a little confused by what this function is doing.  What's considered trivial? What are the two pieces of output?

pennylane_qiskit/converter.py
```diff
@@ -328,56 +322,59 @@ def _function(*args, params: dict = None, wires: list = None, **kwargs):
...
-                            pl_parameters.append(float(p))  # pragma: no cover
-                    else:
-                        pl_parameters.append(p)
-
-                execute_supported_operation(
-                    inv_map[instruction_name], pl_parameters, operation_wires
-                )
+            # Define operator builders and helpers
+            operation_func = None
+            operation_overlapper = lambda op: op
```
albi3ro, line 340:
```suggestion
            def operation_overlapped(op):
                return op
```

pennylane_qiskit/converter.py
obliviateandsurrender: The codefactor was complaining about the `_format_params_dict` being too `complex` and failing. So, I separated out some initial conditions into this separate function. About the output - the first one is the `dict` of the parameters, and the second one is a bool `flag` if these params are ready to be returned.

pennylane_qiskit/converter.py
```diff
@@ -328,56 +327,62 @@ def _function(*args, params: dict = None, wires: list = None, **kwargs):
...
+                        ordered_params = tuple(p.parameters)
+                        f = lambdify(ordered_params, p._symbol_expr, modules=qml.numpy)
+                        f_args = []
+                        for i_ordered_params in ordered_params:
+                            f_args.append(unbound_params.get(i_ordered_params))
+                        operation_params.append(f(*f_args))
+                    else:  # needed for qiskit<0.43.1
+                        operation_params.append(float(p))  # pragma: no cover
+                else:
+                    operation_params.append(p)
```
albi3ro, line 372: Looks like this could be extracted into a helper function `operation_params = _get_operation_params(ops)`.

pennylane_qiskit/converter.py
```diff
@@ -328,56 +327,62 @@ def _function(*args, params: dict = None, wires: list = None, **kwargs):
...
+        mid_circ_meas, mid_circ_regs = [], {}
 
         # Processing the dictionary of parameters passed
-        for idx, (op, qargs, _) in enumerate(qc.data):
-            # the new Singleton classes have different names than the objects they represent, but base_class.__name__ still matches
-            instruction_name = getattr(op, "base_class", op.__class__).__name__
-
-            operation_wires = [wire_map[hash(qubit)] for qubit in qargs]
-
+        for idx, (ops, qargs, cargs) in enumerate(qc.data):
```
albi3ro, line 334: Is `ops` multiple operations or a single operation?

lillian542 (commented):
Thanks for the walkthrough, here are my initial thoughts from yesterday - I'll go through it again this afternoon :)

pennylane_qiskit/converter.py
```diff
@@ -328,56 +327,62 @@ def _function(*args, params: dict = None, wires: list = None, **kwargs):
...
-                    inv_map[instruction_name], pl_parameters, operation_wires
-                )
+            # Define operator builders and helpers
+            operation_func = None
 
-            elif instruction_name in dagger_map:
-                gate = dagger_map[instruction_name]
-                qml.adjoint(gate)(wires=operation_wires)
+            def operation_overlapper(op):
+                return op
```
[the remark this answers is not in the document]
lillian542: Actually - it looks like since it didn't work out for this to also include `conds`, this is only ever either "do nothing" or `adjoint`. Then when it's finally applied, it's applied as `operation_overlapper(operation_func)(*operation_args, **operation_kwargs)`. 

Can't we just apply `adjoint` to `operation_func` the relevant spot, i.e. if line 374 is:

```
if instruction_name in dagger_map:
    operation_func = qml.adjoint(dagger_map[instruction_name])
```

and then we don't need the `operation_overlapper` at all?

lillian542 (commented):
a few more little things here and there

pennylane_qiskit/converter.py
obliviateandsurrender: It's a single operation. Changed it to `instruction`.

pennylane_qiskit/converter.py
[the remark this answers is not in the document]
obliviateandsurrender: I thought so, too! Removed the `overlapper` based on this.

pennylane_qiskit/converter.py
```diff
@@ -328,56 +327,62 @@ def _function(*args, params: dict = None, wires: list = None, **kwargs):
         wire_map = map_wires(qc_wires, wires)
 
         # Stores the measurements encountered in the circuit
-        mid_circ_meas, terminal_meas = [], []
+        terminal_meas = []
+        mid_circ_meas, mid_circ_regs = [], {}
```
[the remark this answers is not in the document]
obliviateandsurrender: Added a comment for the explanation. It is basically a dictionary that maps the classical register used by Qiskit to the measurements we are doing so that we know which ones to use for conditioning the control flows.

pennylane_qiskit/converter.py
```diff
@@ -391,22 +396,91 @@ def _function(*args, params: dict = None, wires: list = None, **kwargs):
...
+
                 except (AttributeError, QiskitError):
                     warnings.warn(
                         f"{__name__}: The {instruction_name} instruction is not supported by PennyLane,"
                         " and has not been added to the template.",
                         UserWarning,
                     )
+
+            # Check if it is a conditional operation
+            if operation_cond or (ops.condition and ops.condition[0] in mid_circ_regs):
```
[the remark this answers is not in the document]
obliviateandsurrender: I don't think I will have extra logic until we begin to add support for more `ControlFlowOps`. So, for now I can remove it and use `isinstance` instead.

pennylane_qiskit/converter.py
```diff
@@ -391,22 +396,91 @@ def _function(*args, params: dict = None, wires: list = None, **kwargs):
...
+                # Iteratively recurse over to build different branches
+                with qml.QueuingManager.stop_recording():
+                    branch_funcs = [
+                        partial(load(branch_inst, measurements=None), params=params, wires=wires)
+                        for branch_inst in operation_params
+                        if isinstance(branch_inst, QuantumCircuit)
+                    ]
```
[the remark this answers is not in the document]
obliviateandsurrender: In reality, it's just another parameter. I am calling it `branch_inst` because it is an `Instruction` on what to do in the conditional branch. About the `branch_funcs` - I wanted to indicate that each function corresponds to a specific branch which `qml.cond` would use. I can call it `branch_qfuncs`.

pennylane_qiskit/converter.py
```diff
@@ -328,56 +327,62 @@ def _function(*args, params: dict = None, wires: list = None, **kwargs):
...
-                        else:  # needed for qiskit<0.43.1
-                            pl_parameters.append(float(p))  # pragma: no cover
-                    else:
-                        pl_parameters.append(p)
-
-                execute_supported_operation(
-                    inv_map[instruction_name], pl_parameters, operation_wires
-                )
+            # Define operator builders and helpers
+            operation_func = None
```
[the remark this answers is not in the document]
obliviateandsurrender: For sure, I changed it to `operation_class`!

pennylane_qiskit/converter.py
```diff
@@ -391,22 +396,91 @@ def _function(*args, params: dict = None, wires: list = None, **kwargs):
...
+
+                # Get the functions for handling condition
+                true_fn, false_fn, elif_fns, cond_op = _conditional_funcs(
+                    ops, cargs, operation_func, branch_funcs, instruction_name
+                )
+                res_reg, res_bit = cond_op
+
+                # Check for multi-qubit register
+                if tuple(cargs) not in mid_circ_regs:
+                    ctrl_cargs = min(len(cargs), len(qargs))
```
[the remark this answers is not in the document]
obliviateandsurrender: Added a small note regarding it.

pennylane_qiskit/converter.py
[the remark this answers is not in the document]
lillian542: no, that seems great

pennylane_qiskit/converter.py
```diff
@@ -366,56 +402,46 @@ def _function(*args, params: dict = None, wires: list = None, **kwargs):
...
-                    else:
-                        pl_parameters.append(p)
-
-                execute_supported_operation(
-                    inv_map[instruction_name], pl_parameters, operation_wires
-                )
+            # Define operator builders and helpers
+            # _class -> PennyLane operation class object mapped from the Qiskit operation
+            # _args and _kwargs -> Parameters required for instantiation of `_class`
+            # _cond -> Flag regarding if we have encountered a classical control flow op
```
lillian542, line 424:
```suggestion
```

pennylane_qiskit/converter.py
```diff
@@ -474,3 +560,31 @@ def load_qasm_from_file(file: str):
         function: the new PennyLane template
     """
     return load(QuantumCircuit.from_qasm_file(file))
+
+
+# pylint:disable=fixme, protected-access
+def _conditional_funcs(ops, cargs, operation_class, branch_funcs, ctrl_flow_type):
+    """Builds the conditional functions for Controlled flows"""
```
lillian542, line 567: Can we expand on this docstring to include the args and returns?

pennylane_qiskit/converter.py
```diff
@@ -474,3 +560,31 @@ def load_qasm_from_file(file: str):
...
+    # Logic for handling SwitchCaseOp
+    elif ctrl_flow_type == "SwitchCaseOp":
+        elif_fns = []
+        for res_bit, case in ops._case_map.items():
+            if not isinstance(case, _DefaultCaseType):
+                elif_fns.append((res_bit, branch_funcs[case]))
+        ops.condition = [tuple(cargs), "SwitchCase"]
+        if any((isinstance(case, _DefaultCaseType) for case in ops._case_map)):
+            true_fn = branch_funcs[-1]
+            ops.condition = [tuple(cargs), "SwitchDefault"]
```
lillian542, line 588: Why modify the `condition` attribute on `ops` instead of just unpacking the `ops.condition in the return in line 571 and having this function return `res_reg` and `res_big` instead of `ops.condition`. Do we use `ops.condition` later?

lillian542 (approved):
Left a few minor suggestions, but nothing blocking. Looks good to me 🚀

pennylane_qiskit/converter.py
obliviateandsurrender: This allows for a standardized way of storing the info we require since `SwitchOp` by default is `None`, so we are not overriding any information. I don't exactly use the info later in this PR but do so in the `expr` one, so would like to keep it.

pennylane_qiskit/converter.py
```diff
@@ -366,56 +402,45 @@ def _function(*args, params: dict = None, wires: list = None, **kwargs):
...
-                            pl_parameters.append(float(p))  # pragma: no cover
-                    else:
-                        pl_parameters.append(p)
-
-                execute_supported_operation(
-                    inv_map[instruction_name], pl_parameters, operation_wires
-                )
+            # Define operator builders and helpers
+            # _class -> PennyLane operation class object mapped from the Qiskit operation
+            # _args and _kwargs -> Parameters required for instantiation of `_class`
```
albi3ro, line 423:
```suggestion
            # operation_class -> PennyLane operation class object mapped from the Qiskit operation
            # operation_args and operation_kwargs -> Parameters required for instantiation of `operation_class`
```

tests/test_converter.py
```diff
@@ -1397,6 +1399,104 @@ def test_diff_meas_circuit(self):
...
+                qml.PauliX(wires=0)
+
+            qml.cond(m0, ansatz_true, ansatz_false)()
+
+            qml.RZ(0.24, wires=0)
+            qml.CNOT([0, 1])
+
+            return [qml.expval(m) for m in [m0, m1, qml.measure(0), qml.measure(1), qml.measure(2)]]
+
+        assert loaded_qiskit_circuit() == built_pl_circuit()
```
albi3ro, line 1455:
```suggestion
        assert loaded_qiskit_circuit() == built_pl_circuit()
        assert qml.equal(loaded_qiskit_circuit.tape, built_pl_circuit.tape)
```
Might also be worth checking this for each of the tests.

tests/test_converter.py
```diff
@@ -1397,6 +1399,104 @@ def test_diff_meas_circuit(self):
...
+            qml.MultiControlledX(wires=[4, 3, 5, 1], control_values="001")
+            qml.MultiControlledX(wires=[4, 3, 5, 1], control_values="010")
+            qml.MultiControlledX(wires=[4, 3, 5, 2], control_values="011")
+            qml.MultiControlledX(wires=[4, 3, 5, 2], control_values="100")
+            qml.MultiControlledX(wires=[4, 3, 5, 2], control_values="101")
+            qml.MultiControlledX(wires=[4, 3, 5, 2], control_values="110")
+            qml.MultiControlledX(wires=[4, 3, 5, 2], control_values="111")
+            return [qml.expval(m) for m in [qml.measure(0), qml.measure(1), qml.measure(2)]]
+
+        assert loaded_qiskit_circuit() == built_pl_circuit()
```
albi3ro, line 1498:
```suggestion
        assert loaded_qiskit_circuit() == built_pl_circuit()
        assert qml.equal(loaded_qiskit_circuit.tape, built_pl_circuit.tape)
```

tests/test_converter.py
```diff
@@ -1397,6 +1399,104 @@ def test_diff_meas_circuit(self):
...
+
+    def test_control_flow_ops_circuit_switch(self):
+        """Tests mid-measurements are recognized and returned correctly."""
+
+        qreg = QuantumRegister(3)
+        creg = ClassicalRegister(3)
+        qc = QuantumCircuit(qreg, creg)
+        qc.measure([0, 1, 2], [0, 1, 2])
```
albi3ro, line 1463: Should we start of the circuit with a rotation to make the example a little less trivial? We are comparing results, but since we start in the zero state, a lot of those operations will have no effect on the outcome.

tests/test_converter.py
obliviateandsurrender: Tried doing it, and the operations comes out to be same but the `qml.equal` returns `False` because `qml.measure` assertion fails 🤔 

```
assert [Hadamard(wires=[0]), CNOT(wires=[0, 1]), measure(wires=[0]), Hadamard(wires=[0]), CNOT(wires=[0, 1]),
 measure(wires=[0]), Conditional(wires=[0]), Conditional(wires=[0]), Conditional(wires=[2]), RZ(0.24, wires=[0]),
 CNOT(wires=[0, 1]), measure(wires=[0]), measure(wires=[1]), measure(wires=[2])] == 
[Hadamard(wires=[0]), CNOT(wires=[0, 1]), measure(wires=[0]), Hadamard(wires=[0]), CNOT(wires=[0, 1]), 
 measure(wires=[0]), Conditional(wires=[0]), Conditional(wires=[0]), Conditional(wires=[2]), RZ(0.24, wires=[0]),
 CNOT(wires=[0, 1]), measure(wires=[0]), measure(wires=[1]), measure(wires=[2])]

E         At index 2 diff: measure(wires=[0]) != measure(wires=[0])
```

obliviateandsurrender: Added a manual check for testing it.

tests/test_converter.py
obliviateandsurrender: Added a manual check for testing it.

tests/test_converter.py
obliviateandsurrender: Added some rotation gates before it.

albi3ro (approved):
Thanks for adding in those test additions.  Looks good to me :)

PR #418, opened 2024-02-12
Param vector support

lillian542:
Allows `ParameterVector` to be converted to a PennyLane circuit, i. e.

```
theta = ParameterVector("v", 3)

qc = QuantumCircuit(2)
qc.rx(theta[0], [0])
qc.ry(theta[1], [0])
qc.rz(theta[2], [0])

@qml.qnode(qubit_device_2_wires)
def circuit(x, y, z):
    load(qc)(v=[x, y, z])
```

Currently does not support differentiability for `ParameterVector`. In a later PR we will either implement it, or add a warning that it is not supported (that triggers if trainable parameters are passed as values for a `ParameterVector`).


tests/test_converter.py
```diff
@@ -1095,87 +1156,6 @@ def circuit_native_pennylane(angle):
             rotation_angle
         )
 
-    def test_passing_parameters_new_interface_args(self, qubit_device_2_wires):
```
lillian542, line 1098: these are just copy-pasted further down to the section for testing passing parameters

lillian542:
[sc-56074]

pennylane_qiskit/converter.py
```diff
@@ -49,6 +49,36 @@ def _check_parameter_bound(param: Parameter, unbound_params: Dict[Parameter, Any
...
+
+    becomes [Parameter('a'), ParameterVector(name='v', length=3)].
+
+    Returns:
+        expected_params: The reorganized list of Parameters, containing Parameter and ParameterVector
+        param_name_string: a string listing the parameter names, i.e. in the example above, 'a, v'
+
+    """
+
+    # note we can't use `set` because it reorders parameters randomly
```
astralcai, line 67: The `dict` object maintains insertion order, you might want to try using that instead. The keys would be the parameter names and the values would be the parameters themselves.

astralcai (commented):
Looks good! I just have one small suggestion here.

astralcai (approved):
Looks good to me!

pennylane_qiskit/converter.py
```diff
@@ -164,20 +196,21 @@ def _check_circuit_and_assign_parameters(
                 f"quantum function, as this argument is reserved"
             )
 
+    expected_params, param_name_string = _expected_parameters(quantum_circuit)
```
obliviateandsurrender, line 199: Right now, are we assuming all the elements for each parameter vector to be present in the circuit? Cause doing the following throws an error for me - 
`CircuitError: 'Cannot bind parameters (t[1], v[0]) not present in the circuit.'`
Do you think it would be possible to adjust for such situations? Or you expect them to never happen?

image

lillian542: > Right now, are we assuming all the elements for each parameter vector to be present in the circuit?

Interesting point. Yes, currently we are. Unless I have misunderstood, the above would be a strange thing to do in Qiskit. You can't assign a full vector the `ParameterVector` in the Qiskit `assign_parameters` function, because you'll get the "not present in the circuit" error:

```
>>> new_qc = qc.assign_parameters({phi:2, rho:3, theta1: [1, 2], theta2: [2, 2]})
CircuitError: 'Cannot bind parameters (t[1], v[0]) not present in the circuit.'
```

Instead, you have to assign element-by-element, specifically avoiding any indices you didn't use, i.e.:

```
new_qc = qc.assign_parameters({phi:2, rho:3, theta1[0]: 1, theta2[1]: 2})
```

But once you are iterating through your `ParameterVector` and assigning the elements individually, why are you even using a `ParameterVector` instead of just a `Parameter`?

Do we want to accommodate this? It doesn't seem like it's really the intention even within Qiskit that it should be used that way.

obliviateandsurrender: Do we support `element-by-element` via our new parameter UI for a user who would want to do so? 🤔 I feel it would be nice if it won't be too much work. Otherwise, we can revisit it later.

lillian542: It's not included in the UI, and I think it would require some discussion to decide what it would look like. 

For a kwarg, for example, what would you use as an argument to denote the element? Their names are defined by qiskit as `vectorname[idx]`, and you can't include `[]` characters in a kwarg name.

Instead of going element-wise, we could allow the `ParamVectors` to be shorter than the full length and then pass them in as individual elements behind the scenes in order, but that differs from the Qiskit UI for defining them. 

I think we can add it as another "reach" story, but I think we would need to confirm what the UI should look like before we were ready to implement it.

lillian542: Another option would be to expect the full vector to be passed to PennyLane, and only use the elements Qiskit is expecting

obliviateandsurrender: I think we can safely skip this support until then.

obliviateandsurrender (approved):

PR #429, opened 2024-02-15
Add test for direct QNode UI

lillian542:


lillian542:
[sc-56077]

tests/test_converter.py
```diff
@@ -1416,3 +1416,30 @@ def test_diff_meas_circuit(self):
...
+
+        @qml.qnode(dev)
+        def circuit_native_pennylane(angle):
+            qml.RX(angle, wires=0)
+            qml.measure(0)
+            qml.RX(angle, wires=1)
+            qml.CZ([0, 1])
+            return qml.expval(qml.PauliZ(0)), qml.vn_entropy([1])
+
+        qnode = qml.QNode(load(qc, measurements), dev)
```
lillian542, line 1443: I would prefer to use `qml.from_qiskit` here, since that's the actual intended UI, but we run the tests with the released version of PL, so `qml.from_qiskit` can't take the measurements argument.

PR #431, opened 2024-02-15
Qiskit 1.0 compatibility for 0.35

lillian542:
Initial changes for 0.35 with 1.0. 

Iteroperability/conversion from Qiskit to PennyLane should work with 1.0. A few small changes are made here to accomodate that, and a workflow is added to test it.

Device functionality will not be supported until the following release, so using a device raises an error saying the Qiskit version is too high.

### DIFF
--- a/.github/workflows/tests_qiskit_1.yml
+++ b/.github/workflows/tests_qiskit_1.yml
@@ -1,0 +1,50 @@
+name: Tests
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        python-version: [3.9, '3.10', '3.11']
+
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.4.1
+        with:
+          access_token: ${{ github.token }}
+
+      - uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements-ci0.txt
+          pip install wheel pytest pytest-cov pytest-mock flaky --upgrade
+
+      - name: Install Plugin
+        run: |
+          python setup.py bdist_wheel
+          pip install dist/PennyLane*.whl
+
+      - name: Run Qiskit converter tests
+        # Test conversion to PennyLane with Qiskit 1.0.0
+        run: python -m pytest tests/test_converter.py
+        env:
+          IBMQX_TOKEN: ${{ secrets.IBMQX_TOKEN_TEST }}
+
+      - name: Run temporary tests
+        # tests that test intermediate functionality, will be removed when everything is fully compatible with 1.0
+        run: python -m pytest tests/test_new_qiskit_temp.py
+        env:
+          IBMQX_TOKEN: ${{ secrets.IBMQX_TOKEN_TEST }}
--- a/pennylane_qiskit/qiskit_device.py
+++ b/pennylane_qiskit/qiskit_device.py
@@ -22,6 +22,7 @@ import inspect
 import warnings
 
 import numpy as np
+import qiskit
 from qiskit import ClassicalRegister, QuantumCircuit, QuantumRegister
 from qiskit.circuit import library as lib
 from qiskit.compiler import transpile
@@ -141,6 +142,13 @@ class QiskitDevice(QubitDevice, abc.ABC):
     _eigs = {}
 
     def __init__(self, wires, provider, backend, shots=1024, **kwargs):
+
+        if qiskit.__version__ in ["0.46.0", "1.0.0"]:
+            raise RuntimeError(f"The devices in the PennyLane Qiskit plugin are currently only compatible "
+                               f"with version of Qiskit below 0.46. It looks like you have {qiskit.__version__} "
+                               f"installed. Please downgrade to use the devices. The devices will be updated "
+                               f"in the coming weeks to be compatible with Qiskit 1.0!")
+
         super().__init__(wires=wires, shots=shots)
 
         self.provider = provider
--- a/pennylane_qiskit/qiskit_device.py
+++ b/pennylane_qiskit/qiskit_device.py
@@ -23,7 +23,7 @@ import warnings
 
 import numpy as np
 from qiskit import ClassicalRegister, QuantumCircuit, QuantumRegister
-from qiskit import extensions as ex
+from qiskit.circuit import library as lib
 from qiskit.compiler import transpile
 from qiskit.converters import circuit_to_dag, dag_to_circuit
 from qiskit.providers import Backend, QiskitBackendNotFoundError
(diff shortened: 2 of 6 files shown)


lillian542:
[sc-56961]

.github/workflows/tests_qiskit_1.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Install Plugin
+        run: |
+          python setup.py bdist_wheel
+          pip install dist/PennyLane*.whl
+
+      - name: Run Qiskit converter tests
+        # Test conversion to PennyLane with Qiskit 1.0.0
+        run: python -m pytest tests/test_converter.py
+        env:
+          IBMQX_TOKEN: ${{ secrets.IBMQX_TOKEN_TEST }}
```
trbromley, line 44: Out of curiosity, do we need this environment variable?

pennylane_qiskit/qiskit_device.py
```diff
@@ -141,6 +142,15 @@ class QiskitDevice(QubitDevice, abc.ABC):
     _eigs = {}
 
     def __init__(self, wires, provider, backend, shots=1024, **kwargs):
+
+        if qiskit.__version__ in ["0.46.0", "1.0.0"]:
+            raise RuntimeError(
+                f"The devices in the PennyLane Qiskit plugin are currently only compatible "
+                f"with version of Qiskit below 0.46. It looks like you have {qiskit.__version__} "
```
trbromley, line 149:
```suggestion
                f"with versions of Qiskit below 0.46. You have version {qiskit.__version__} "
```

pennylane_qiskit/qiskit_device.py
```diff
@@ -141,6 +142,15 @@ class QiskitDevice(QubitDevice, abc.ABC):
...
+                f"installed. Please downgrade to use the devices. The devices will be updated "
+                f"in the coming weeks to be compatible with Qiskit 1.0!"
```
trbromley, line 151:
```suggestion
                f"installed. Please downgrade Qiskit to use the devices. The devices will be "
                f"updated in the coming weeks to be compatible with Qiskit 1.0!"
```

trbromley (approved):

.github/workflows/tests_qiskit_1.yml
timmysilv: needed for tests that use the circuit-runner or sampler devices (or any bespoke test device that inherits from `IBMQDevice` because its [connection](https://github.com/PennyLaneAI/pennylane-qiskit/blob/2344aa4a866e7f8614fc87df3004f645156838ee/pennylane_qiskit/ibmq.py#L77)) which I don't think applies to this test, so we can probably remove it from this step and the one below 👍

pennylane_qiskit/qiskit_device.py
```diff
@@ -22,8 +22,9 @@
 import warnings
 
 import numpy as np
+import qiskit
 from qiskit import ClassicalRegister, QuantumCircuit, QuantumRegister
-from qiskit import extensions as ex
+from qiskit.circuit import library as lib
```
timmysilv, line 27: to be sure, this existed in qiskit v0.x as well? do you know the minimum version at which it was available as `qiskit.circuit.library`?

pennylane_qiskit/qiskit_device.py
```diff
@@ -141,6 +142,15 @@ class QiskitDevice(QubitDevice, abc.ABC):
     _eigs = {}
 
     def __init__(self, wires, provider, backend, shots=1024, **kwargs):
+
+        if qiskit.__version__ in ["0.46.0", "1.0.0"]:
```
timmysilv, line 146: so we're allowing installation with the new qiskit version only for conversion purposes because they won't be able to make a device right?

also, if the goal here is "must be less than 0.46.0", you might want something more general, like this:
```pycon
>>> from semantic_version import Version
>>> max_ver = Version("0.46.0")
# here you'd pass `qiskit.__version__` into `Version(...)` and compare
>>> Version("0.45.9") < max_ver, Version("1.1.0") < max_ver
(True, False)
```

pennylane_qiskit/qiskit_device.py
obliviateandsurrender: Yep. Has been there since [v0.24](https://docs.quantum.ibm.com/api/qiskit/0.24/circuit_library) at least.

pennylane_qiskit/qiskit_device.py
obliviateandsurrender: +1. We should ensure only the version less than `0.46/1.0` should be present. 
As a side note - Qiskit won't allow installing `1.0` along with `0.46`, in case that would help with version checking.

requirements-ci0.txt
```diff
@@ -0,0 +1,4 @@
+pennylane>=0.32
+qiskit
```
obliviateandsurrender, line 2: We don't have to pin it here?

lillian542: This is the CI file for the tests that we *want* to run with the latest qiskit, so not here

.github/workflows/tests_qiskit_1.yml
lillian542: Actually I think we should add an IBMQ device to the test instead, since I'm realizing different devices have different import errors that might be raised before the more informative error message we want to come up

pennylane_qiskit/qiskit_device.py
lillian542: done!

timmysilv (approved):

setup.py
```diff
@@ -22,7 +22,7 @@
     long_description = fh.read()
 
 requirements = [
-    "qiskit>=0.32,<0.46",
+    "qiskit>=0.32",
```
obliviateandsurrender, line 25: We no longer need an upper bound here? 🤔

lillian542: The issue was that we don't want to break people's environment if they install 1.0, and then install the plugin with the intention of using it for conversion. We want Qiskit 1.0 plus the PL-plugin for conversion to be an acceptable combination.

But this does mean if people pip install without an older version of Qiskit installed, then the devices won't work. Is there a more complicated version of a setup specification that allows us to say "install version 0.45.3 if they don't have Qiskit, but if they already have it and its above 0.32, leave their installation"? I.e. to have the default installed dependency differ from the highest allowed version?

Mandrenkov: It's worth noting that `setup.py` only defines which versions of packages (e.g., Qiskit) are marked as "compatible" with the current package (e.g, the PL plugin). Consequently,
* If a user attempts to `pip install` the PL plugin and doesn't have Qiskit installed, `pip` will automatically search for a version of Qiskit which satisfies all of the constraints of the user's packages and install it.
* If a user attempts to `pip install` the PL plugin and already has a compatible version of Qiskit installed, nothing changes about their Qiskit installation.
* If a user attempts to `pip install` the PL plugin and has a version of Qiskit installed which is _not_ compatible, `pip` will attempt to reinstall a compatible version of Qiskit (as defined by the range in the PL plugin).

This is to say that we should only put an upper bound here if this plugin will not work with Qiskit 1.0 or later.

lillian542: After speaking to @trbromley offline, the decision was to remove the upper bound (we want people to be able to use Qiskit 1.0 and the conversion functionality), and update the install instructions so that if you just follow the instructions, you get `qiskit<0.46` and everything works. This means some people might have issues if they just pip install the plugin without looking at the documentation and they don't have a previous version of Qiskit and they want to use the devices. But something has to not work.

README.rst
```diff
@@ -58,8 +58,13 @@ Installation of this plugin, as well as all dependencies, can be done using ``pi
 
 .. code-block:: bash
 
+    pip install "qiskit<0.46"
     pip install pennylane-qiskit
 
+If you prefer to use Qiskit 1.0, you can replace the first line above with ``pip install qiskit``,
+or omit it entirely. Note that only conversion to PennyLane is supported for Qiskit 1.0; support
```
trbromley, line 65:
```suggestion
If you prefer to use Qiskit 1.0, you can omit the first line above and instead consult the
`Qiskit installation guide <https://docs.quantum.ibm.com/start/install>`__, which includes
details on how to migrate to 1.0 if you already have Qiskit installed.
Note that only conversion to PennyLane is supported with Qiskit 1.0; support
```

pennylane_qiskit/qiskit_device.py
Josh-H-555: Hey there! By passing this error message and using Qiskit 1.0.1, I'm able to create a device perfectly fine. Is there something else I'm missing as to why the Qiskit device functionality is technically unsupported in 0.35?

lillian542: Hi @Josh-H-555! Thanks for asking, this is a good point. Qiskit 0.46/1.0 came out a bit close to our scheduled release. We didn't have enough time check everything we would need to in order to guarantee everything was working. We did some of the groundwork for compatibility, so some areas of the code base surrounding the devices will work as expected, but others won't. 

Rather than recommend something we weren't confident in, we opted to impose a version requirement until we can go through things more thoroughly and fix the remaining incompatibilities. We're planning to have a fix for more comprehensive 1.0 compatibility in the coming weeks!

Josh-H-555: I see, that makes sense. Thanks for the quick reply! I look forward to the upcoming updates. :)

PR #432, opened 2024-02-15
Add support for `expr` expressions and related warning

obliviateandsurrender:
1. Adds support for evaluating `expr` expression for classical control flows. 
2. Adds a warning if `expr` being performed on classical registers cannot be evaluated, i.e., classical registers would store information that we cannot process and return.


obliviateandsurrender:
[sc-56410]

pennylane_qiskit/converter.py
```diff
@@ -561,36 +543,242 @@ def load_qasm_from_file(file: str):
...
+        if isinstance(carg, expr.Value):
+            clvals.append([carg.value])
+        elif isinstance(carg, expr.Var):
+            clbits.append([carg.var] if isinstance(carg.var, Clbit) else list(carg.var))
+
+    # Proceed only if we have access to all conditioned classical bits
+    for idx, clreg in enumerate(clbits):
+        if not all(clbit in mid_circ_regs for clbit in clreg):
+            return None
+        clbits[idx] = [mid_circ_regs[clbit] for clbit in clreg]
```
timmysilv, line 714: are we supposed to be overriding values here?

timmysilv (commented):
I'll be honest.. I can't really tell what's going on here : ) a lot of assumed knowledge about Qiskit classical stuff that I certainly don't have, so maybe I shouldn't give a final review. just left a few things I saw.

timmysilv (approved):

pennylane_qiskit/converter.py
```diff
@@ -561,36 +543,242 @@ def load_qasm_from_file(file: str):
...
+        true_fns, res_bits = [], []
+        for case, res_bit in inst._case_map.items():
             if not isinstance(case, _DefaultCaseType):
-                elif_fns.append((res_bit, branch_funcs[case]))
-        ops.condition = [tuple(cargs), "SwitchCase"]
-        if any((isinstance(case, _DefaultCaseType) for case in ops._case_map)):
-            true_fn = branch_funcs[-1]
-            ops.condition = [tuple(cargs), "SwitchDefault"]
+                true_fns.append(branch_funcs[res_bit])
+                res_bits.append(case)
```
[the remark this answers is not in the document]
obliviateandsurrender: I will rename the variables to make it clearer, but the `ops._case_map` is a dictionary with keys as the switch's `cases` and the values as the corresponding `branch` index.  That's why I use `res_bit` for indexing he branch and add the case to `res_bits` list.

pennylane_qiskit/converter.py
obliviateandsurrender: Yes, this is intentional. What it does is map Qiskit's classical bits for each register to the corresponding mid-circuit PL measurement that we keep track of in the `mid_circ_regs`.

lillian542 (approved):
May they never deprecate or expand on any of this functionality in a way that requires us to modify this code 🤞 

Looks like it works to me 🎉 🚀

PR #453, opened 2024-02-22
Add PennyLane IO entry point for loading Qiskit operators

Mandrenkov:
This PR adds a `pennylane.io` entrypoint for the `load_pauli_op()` function introduced in #401.

For usage details, see PennyLaneAI/pennylane#5251.


Mandrenkov:
As a heads up @lillian542, I'll wait until https://github.com/PennyLaneAI/pennylane/pull/5251 is approved and then port over any documentation changes to this PR before merging.

Mandrenkov:
@lillian542 I've finished porting the documentation changes and also added a missing check for the type of `pauli_op`. We do a similar check in `load()` but it was hidden in a private function and so I missed it before.

lillian542 (approved):

PR #459, opened 2024-02-26
Gracefully handle batch execution of no circuits

Mandrenkov:
Recently, a [PR to PennyLane](https://github.com/PennyLaneAI/pennylane/pull/5243) was merged which changed the default caching behaviour and caused 

```
tests/test_integration.py::TestPLTemplates::test_tensor_unwrapped_gradient_no_error
```

to fail with

```
TypeError: bad input to run() function;circuits must be either circuits or schedules
```

The error stems from `QiskitDevice.batch_execute()` being called with an empty list of circuits.

### DIFF
--- a/pennylane_qiskit/qiskit_device.py
+++ b/pennylane_qiskit/qiskit_device.py
@@ -496,6 +496,10 @@ class QiskitDevice(QubitDevice, abc.ABC):
 
         compiled_circuits = self.compile_circuits(circuits)
 
+        if not compiled_circuits:
+            # At least one circuit must always be provided to the backend.
+            return []
+
         # Send the batch of circuit objects using backend.run
         self._current_job = self.backend.run(compiled_circuits, shots=self.shots, **self.run_args)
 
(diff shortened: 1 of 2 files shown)


lillian542 (approved):
Seems reasonable. Does the test then still test what it was supposed to, or does the test also need updating?

obliviateandsurrender (approved):

Mandrenkov:
Thanks for the reviews, @lillian542 and @obliviateandsurrender!

> Does the test then still test what it was supposed to, or does the test also need updating?

Are you referring to `test_tensor_unwrapped_gradient_no_error()`? If so, I think it's still correct. It's just that caching is now (mostly) disabled by default in PennyLane and we were accidentally relying on caching in this test previously.

Mandrenkov:
> Should this PR fix https://github.com/PennyLaneAI/plugin-test-matrix/actions/runs/8073162814/job/22056323586?

@trbromley Indeed it should! 🎉

PR #467, opened 2024-02-28
Add condition for treating all measurements as equal

obliviateandsurrender:
Allows for treating all measurements in the circuit as equal and returning them in order to the user when `measurements` are not specified.


Mandrenkov (approved):
Looks good to me! 🥳

PR #492, opened 2024-03-08
Change qml.load to qml.from_qiskit

astralcai:
`qml.load` is being deprecated in PennyLane. Updating the test case to use `qml.from_qiskit` instead.


lillian542 (approved):
Thanks @astralcai!

PR #501, opened 2024-03-12
Remove use of from_qasm_file in test case

astralcai:
`from_qasm_file` is deprecated. Removes the use of `from_qasm_file` from test cases. Instead, read a file manually and use `from_qasm`.


lillian542 (approved):
lgtm 🎉

PR #514, opened 2024-04-19
BackendV2 support for old device API

austingmhuang:
This pull request aims to provide support for V2 devices using the old device API. This addresses an [issue](https://github.com/PennyLaneAI/pennylane-qiskit/issues/332) from a user requesting this feature. There is a similar PR open for the new device API as well, but that might not be ready with the 0.36 release.


tests/test_qiskit_device.py
```diff
@@ -13,6 +73,23 @@
 ]
 
 test_device_options = [{}, {"optimization_level": 3}, {"optimization_level": 1}]
+backend = MockedBackend()
+legacy_backend = MockedBackendLegacy()
```
lillian542, line 77: So I know this is a bit of a 180 from what I said previously wrt the mocking, but what does stop us from just using the `FakeManila` and `FakeManilaV1` backends? Do they require a token? I.e. just:

```
from qiskit_ibm_runtime.fake_provider import FakeManilaV2, FakeManila
```

and then we replace this with 

```
backend = FakeManillaV2()
legacy_backend = FakeManila()
```

lillian542 (commented):
Since this change adds support for new functionality, add it (and your name) to the changelog 🚀

tests/test_qiskit_device.py
austingmhuang: As far as I know, they don't require a token. I think we could just use `FakeManila` for the tests as well. I'll add it as an additional test case.

pennylane_qiskit/qiskit_device.py
```diff
@@ -181,8 +181,11 @@ def __init__(self, wires, provider, backend, shots=1024, **kwargs):
         self._capabilities["returns_state"] = self._is_state_backend
 
         # Perform validation against backend
-        backend_qubits = self.backend.configuration().n_qubits
-        # if the backend has a set number of qubits, ensure wires doesn't exceed (some simulators have n_qubits=None)
+        backend_qubits = (
```
obliviateandsurrender, line 184: This change seems to suggest that we are adding support for the `BackendV2`, which is quite the opposite of what you mention in the `changelog`. Could you please confirm this once? 🤔

tests/test_qiskit_device.py
```diff
@@ -13,6 +74,43 @@
...
+    @pytest.mark.parametrize(
+        "backend",
+        [
+            legacy_backend,
+            backend,
+        ],
+    )
+    def test_v1_and_v2_mocked(self, backend):
+        """Test that device initializes with no error mocked"""
+        dev = qml.device("qiskit.remote", wires=10, backend=backend, use_primitives=True)
```
obliviateandsurrender, line 93: We should add `assertions` to test that the device indeed gets built with the correct `backend` here.

tests/test_qiskit_device.py
```diff
@@ -13,6 +74,43 @@
...
+        """Test that device initializes with no error with V1 and V2 backends by Qiskit"""
+        dev = qml.device("qiskit.remote", wires=5, backend=backend, use_primitives=True)
+        
+        @qml.qnode(dev)
+        def circuit(x):
+            qml.RX(x, wires=[0])
+            qml.CNOT(wires=[0, 1])
+            return qml.expval(qml.PauliZ(0))
+        
+        circuit(np.pi/2)
```
obliviateandsurrender, line 113: Optional: we may add an `assertion` to test the circuit results. Could be skipped if the outputs are noisy.

obliviateandsurrender (commented):
Thanks @austingmhuang, this looks great! Just leaving a couple of suggestions

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@
   device, ``BasicSimulator``, as the backend.
   [(#493)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/493)
 
+* Backwards compatibility with Qiskit BackendV1 has now been implemented. Previously, only backends of type BackendV2 were supported but now users can choose to use BackendV1 as well.
```
lillian542, line 11: I believe it's the other way around - we previously only supported V1 backends. Adding the `backend.num_qubits` syntax supports V2.

tests/test_qiskit_device.py
austingmhuang: Outputs are noisy here since we are simulating a real backend. I talked about this with @lillian542 but didn't think of a good way to test the circuit results (as far as I know Qiskit doesn't provide a BackendV1 that is noise-free).

tests/test_qiskit_device.py
austingmhuang: That's a good point. Will do that.

pennylane_qiskit/qiskit_device.py
austingmhuang: Ah you're right. Sorry let me fix that. We are adding support for BackendV2.

(The context is that most Qiskit devices that are BackendV2 are actually backwards compatible with BackendV1. So on the surface it seems like the current code supports BackendV2, but actually it simply supports BackendV2s that are backwards compatible -- which is the majority of the devices at Qiskit).

soranjh: The title also says "BackendV1 support for old device API".

CHANGELOG.md
austingmhuang: Yep. Utkarsh mentioned the same thing as well. Thanks Lillian!

lillian542 (approved):
Thanks @austingmhuang! 

Let's remove the `.DS_Store` file, that's just a thing that happens locally on Macs, we shouldn't need it on GitHub. But otherwise looks good to go to me! 🎉

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
   device, ``BasicSimulator``, as the backend.
   [(#493)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/493)
 
+* Backwards compatibility with Qiskit BackendV2 has now been implemented. Previously, only backends of type
+  BackendV1 were supported but now users can choose to use BackendV2 as well.
+  [(#493)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/514)
```
soranjh, line 13:
```suggestion
  [(#514)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/514)
```

CHANGELOG.md
```diff
@@ -30,6 +34,7 @@
 
 This release contains contributions from (in alphabetical order):
 Lillian M. A. Frederiksen
+Austin Huang
```
soranjh, line 37: Should be in alphabetical order.

obliviateandsurrender (approved):
Thanks @austingmhuang!

CHANGELOG.md
austingmhuang: It is in alphabetical order.... unless I'm missing something?

soranjh: The middle names are a bit tricky!

PR #517, opened 2024-04-25
Generalize mp_to_pauli so that it can work with any observable that has a Pauli Rep

austingmhuang:
Right now mp_to_pauli only works for observables that act only on a single qubit. To improve upon this, we can actually make it so that any observable can be converted to its respective SparsePauliOp. This includes Hamiltonians, LinearCombinations, SProd, Prod, Sum, etc...

This PR also includes some minor refactoring of the converter.py file in general (using f strings, deleting unneeded variables, etc..."


tests/test_converter.py
```diff
@@ -1745,14 +1773,27 @@ def test_with_predefined_creg(self):
...
+            (qml.PauliZ, "Z"),
+            (qml.Identity, "I"),
+            (qml.X, "X"),
+            (qml.Y, "Y"),
+            (qml.Z, "Z"),
+            (qml.I, "I"),
+            (X, "X"),
+            (Y, "Y"),
+            (Z, "Z"),
+            (I, "I"),
```
obliviateandsurrender, line 1791: Aren't we repeating these, i.e., `qml.X` being equal to `X`?

tests/test_converter.py
```diff
@@ -1775,10 +1816,137 @@ def test_mp_to_pauli(self, measurement_type, observable, obs_string, wire, regis
         # remaining wires are all Identity
         assert np.all([op == "I" for op in pauli_op_list])
 
+    @pytest.mark.parametrize("measurement_type", [qml.expval, qml.var])
+    @pytest.mark.parametrize("register_size", [3, 5, 7])
+    @pytest.mark.parametrize(
+        "observables, coeffs",
```
obliviateandsurrender, line 1822: We should test it for multi-qubit observables as well.

tests/test_converter.py
```diff
@@ -1775,10 +1816,137 @@ def test_mp_to_pauli(self, measurement_type, observable, obs_string, wire, regis
...
+        coeffs = []
+        for op in operator:
+            for pauli_term, coeff in op.pauli_rep.items():
+                pauli_string = ["I"] * register_size
+                coeffs.append(coeff)
+                for _, val in enumerate(pauli_term.wires):
+                    pauli_string[val] = pauli_term[val]
+                pauli_string.reverse()
+                pauli_string = ("").join(pauli_string)
+                data.append(pauli_string)
```
obliviateandsurrender, line 1887: This code is pretty much a duplication of the code used for conversion to `SparsePauliOp` above. Doing this makes us prone to weak tests. A better method would be to compute the `SparsePauliOp` for the test cases manually and put them in `@pytest.mark.parametrize` in the way you do below.

obliviateandsurrender (changes requested):
Thanks @austingmhuang! Leaving a few initial comments.

pennylane_qiskit/converter.py
```diff
@@ -541,7 +543,11 @@ def _function(*args, params: dict = None, wires: list = None, **kwargs):
                 # Iteratively recurse over to build different branches for the condition
                 with qml.QueuingManager.stop_recording():
                     branch_funcs = [
-                        partial(load(branch_inst, measurements=None), params=params, wires=wires)
+                        partial(
```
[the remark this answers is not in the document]
austingmhuang: @soranjh and I were trying to figure this out actually... but basically, here's what I did. I put a linter in my local file and ran a `pre-commit` hook the same way you do for the pennylane repo and it reformatted these parts. If I manually change it back to the original formatting and run `black -l 100 tests` on `generalize_ops` it will reformat it back to this. But when I run black on `new_device_feature_branch` nothing gets changed so... :shrug:

My solution is to just ignore it for now, it does make the PR hard to read though, sorry.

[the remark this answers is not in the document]
obliviateandsurrender: It's a line I wrote to handle controlled flow branches, which themselves include circuits. So `black -l 100 tests` shouldn't have reformated it. It's fine like this, just wanted to figure out if you were modifying something here apart from formatting.

pennylane_qiskit/converter.py
```diff
@@ -706,20 +713,20 @@ def mp_to_pauli(mp, register_size):
...
-
-    return SparsePauliOp(pauli_string)
+    pauli_strings = []
+    coeffs = []
+    op = mp.obs
+
+    for pauli_term, coeff in op.pauli_rep.items():
+        pauli_string = ["I"] * register_size
+        coeffs.append(coeff)
+        for _, val in enumerate(pauli_term.wires):
```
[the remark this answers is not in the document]
austingmhuang: Yeah a list comprehension is a good idea. Custom labels are not supported because we don't know the order of the wires.

e.g. 

op = (qml.X(1) + qml.Y(1)) @ (qml.Z(3) + qml.Z("ancilla"))
register_size = 4

The ancilla qubit could be index = 0 or index = 2 but we can't know for sure from the information passed to the mp_to_pauli function. We could fix this by adding a `wire_order` kwarg and then translating `wire_order` to some `wire_map`, but I think that would be better as a separate function.

In general, this conversion only works assuming that 1. the wire_order is ascending and 2. the wires are all ints.

pennylane_qiskit/converter.py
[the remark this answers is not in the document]
austingmhuang: Yeah it shouldn't have reformatted it which is why Soran and I were confused. Not sure what exactly changed 😨

tests/test_converter.py
```diff
@@ -777,7 +789,7 @@ def test_parameter_is_not_unbound(self):
 class TestConverterUtils:
     """Tests the utility functions used by the converter function."""
 
-    def test_map_wires(self, recorder):
```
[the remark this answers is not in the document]
austingmhuang: It was an unused variable.

tests/test_converter.py
austingmhuang: They're all equal to each other but I figured I'd add something to confirm that all 3 representations work properly. If this is unnecessary or too much clutter I can delete as well.

tests/test_converter.py
austingmhuang: Okay 👍

tests/test_converter.py
austingmhuang: Actually that was the original test back when the test only allowed for single qubit observables. This type of test doesn't work on multi-qubit observables. Probably better to just compare to manual test cases every time.

tests/test_converter.py
austingmhuang: I removed some of these tests but the coverage should still be the same.

tests/test_converter.py
```diff
@@ -1745,40 +1773,185 @@ def test_with_predefined_creg(self):
...
+                0.5 * X(0) + 0.5 * qml.Y(0) - 1.5 * qml.X(0) - 0.5 * qml.Y(0),
+                0.5 * SparsePauliOp("IIIIX")
+                + 0.5 * SparsePauliOp("IIIIY")
+                - 1.5 * SparsePauliOp("IIIIX")
+                - 0.5 * SparsePauliOp("IIIIY"),
+            ),
+            (
+                qml.ops.LinearCombination(
+                    [1, 3, 4],
+                    qml.Hamiltonian([1, 3], qml.dot([2, 3], [qml.X(0), qml.Y(0)])) - 2 * qml.Y(2),
```
lillian542, line 1826: The second argument of `LinearCombination` should be a list with a length that matches `len(coeffs)`, not a single operator. This works because the operator happens to be a Sum with 3 terms, but it doesn't match the docstring or type hints for how this operator should be created. I would suggest updating the test to follow the standard convention for the function.

lillian542 (commented):
Looking good @austingmhuang, thanks for working on this! I think there are a few small loose ends to wrap up before this is ready to merge. 

I left a couple of comments in the code, and I would also like to suggest that a few more "integration" style tests are added, probably in `test_base_device` tests, using this functionality within an execution. For example maybe adding additional parameterizations in `test_process_estimator_job`, and adding a test similar to `test_estimator_with_different_pauli_obs` that checks more complicated observables (you can compare to the results of default.qubit instead of to analytical results).

pennylane_qiskit/converter.py
```diff
@@ -706,20 +713,22 @@ def mp_to_pauli(mp, register_size):
...
+            "".join(
+                ["I" if i not in pauli_term.wires else pauli_term[i] for i in range(register_size)][
+                    ::-1
+                ]  ## Qiskit follows opposite wire order convention
+            )
+            for pauli_term in op.pauli_rep.keys()
+        ]
+        coeffs = list(op.pauli_rep.values())
+    else:
+        raise ValueError(f"The operator")
```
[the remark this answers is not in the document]
lillian542: Ah, never mind, I see you changed it 😅

austingmhuang:
> Looking good @austingmhuang, thanks for working on this! I think there are a few small loose ends to wrap up before this is ready to merge.
> 
> I left a couple of comments in the code, and I would also like to suggest that a few more "integration" style tests are added, probably in `test_base_device` tests, using this functionality within an execution. For example maybe adding additional parameterizations in `test_process_estimator_job`, and adding a test similar to `test_estimator_with_different_pauli_obs` that checks more complicated observables (you can compare to the results of default.qubit instead of to analytical results).

Yeah that makes sense. I'll have to double check and see what observables default_qubit supports and edit qiskitdevice2 to support the same things before I write these "integration" style tests. This is a good point to bring up though, thanks @lillian542!

tests/test_base_device.py
```diff
@@ -416,7 +416,7 @@ def test_stopping_conditions(self, op, expected):
         [
             (qml.PauliX(0), True),
             (qml.Hadamard(3), True),
-            (qml.prod(qml.PauliY(1), qml.PauliZ(0)), False),
+            (qml.prod(qml.PauliY(1), qml.PauliZ(0)), True),
```
austingmhuang, line 419: prod (and others) now supported!

pennylane_qiskit/qiskit_device2.py
```diff
@@ -239,10 +239,18 @@ class QiskitDevice2(Device):
...
+        "Identity",
         "Projector",
+        "SparseHamiltonian",
+        "Hamiltonian",
+        "LinearCombination",
+        "Sum",
+        "SProd",
+        "Prod",
+        "Exp",  # this may not work
+        "Evolution",  # this may not work
```
lillian542, line 253: What is the motivation for putting these here? `Exp` isn't really an observable, and I'm a little hesitant to add `Evolution` to the list of supported observables, especially without tests

austingmhuang: Ah, no real motivation here, I just copy and pasted the supported observables from the default qubit and wrote these two notes to myself and then forgot about it 😓 . You're right though, they're not observables they're operators. And as operators they are not supported either so I will delete this.

austingmhuang: I also removed SparseHamiltonian for now, as it is only supported for Expval and not Var and Samples.

lillian542: They should both, when they are based on appropriate inputs, be decomposed to something that the device supports - they are listed on default.qubit because as a simulator, it can support them natively.

I'm actually a little hesitant to depart from the observables included in the current QiskitDevice, i.e. 

```
    observables = {
        "PauliX",
        "PauliY",
        "PauliZ",
        "Identity",
        "Hadamard",
        "Hermitian",
        "Projector",
    }
```
at least without testing them using both Estimator, Sampler and the old `run` function on a hardware device. 

Things like `Hamiltonian` currently work on the old version of the Qiskit device because they are broken down by `hamiltonian_expand`. It should be compatible with Estimator specifically, but what about Sampler? 

Also out of curiosity, does your test involving `Prod` fail if you remove `Prod` from this list?

lillian542: > Also out of curiosity, does your test involving Prod fail if you remove Prod from this list?

to clarify, not the stopping condition test, but the integration tests

austingmhuang: > to clarify, not the stopping condition test, but the integration tests

So removing the stopping condition and removing Prod from the list, the integration tests pass without issue.

austingmhuang: > They should both, when they are based on appropriate inputs, be decomposed to something that the device supports - they are listed on default.qubit because as a simulator, it can support them natively.
> 
> I'm actually a little hesitant to depart from the observables included in the current QiskitDevice, i.e.
> 
> ```
>     observables = {
>         "PauliX",
>         "PauliY",
>         "PauliZ",
>         "Identity",
>         "Hadamard",
>         "Hermitian",
>         "Projector",
>     }
> ```
> 
> at least without testing them using both Estimator, Sampler and the old `run` function on a hardware device.
> 
> Things like `Hamiltonian` currently work on the old version of the Qiskit device because they are broken down by `hamiltonian_expand`. I think it should be compatible with Estimator specifically without breaking it down like that, but what about Sampler?

In this case, we could restrict things like 'Hamiltonian' to only work with Expval, the same way 'SparseHamiltonian' is restricted on the defaultqubit device. Alternatively I can add tests for the Sampler and the old run function as well, although that to me is out of scope for this generalization of mp_to_pauli. That to me would be a separate PR i.e. "Ensure/make all observables compatible with all measurement processes". What do you think?

austingmhuang: > They should both, when they are based on appropriate inputs, be decomposed to something that the device supports - they are listed on default.qubit because as a simulator, it can support them natively.
> 

Can you clarify further what you mean by this? Do you mean to say that there _is_ a pauli_rep for these observables that I can use to convert into an SPO? Or just that theoretically there should exist a pauli_rep that is just not implemented right now?

lillian542: Don't worry about the `pauli_rep` in the context of the set of allowed `observables`. The device can support lots of observables on the user end that aren't listed here. This is a list of what is supported on the device at the time the device runs `execute`. Most things have `decomposition` methods defined, and can be decomposed into things that the device supports at execution, but some things can't. Though now that I'm giving it more thought, I'm not sure we can decompose `Exp` specifically to be a valid observable for this device. But we can run many things that aren't on the `observables` list.

Wrt. how to deal with Hamiltonian, it shouldn't be necessary to restrict what type of measurement it works with once the transform is added to preprocessing. That is planned to happen in [this story](https://app.shortcut.com/xanaduai/story/54392/transforms-are-included-in-preprocessing-for-observables-that-require-hamiltonian-expand-sum-expand-and-grouping, doesn't involve adding `Hamiltonian` to supported observables.

This might be easier to discuss in person, should we try to chat sometime tomorrow?

austingmhuang: Yes, let's discuss in person instead. That would probably be easier 😅 .

tests/test_converter.py
```diff
@@ -1946,6 +2096,7 @@ def built_pl_circuit():
             )
         )
 
+    # pylint:disable=not-context-manager
```
obliviateandsurrender, line 2099: I think you can place this just once at the top of the class, instead of covering every method with it.

tests/test_base_device.py
```diff
@@ -1057,10 +1115,22 @@ def test_estimator_with_different_pauli_obs(self, mocker, wire, angle, op, expec
...
+                        + qml.X(4)
+                    )
+                ],
+                (16),
+            ),
         ],
     )
     def test_process_estimator_job(self, measurements, expectation):
-        """for variance and for expval and for a combination"""
+        """Tests that the estimator returns expected and accurate results for a Expval and Var for a variety of multi qubit observables"""
```
obliviateandsurrender, line 1133:
```suggestion
        """Tests that the estimator returns expected and accurate results for an ``expval`` and ``var`` for a variety of multi-qubit observables"""
```

tests/test_base_device.py
```diff
@@ -1086,10 +1156,9 @@ def test_process_estimator_job(self, measurements, expectation):
         for data in result.metadata:
             assert isinstance(data, dict)
             assert list(data.keys()) == ["variance", "shots"]
-
         processed_result = QiskitDevice2._process_estimator_job(qs.measurements, result)
         assert isinstance(processed_result, tuple)
-        assert np.allclose(processed_result, expectation, atol=0.05)
+        assert np.allclose(processed_result, expectation, atol=0.1)
```
obliviateandsurrender, line 1161: Is this change required for the new observables?

tests/test_converter.py
```diff
@@ -1627,6 +1626,7 @@ def test_circuit_to_qiskit_register_size(self, operations, register_size):
         [([qml.PauliX(0), qml.PauliY(1)], "y"), ([[qml.CNOT([0, 1]), qml.Hadamard(1)], "h"])],
     )
     @pytest.mark.parametrize("measure", [True, False])
+    # pylint:disable=pointless-statement
```
obliviateandsurrender, line 1629: Why is this used? Maybe we could improve this test to skip this? 🤔

austingmhuang: Improved 😎

tests/test_qiskit_device.py
```diff
@@ -25,11 +25,6 @@
 import pennylane as qml
 from pennylane_qiskit import AerDevice
 from pennylane_qiskit.qiskit_device import QiskitDevice
-from qiskit_aer import noise
-from qiskit.providers import BackendV1, BackendV2
-from qiskit_ibm_runtime.fake_provider import FakeManila, FakeManilaV2
-from unittest.mock import Mock
-from qiskit_ibm_runtime.options import Options
```
austingmhuang, line 32: Linter errors here.

tests/test_base_device.py
```diff
@@ -39,23 +48,11 @@
...
-from qiskit_ibm_runtime.fake_provider import FakeManila, FakeManilaV2
-
-# do not import Estimator (imported above) from qiskit.primitives - the identically
-# named Estimator object has a different call signature than the remote device Estimator,
-# and only runs local simulations. We need the Estimator from qiskit_ibm_runtime. They
-# both use this EstimatorResults, however:
-from qiskit.primitives import EstimatorResult
-from qiskit.providers import BackendV1, BackendV2
-
-from qiskit import QuantumCircuit
```
austingmhuang, line 54: Moving/deleting these imports for linter.

austingmhuang:
[sc-62045]

obliviateandsurrender (approved):
Looks good to me, thanks @austingmhuang!

PR #521, opened 2024-04-30
Move off of "ibmq_qasm_simulator" and towards using AerSimulator for qiskit tests

austingmhuang:
As of right now, many of our tests are still using the ibmq_qasm_simulator cloud-service backend that is going to be deprecated in 2 weeks on May 15th. In response to this we are migrating from "ibmq_qasm_simulator" to a local simulator, AerSimulator, which is what Qiskit recommends for the migration.

### DIFF
--- a/pennylane_qiskit/qiskit_device2.py
+++ b/pennylane_qiskit/qiskit_device2.py
@@ -288,7 +288,6 @@ class QiskitDevice2(Device):
         self._backend = backend
         self._compile_backend = compile_backend if compile_backend else backend
 
-        # ToDo: possibly things fail if this is not a QiskitRuntimeService - confirm and decide how to handle (SC 55725)
         self._service = getattr(backend, "_service", None)
         self._use_primitives = use_primitives
         self._session = session
--- a/pennylane_qiskit/qiskit_device2.py
+++ b/pennylane_qiskit/qiskit_device2.py
@@ -550,11 +550,19 @@ class QiskitDevice2(Device):
             else self.backend.configuration().backend_name
         )
 
-        options = {
+        circuit_runner_options = {
             "backend": backend_name,
             "log_level": self.options.environment.log_level,
             "job_tags": self.options.environment.job_tags,
             "max_execution_time": self.options.max_execution_time,
+        }
+
+        # Does not support AerSimulator specific options e.g. choose a specific method
+        # refer to this https://qiskit.github.io/qiskit-aer/stubs/qiskit_aer.AerSimulator.html
+        # To support this would be confusing in terms of option setting (The options we currently support are runtime options)
+        # This here is just to capture and track shots information
+        aer_options = {
+            "shots": circuits[0].shots.total_shots or self.shots.total_shots,
         }
 
         # Send circuits to the cloud for execution by the circuit-runner program.
@@ -562,14 +570,13 @@ class QiskitDevice2(Device):
         if self.service:
             job = self.service.run(
                 program_id="circuit-runner",
-                options=options,
+                options=circuit_runner_options,
                 inputs=program_inputs,
                 session_id=session.session_id,
             )
             self._current_job = job.result(decoder=RunnerResult)
         else:  # Uses local simulator instead. After May 15th, all simulations will use this logic instead.
-            self.options.execution.shots = program_inputs["shots"]
-            self.backend.set_options(shots=self.options.execution.shots)
+            self.backend.set_options(**aer_options)
             job = self.backend.run(
                 compiled_circuits,
             )
--- a/pennylane_qiskit/qiskit_device2.py
+++ b/pennylane_qiskit/qiskit_device2.py
@@ -279,7 +279,7 @@ class QiskitDevice2(Device):
         self._compile_backend = compile_backend if compile_backend else backend
 
         # ToDo: possibly things fail if this is not a QiskitRuntimeService - confirm and decide how to handle (SC 55725)
-        self._service = backend._service
+        self._service = getattr(backend, "_service", None)
         self._use_primitives = use_primitives
         self._session = session
 
--- a/pennylane_qiskit/qiskit_device2.py
+++ b/pennylane_qiskit/qiskit_device2.py
@@ -546,13 +546,18 @@ class QiskitDevice2(Device):
         }
 
         # Send circuits to the cloud for execution by the circuit-runner program.
-        job = self.service.run(
-            program_id="circuit-runner",
-            options=options,
-            inputs=program_inputs,
-            session_id=session.session_id,
-        )
-        self._current_job = job.result(decoder=RunnerResult)
+        # Cloud simulators will be deprecated on May 15th so this will be exclusively for real hardware devices.
+        if self.service:
+            job = self.service.run(
+                program_id="circuit-runner",
+                options=options,
+                inputs=program_inputs,
+                session_id=session.session_id,
+            )
+            self._current_job = job.result(decoder=RunnerResult)
+        else:  # Uses local simulator instead. After May 15th, all simulations will use this logic instead.
+            job = self.backend.run(compiled_circuits, options=options)  # Missing shots
+            self._current_job = job.result()
 
         results = []
 
--- a/pennylane_qiskit/qiskit_device2.py
+++ b/pennylane_qiskit/qiskit_device2.py
@@ -523,6 +523,12 @@ class QiskitDevice2(Device):
 
     def _execute_runtime_service(self, circuits, session):
         """Execution using old runtime_service (can't use runtime sessions)"""
+
+        # The legacy backend.run() interface in Qiskit Runtime, which was used as the dedicated “direct hardware access” entry point, has been deprecated by Qiskit.
+        # The new SamplerV2 class now fulfills this role. Support for the backend.run() will be dropped on or around October 15, 2024.
+        # Please refer to the migration guide (https://docs.quantum.ibm.com/api/migration-guides/qiskit-runtime) for instructions on how to migrate any existing code.
+        # This corresponds to the "circuit-runner" and "qasm3-runner" programs if you are invoking the REST API directly.
+
         # update kwargs in case Options has been modified since last execution
         self._update_kwargs()
 
--- a/pennylane_qiskit/qiskit_device2.py
+++ b/pennylane_qiskit/qiskit_device2.py
@@ -31,6 +31,7 @@ from qiskit.providers import BackendV2
 from qiskit_ibm_runtime import Session, Sampler, Estimator
 from qiskit_ibm_runtime.constants import RunnerResult
 from qiskit_ibm_runtime.options import Options
+from qiskit.providers.exceptions import QiskitBackendNotFoundError
 
 from pennylane import transform
 from pennylane.transforms.core import TransformProgram
@@ -475,6 +476,7 @@ class QiskitDevice2(Device):
         circuits: QuantumTape_or_Batch,
         execution_config: ExecutionConfig = DefaultExecutionConfig,
     ) -> Result_or_ResultBatch:
+
         session = self._session or Session(backend=self.backend)
 
         if not self._use_primitives:
--- a/pennylane_qiskit/qiskit_device2.py
+++ b/pennylane_qiskit/qiskit_device2.py
@@ -476,11 +476,8 @@ class QiskitDevice2(Device):
         circuits: QuantumTape_or_Batch,
         execution_config: ExecutionConfig = DefaultExecutionConfig,
     ) -> Result_or_ResultBatch:
-        
-        try: 
-            session = self._session or Session(backend=self.backend)
-        except QiskitBackendNotFoundError:
-            session = None
+
+        session = self._session or Session(backend=self.backend)
 
         if not self._use_primitives:
             results = self._execute_runtime_service(circuits, session=session)
@@ -509,8 +506,7 @@ class QiskitDevice2(Device):
                     results.append(execute_fn(circ, session))
                 yield results
             finally:
-                if session:
-                    session.close()
+                session.close()
 
         with execute_circuits(session) as results:
             return results
--- a/pennylane_qiskit/qiskit_device2.py
+++ b/pennylane_qiskit/qiskit_device2.py
@@ -526,7 +526,7 @@ class QiskitDevice2(Device):
 
         # The legacy ``backend.run()`` interface in Qiskit Runtime, which was used as the dedicated “direct hardware access” entry point, has been deprecated by Qiskit.
         # The new SamplerV2 class now fulfills this role. Support for the backend.run() will be dropped on or around October 15, 2024.
-        # Please refer to the migration guide (https://docs.quantum.ibm.com/api/migration-guides/qiskit-runtime) for instructions on how to migrate any existing code.
+        # Please refer to the `migration guide <https://docs.quantum.ibm.com/api/migration-guides/qiskit-runtime>`_ for instructions on how to migrate any existing code.
         # This corresponds to the "circuit-runner" and "qasm3-runner" programs if you are invoking the REST API directly.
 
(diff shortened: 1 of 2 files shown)


austingmhuang:
[sc-62210]

austingmhuang:
[SC-55725]

tests/test_base_device.py
```diff
@@ -141,29 +138,30 @@ class TestSupportForV1andV2:
...
     @pytest.mark.parametrize(
-        "backend",
+        "backend, use_primitives, shape",
         [
-            FakeManila(),
-            FakeManilaV2(),
+            (FakeManila(), True, (1, 1024)),
+            (FakeManila(), False, (1024,)),
+            (FakeManilaV2(), True, (1, 1024)),
+            (FakeManilaV2(), False, (1024,)),
```
lillian542, line 154: I think this is fine for now if its hard to track down where its coming from, but the return shape should preferably be the same regardless of whether we are using primitives, so I would say you found a small bug here.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -550,21 +550,37 @@ def _execute_runtime_service(self, circuits, session):
...
+
         # Send circuits to the cloud for execution by the circuit-runner program.
-        job = self.service.run(
-            program_id="circuit-runner",
-            options=options,
-            inputs=program_inputs,
-            session_id=session.session_id,
-        )
-        self._current_job = job.result(decoder=RunnerResult)
+        # Cloud simulators will be deprecated on May 15th so this will be exclusively for real hardware devices.
```
lillian542, line 569: Just to confirm, have you run your hardware test notebook on this branch?

pennylane_qiskit/qiskit_device2.py
```diff
@@ -550,21 +550,37 @@ def _execute_runtime_service(self, circuits, session):
...
             "job_tags": self.options.environment.job_tags,
             "max_execution_time": self.options.max_execution_time,
         }
 
+        # Does not support AerSimulator specific options e.g. choose a specific method
+        # refer to this https://qiskit.github.io/qiskit-aer/stubs/qiskit_aer.AerSimulator.html
+        # To support this would be confusing in terms of option setting (The options we currently support are runtime options)
+        # This here is just to capture and track shots information
+        aer_options = {
+            "shots": circuits[0].shots.total_shots or self.shots.total_shots,
```
lillian542, line 565: Instead of defining shots two places, could we add `shots = circuits[0].shots.total_shots or self.shots.total_shots` above in line 538, and then use it both in `program_inputs` and in the `aer` run function? Future updates are always harder when identical things are defined multiple places. 

Also instead of defining a full `options` dictionary if this is the only thing in it, should we just pass the `shots` kwarg in `self.backend.set_options(**aer_options)`?

lillian542 (commented):
Thanks Austin! This is looking good. I think if possible we should try to test more things with the Qiskit hardware backend mocks as well, since it sounds like testing with the simulator backends will no longer tests the same API as the hardware backends. What are your thoughts?

tests/test_base_device.py
```diff
@@ -1166,7 +1170,6 @@ def test_process_estimator_job(self, measurements, expectation):
 
         for data in result.metadata:
             assert isinstance(data, dict)
-            assert list(data.keys()) == ["variance", "shots"]
```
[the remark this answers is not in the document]
austingmhuang: The differences are pretty significant. Technically I could hardcode it so that the formats match but that didn't seem like a good solution either so I deleted it for now. Let me know if you think there's some kind of better assertion here! Open to suggestions

tests/test_base_device.py
austingmhuang: > Is the discrepancy in shape only for `sample` and other measurements that go through the `execute_runtime` function, or does it happen for other execution types like `expval` or `probs` as well?

This is a good point. I didn't investigate **that** closely but I imagine it should be due to `execute_runtime` itself, so it should include other execution types like expval and probs. That's my guess though, I can write a note here to investigate further

austingmhuang: > I think this is fine for now if its hard to track down where its coming from, but the return shape should preferably be the same regardless of whether we are using primitives, so I would say you found a small bug here.

Yeah I think it's a small bug. Best to address in a separate PR though I figured.

pennylane_qiskit/qiskit_device2.py
austingmhuang: No but I should. I'll follow up on this.

tests/test_base_device.py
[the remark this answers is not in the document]
lillian542: I don't think this was an important line of the test, I'm happy to just have it removed 😅

tests/test_base_device.py
lillian542: Makes sense - I'd suggest adding a shortcut story to track it, but once that's done, I'd say this is good for this PR

pennylane_qiskit/qiskit_device2.py
austingmhuang: > Instead of defining shots two places, could we add `shots = circuits[0].shots.total_shots or self.shots.total_shots` above in line 538, and then use it both in `program_inputs` and in the `aer` run function? Future updates are always harder when identical things are defined multiple places.
> 
> Also instead of defining a full `options` dictionary if this is the only thing in it, should we just pass the `shots` kwarg in `self.backend.set_options(**aer_options)`?

Yeah I agree some refactoring is necessary here. I defined a full 'options' dictionary at first because I thought there would be more compatible options with the AerSimulator 😅 (seems like other than 'shots' nothing else though)

tests/test_base_device.py
```diff
@@ -115,14 +117,9 @@ def close(self):  # This is just to appease a test
...
 
-# pylint: disable=bare-except
-try:
-    service = QiskitRuntimeService(channel="ibm_quantum")
-    backend = service.backend("ibmq_qasm_simulator")
-except:
-    backend = MockedBackend()
-
+mocked_backend = MockedBackend()
 legacy_backend = MockedBackendLegacy()
```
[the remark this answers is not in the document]
austingmhuang: Yeah that one in my mind is a separate PR. I've been looking at `conftest.py`, and as a whole might it probably needs to be updated to use ibm_runtime_service since ibm_provider has officially been deprecated. With the cloud simulators being deprecated as well it's probably a good time to make some changes there.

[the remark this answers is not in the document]
lillian542: Awesome, sounds like a plan 🚀

tests/test_base_device.py
austingmhuang: [sc-62867] Added!

pennylane_qiskit/qiskit_device2.py
austingmhuang: There are some issues with running this on real hardware. Seems like last time I used the hardware I ran circuits that had operations that were supported by the hardware but this is not always the case. Per Qiskit's documentation, one needs to do the transpilation beforehand before submitting it to a real hardware device. I created a separate story to address this issue. Shouldn't be that difficult but it falls under the general "migrate everything" task, so I will probably tackle all the migration stuff at the same time.

For reference:
[sc-62928]
[Qiskit docs](https://docs.quantum.ibm.com/run/primitives-examples)

Additionally, the legacy backend.run() interface in Qiskit Runtime, which was used as the dedicated “direct hardware access” entry point, has been deprecated by Qiskit. The new SamplerV2 class now fulfills this role. Support for the backend.run() will be dropped on or around October 15, 2024. Please refer to the migration guide (https://docs.quantum.ibm.com/api/migration-guides/qiskit-runtime) for instructions on how to migrate any existing code. This corresponds to the "circuit-runner" and "qasm3-runner" programs if you are invoking the REST API directly.

austingmhuang:
@lillian542 @obliviateandsurrender I think this looks ready. There is some additional work to be done for hardware support and for additional tests on that front (mocks, etc...). I have created 2 PRs related to this but probably need one more for the mocking/conftest changes.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -523,13 +523,21 @@ def execute_circuits(session):
 
     def _execute_runtime_service(self, circuits, session):
         """Execution using old runtime_service (can't use runtime sessions)"""
+
+        # The legacy backend.run() interface in Qiskit Runtime, which was used as the dedicated “direct hardware access” entry point, has been deprecated by Qiskit.
```
obliviateandsurrender, line 527:
```suggestion
        # The legacy ``backend.run()`` interface in Qiskit Runtime, which was used as the dedicated “direct hardware access” entry point, has been deprecated by Qiskit.
```

pennylane_qiskit/qiskit_device2.py
```diff
@@ -523,13 +523,21 @@ def execute_circuits(session):
...
+        # The new SamplerV2 class now fulfills this role. Support for the backend.run() will be dropped on or around October 15, 2024.
+        # Please refer to the migration guide (https://docs.quantum.ibm.com/api/migration-guides/qiskit-runtime) for instructions on how to migrate any existing code.
```
obliviateandsurrender, line 529: As a good practice, we should try having the correct syntax for docstrings of private methods as well. 

```suggestion
        # Please refer to the `migration guide <https://docs.quantum.ibm.com/api/migration-guides/qiskit-runtime>`_ for instructions on how to migrate any existing code.
```

pennylane_qiskit/qiskit_device2.py
```diff
@@ -538,7 +546,7 @@ def _execute_runtime_service(self, circuits, session):
 
         program_inputs = {
             "circuits": compiled_circuits,
```
obliviateandsurrender, line 548: Does the compilation of the circuit use Qiskit's transpilation?

pennylane_qiskit/qiskit_device2.py
```diff
@@ -550,21 +558,33 @@ def _execute_runtime_service(self, circuits, session):
...
+                session_id=session.session_id,
+            )
+            self._current_job = job.result(decoder=RunnerResult)
+        else:  # Uses local simulator instead. After May 15th, all simulations will use this logic instead.
+            # Does not support AerSimulator specific options e.g. choose a specific method
+            # refer to this https://qiskit.github.io/qiskit-aer/stubs/qiskit_aer.AerSimulator.html
+            # To support this would be confusing in terms of option setting (The options we currently support are runtime options)
+            # This here is just to capture and track shots information
+            self.backend.set_options(shots=shots)
+            job = self.backend.run(
```
obliviateandsurrender, line 584: Do we have tests for executing this, or is writing them not so straightforward?

pennylane_qiskit/qiskit_device2.py
austingmhuang: Not at the moment. This is a 1 to 1 conversion of a circuit written in Qiskit to a circuit written in Pennylane. 

There should be an additional, separate step like transpile_qiskit_circuit that takes the qiskit_circ and transpiles it according to the specifications of the backend. It should be a separate step since this is only necessary if you want to access the hardware.

pennylane_qiskit/qiskit_device2.py
austingmhuang: The tests for tracking the shots for the local simulator exist and are passing. It's not clearly reflected in the PR but the test `test_tape_shots_used_runtime_service` has had its backend updated to use the AerSimulator instead of the cloud simulator and the test is passing.

pennylane_qiskit/qiskit_device2.py
lillian542: > one needs to do the transpilation beforehand before submitting it to a real hardware device

Ah. That's new. Is it using the same `compile.transpile` function that the device already uses in `_execute_runtime_service`? If so we can just move all of that up to the main execute instead.

> "the legacy backend.run() interface in Qiskit Runtime, which was used as the dedicated “direct hardware access” entry point, has been deprecated by Qiskit. The new SamplerV2 class now fulfills this role. Support for the backend.run() will be dropped on or around October 15, 2024

Is `SamplerV2` supported before Qiskit 1.0? Or would switching to that force a dependency on 1.0? Also, do you know if "dropping support" means that they don't guarantee it and won't debug it, but you can still use it if you have an old version of Qiskit installed? Or does it mean the backends won't recognize it anymore, even if you have an older version of Qiskit?

austingmhuang: Discussed offline.

lillian542 (approved):
Lots of open questions, but none of them for this PR - looks good to me! 🎉

pennylane_qiskit/qiskit_device2.py
```diff
@@ -288,8 +288,7 @@ def __init__(
         self._backend = backend
         self._compile_backend = compile_backend if compile_backend else backend
 
-        # ToDo: possibly things fail if this is not a QiskitRuntimeService - confirm and decide how to handle (SC 55725)
-        self._service = backend._service
+        self._service = getattr(backend, "_service", None)
```
obliviateandsurrender, line 291: What happens with the fallback `None` value? Do we use it to raise some intuitive error?

pennylane_qiskit/qiskit_device2.py
```diff
@@ -523,13 +522,22 @@ def execute_circuits(session):
...
+        # ToDo: deprecate this by or around October 15, 2024.
+
         # update kwargs in case Options has been modified since last execution
         self._update_kwargs()
 
         # in case a single circuit is passed
         if isinstance(circuits, QuantumScript):
             circuits = [circuits]
 
+        shots = circuits[0].shots.total_shots or self.shots.total_shots
```
obliviateandsurrender, line 539: So, the underlying assumption is that all circuits (in the list) will have an equal number of `shots`?

pennylane_qiskit/qiskit_device2.py
```diff
@@ -550,21 +558,33 @@ def _execute_runtime_service(self, circuits, session):
...
         # Send circuits to the cloud for execution by the circuit-runner program.
-        job = self.service.run(
-            program_id="circuit-runner",
-            options=options,
-            inputs=program_inputs,
-            session_id=session.session_id,
-        )
-        self._current_job = job.result(decoder=RunnerResult)
+        # Cloud simulators will be deprecated on May 15th so this will be exclusively for real hardware devices.
+        if self.service:
```
obliviateandsurrender, line 570: Does this use from `self._service`? So when we have `None` there, we use a local simulator?

pennylane_qiskit/qiskit_device2.py
austingmhuang: If service is `None` it means that the backend being used is a local simulator. No need for an error.

pennylane_qiskit/qiskit_device2.py
austingmhuang: Suppose the user does something like `circuit(shots=[5, 10, 2])`, we raise a warning letting them know this is not supported due to the way qiskit takes in shots, and then run the circuit with the total amount of shots (17) in this case. If you wanted the exact functionality of circuit(shots=[5, 10, 2]), you just have to run the circuit three times manually e.g. circuit(shots=5), circuit(shots=10), circuit(shots=2). 

The reason is just because Qiskit doesn't support an array of shots like PL does, and the amount of work it would be to parse through an array of shots and send the job to qiskit 3 times and process the results would be challenging and not worth the time/effort.

We talked in more detail about how to track/deal with the number of shots in a different PR, but the solution is basically a compromise where we let the user run the circuit, but we also raise a warning explaining exactly what has happened and what to do instead.

obliviateandsurrender (approved):

pennylane_qiskit/qiskit_device2.py
austingmhuang: Yes, exactly.

PR #529, opened 2024-05-07
A translation function is added for converting the Qiskit probability output dictionary keys (integers indicating state) into the correct format for qml.probs

austingmhuang:
When using primitives the output for qml.probs on qiskitdevice2 is currently in the wrong format. This PR seeks to amend that so that the output formats are consistent.

[On hold for now because if we use PrimitivesV2 the solution will change...]

### DIFF
--- a/tests/test_base_device.py
+++ b/tests/test_base_device.py
@@ -1348,3 +1348,30 @@ class TestExecution:
             match="The observable measured",
         ):
             circuit()
+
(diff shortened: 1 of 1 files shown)


austingmhuang:
[sc-54389]

austingmhuang:
Closing this since these changes are reflected in #539

PR #530, opened 2024-05-09
Ensure results when use_primitives = False vs use_primitives=True are consistent. 

austingmhuang:
To explain this change, some context behind the bug is necessary. We were seeing a difference between running qml.sample() when use_primitives was set to True vs False. When use_primitives was set to True, we were getting a shape of (1, 1024) instead of the correct shape of (1024, ). At first glance it seems like this problem should be resolved in `execute` -> you can simply check the len of the result that's returned there and if it's 1, take the 0th element instead. Although obvious, surprisingly, doing this does not fix the bug.

The root of the problem is actually in this `split_execution_types` function.

When running `use_primitives = False`, you skip ` split_execution_types`, and therefore do not run this line of code. However, suppose `result = [1]`. When` use_primitives = True` and `len(result) == 1`, you will be going from` [1] -> (1,)`, which is bad since pennylane devices will return `1` instead. This is the source of the bug and it only occurs when a user sets use_primitives to false and has a circuit that only returns 1 measurement. (In other words, if the user had a circuit that returned multiple measurements, this bug would not occur).

### DIFF
--- a/pennylane_qiskit/qiskit_device2.py
+++ b/pennylane_qiskit/qiskit_device2.py
@@ -195,7 +195,12 @@ def split_execution_types(
 
         result = dict(zip(flattened_indices, flattened_results))
 
-        return tuple(result[i] for i in sorted(result.keys()))
+        result = tuple(result[i] for i in sorted(result.keys()))
+
+        # Need to check for shot_vector attribute because if an array of shots is inputted
+        # an array of results is expected by the device. Unfortunately, you cannot just
+        # "tell" the device we are not going to run an array of shots, so this is a bit of a hack
+        return result[0] if len(result) == 1 and not hasattr(tape.shots, "shot_vector") else result
 
     return tapes, reorder_fn
 
--- a/pennylane_qiskit/qiskit_device2.py
+++ b/pennylane_qiskit/qiskit_device2.py
@@ -288,8 +288,7 @@ class QiskitDevice2(Device):
         self._backend = backend
         self._compile_backend = compile_backend if compile_backend else backend
 
-        # ToDo: possibly things fail if this is not a QiskitRuntimeService - confirm and decide how to handle (SC 55725)
-        self._service = backend._service
+        self._service = getattr(backend, "_service", None)
         self._use_primitives = use_primitives
         self._session = session
 
@@ -523,12 +522,21 @@ class QiskitDevice2(Device):
 
     def _execute_runtime_service(self, circuits, session):
         """Execution using old runtime_service (can't use runtime sessions)"""
+
+        # The legacy ``backend.run()`` interface in Qiskit Runtime, which was used as the dedicated “direct hardware access” entry point, has been deprecated by Qiskit.
+        # The new SamplerV2 class now fulfills this role. Support for the backend.run() will be dropped on or around October 15, 2024.
+        # Please refer to the `migration guide <https://docs.quantum.ibm.com/api/migration-guides/qiskit-runtime>`_ for instructions on how to migrate any existing code.
+        # This corresponds to the "circuit-runner" and "qasm3-runner" programs if you are invoking the REST API directly.
+        # ToDo: deprecate this by or around October 15, 2024.
+
         # update kwargs in case Options has been modified since last execution
         self._update_kwargs()
 
         # in case a single circuit is passed
         if isinstance(circuits, QuantumScript):
             circuits = [circuits]
+
+        shots = circuits[0].shots.total_shots or self.shots.total_shots
 
         qcirc = [
             circuit_to_qiskit(circ, self.num_wires, diagonalize=True, measure=True)
@@ -538,7 +546,7 @@ class QiskitDevice2(Device):
 
         program_inputs = {
             "circuits": compiled_circuits,
-            "shots": circuits[0].shots.total_shots or self.shots.total_shots,
+            "shots": shots,
         }
 
         for kwarg, value in self._kwargs.items():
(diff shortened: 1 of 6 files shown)


austingmhuang:
[sc-62867]

pennylane_qiskit/qiskit_device2.py
```diff
@@ -195,7 +195,12 @@ def reorder_fn(res):
 
         result = dict(zip(flattened_indices, flattened_results))
 
-        return tuple(result[i] for i in sorted(result.keys()))
+        result = tuple(result[i] for i in sorted(result.keys()))
+
+        # Need to check for shot_vector attribute because if an array of shots is inputted
+        # an array of results is expected by the device. Unfortunately, you cannot just
+        # "tell" the device we are not going to run an array of shots, so this is a bit of a hack
+        return result[0] if len(result) == 1 and len(tape.shots.shot_vector) == 1 else result
```
austingmhuang, line 203: Right now, this solution has a slightly buggy edge case:
If the User initializes a circuit with `use_primitives=True`, and that circuit only has 1 measurement AND that user set an array of `shots`, then the returned result will have the wrong format (it'll have an extra bracket around it). I cannot return result[0] because when the circuit is initialized with an array of shots, it expects an iterable. (You can see this error if you delete `and len(tape.shots.shot_vector) == 1` and run the test cases. Basically, since we are already ignoring the user when they set an array of shots (we just use the total number of shots), I think we could raise the warning here instead and also explain that their result MIGHT be wrong in the case that they set only one measurement in their circuit. Alternatively, we just raise an error and tell them this is a bad idea anyway because Qiskit doesn't do an array of shots. I'm open to either solution, or if someone has a better solution where I can avoid all of this that would be good too.

tests/test_base_device.py
```diff
@@ -1204,9 +1208,53 @@ def test_generate_samples(self, num_wires, num_shots):
...
+        false_res = circuit_false(angle, shots=1024)
+        pl_res = circuit_pl(angle, shots=1024)
+
+        assert np.shape(true_res) == np.shape(pl_res)
+        assert np.shape(false_res) == np.shape(pl_res)
+
     def test_tape_shots_used_runtime_service(self, mocker):
         """Tests that device uses tape shots rather than device shots for _execute_runtime_service"""
-        dev = QiskitDevice2(wires=5, backend=backend, shots=2, use_primitives=True)
+        dev = QiskitDevice2(wires=7, backend=backend, shots=2, use_primitives=True)
```
austingmhuang, line 1257: Changing this because sometimes the assertion passes by accident because the number of wires is also equal to the number of shots. Anyway.

tests/test_base_device.py
```diff
@@ -1204,9 +1208,53 @@ def test_generate_samples(self, num_wires, num_shots):
...
+        [
+            lambda obs: qml.expval(obs),
+            lambda obs: qml.counts(obs),
+            lambda obs: qml.var(obs),
+        ],  # pylint: disable=unnecessary-lambda
+    )
+    @pytest.mark.parametrize("angle", [np.pi / 2, np.pi / 3, np.pi / 4])
+    def test_behavior_for_use_primitives(self, angle, measurement, observable):
+        """Tests that the format for results from ``use_primitives=True`` and ``use_primitives=False``
+        are correct. ``qml.probs()`` is excluded due to the fact that it does not take in an
```
austingmhuang, line 1223: qml.probs() will be addressed in a [separate PR](https://github.com/PennyLaneAI/pennylane-qiskit/pull/529)

tests/test_base_device.py
lillian542: 😅

pennylane_qiskit/qiskit_device2.py
lillian542: There's a lot going on at once in this device and we aren't really able to support `shot_vector` support anyway, so I'm in favour of raising an error if we encounter one. If it's something we want maybe it can be added in later but I'd say it's pretty low priority.

austingmhuang: I'll raise an error for now.

tests/test_base_device.py
```diff
@@ -1001,8 +1005,8 @@ def test_shot_vector_warning_mocked(self):
         )
 
         with patch.object(dev, "_execute_estimator"):
-            with pytest.warns(
-                UserWarning,
+            with pytest.raises(
+                ValueError,
                 match="Setting shot vector",
             ):
```
obliviateandsurrender, line 1011:
```suggestion
            with pytest.raises(ValueError, match="Setting shot vector"):
```

tests/test_base_device.py
obliviateandsurrender: We could use the keyword argument like `lambda obs: qml.measurement(op=obs)`, which should work for `probs` as well, right?

obliviateandsurrender (commented):
Looking great, just leaving a couple of suggestions.

tests/test_base_device.py
austingmhuang: Yes that works. I didn't write a test because right now the output from probs() is already different from the output from pennylane -- this has its own PR so I'll put it off for that PR :)

tests/test_base_device.py
```diff
@@ -403,7 +403,11 @@ def test_split_execution_types(self, measurements, expectation):
         assert [tape.measurements for tape in tapes] == expectation
 
         # reorder_fn puts them back
-        assert reorder_fn([tape.measurements for tape in tapes]) == tuple(qs.measurements)
+        assert (
+            reorder_fn([tape.measurements for tape in tapes]) == qs.measurements[0]
+            if len(qs.measurements) == 0
```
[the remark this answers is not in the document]
austingmhuang: Nice catch. Actually this whole assertion was wrong because it was just asserting tuple(qs.measurements) 😅

austingmhuang:
Closing this since the changes are now in #539

PR #531, opened 2024-05-10
Now transpiles for estimator and sampler

austingmhuang:
Without transpilation, you can't run Estimators and Samplers using Qiskit 1.0. This is a simple change that just transpiles circuits so that it can be submitted to a real hardware backend. Tested on a local notebook that access to the device works as intended and the answers seem reasonable (other than some formatting issues that a different PR is addressing).

### DIFF
--- a/pennylane_qiskit/qiskit_device2.py
+++ b/pennylane_qiskit/qiskit_device2.py
@@ -605,12 +605,13 @@ class QiskitDevice2(Device):
     def _execute_sampler(self, circuit, session):
         """Execution for the Sampler primitive"""
 
-        qcirc = circuit_to_qiskit(circuit, self.num_wires, diagonalize=True, measure=True)
+        qcirc = [circuit_to_qiskit(circuit, self.num_wires, diagonalize=True, measure=True)]
         if circuit.shots:
             self.options.execution.shots = circuit.shots.total_shots
         sampler = Sampler(session=session, options=self.options)
+        compiled_circuits = self.compile_circuits(qcirc)
 
-        result = sampler.run(qcirc).result()
+        result = sampler.run(compiled_circuits).result()
         self._current_job = result
 
         # needs processing function to convert to the correct format for states, and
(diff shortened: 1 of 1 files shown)


austingmhuang:
[sc-62928]

austingmhuang:
Can hold off on this one while we decide on if we are going to change to using V2 primitives instead.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -623,18 +624,21 @@ def _execute_sampler(self, circuit, session):
     def _execute_estimator(self, circuit, session):
         # the Estimator primitive takes care of diagonalization and measurements itself,
         # so diagonalizing gates and measurements are not included in the circuit
-        qcirc = circuit_to_qiskit(circuit, self.num_wires, diagonalize=False, measure=False)
+        qcirc = [circuit_to_qiskit(circuit, self.num_wires, diagonalize=False, measure=False)]
         if circuit.shots:
             self.options.execution.shots = circuit.shots.total_shots
         estimator = Estimator(session=session, options=self.options)
+        compiled_circuits = self.compile_circuits(qcirc)
```
obliviateandsurrender, line 631: I think I have asked this before, but to confirm that this is not the PR that we are compiling based on the device? 🤔

austingmhuang: I think I gave you the wrong answer last time 😅 . 

When you run compile_circuits, it compiles based on the device backend.

austingmhuang: In further detail:

I was confused and thought we didn't transpile when `use_primitives=False`. But actually we were transpiling when `use_primitives=False` and **not** transpiling when` use_primitives = True`. This PR just makes it so when `use_primitives=True`, we transpile.

austingmhuang:
This is just a few lines of code so I merged it into #539 as without the transpilation none of the code works 😅

PR #533, opened 2024-05-14
Tracker functionality

austingmhuang:
This PR adds the tracker functionality to the Qiskit device. There is a small bug with the formatting of the output from `qiskit_tracker.history["results"]` (You get an array of tuples instead of values). I think the bug is due to the post-processing of the device, and I'm pretty sure it's just cosmetic rather than having serious implications (the values in the tuples are correct). Not sure if a straight-forward fix exists for this but in any case I don't think it should block this PR.


austingmhuang:
[sc-54386]

tests/test_base_device.py
```diff
@@ -633,6 +633,65 @@ def test_num_wires_property(self):
...
+        assert qiskit_tracker.history["batches"] == tracker.history["batches"]
+        assert tracker.latest.keys() == qiskit_tracker.latest.keys()
+        assert tracker.history.keys() == qiskit_tracker.history.keys()
+        assert tracker.history["shots"] == qiskit_tracker.history["shots"]
+
+        print(qiskit_tracker.history["batches"])
+        print(tracker.history["batches"])
+
+        # TODO: qiskit_tracker.history["results"] has a format that is inconsistent
+        # to tracker.history["results"], but this is due to the post processing
```
obliviateandsurrender, line 669: Is there an easy way around it or to manually revert the `post-processing` changes? 🤔

obliviateandsurrender: Or maybe look at the `simulator_tracking` and see if we could modify it for the qiskit device?

pennylane_qiskit/qiskit_device2.py
```diff
@@ -207,6 +209,18 @@ class QiskitDevice2(Device):
...
+        :title: Tracking
+
+        ``QiskitDevice2`` tracks:
+
+        * ``executions``: the number of unique circuits that would be required on quantum hardware
+        * ``shots``: the number of shots
+        * ``resources``: the :class:`~.resource.Resources` for the executed circuit.
+        * ``simulations``: the number of simulations performed. One simulation can cover multiple QPU executions, such as for non-commuting measurements and batched parameters.
+        * ``batches``: The number of times :meth:`~.execute` is called.
+        * ``results``: The results of each call of :meth:`~.execute`
```
lillian542, line 223: Do we want the tracker on this device to distinguish between `simulations` and `executions` like the simulator tracker does? And have we thought about how much something like `executions` means when with `Estimator` we send things along to Qiskit to let them deal with non-commuting measurements, etc?

tests/test_base_device.py
austingmhuang: > Is there an easy way around it or to manually revert the `post-processing` changes? 🤔

Maybe there is an easy way around it but I couldn't find one 😅 

> Or maybe look at the simulator_tracking and see if we could modify it for the qiskit device?

I'll look into this...

pennylane_qiskit/qiskit_device2.py
austingmhuang: >Do we want the tracker on this device to distinguish between simulations and executions like the simulator tracker does? 

Hmm I don't think it distinguishes between simulations and executions right now. I'm actually not sure if Qiskit distinguishes simulations/executions between running on Aer vs. on real hardware. In any case I don't think it's important/useful to distinguish between simulations and executions. It might be a bit confusing though to have `simulations` exist though.

>And have we thought about how much something like executions means when with Estimator we send things along to Qiskit to let them deal with non-commuting measurements, etc?

Counting `Executions` are different for the QiskitDevice compared to default.qubit when given non-commuting measurements -- so right now each it's been split, it's counted as a separate execution.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -53,6 +64,136 @@
...
+
+    Returns
+        type: The inputted class that has now been modified to update the tracker upon function calls.
+
+    Simulator style tracking updates:
+
+    * ``executions``: the number of unique circuits that would be required on quantum hardware
+    * ``shots``: the number of shots
+    * ``resources``: the :class:`~.resource.Resources` for the executed circuit.
+    * ``"errors"``: combined algorithmic errors from the quantum operations executed by the qnode.
```
austingmhuang, line 122: Is there a way to do some kind of wrapping for this instead? I literally experimented for like 3 hours today but it kept on running the original tracker function as well.

```
from pennylane.devices.modifiers import simulator_tracking

def custom_simulator_tracking(cls):
    """Decorator that adds custom tracking to the device class."""
    cls = simulator_tracking(cls)
    
    original_init = cls.__init__
    
    def new_init(self, *args, **kwargs):
        original_init(self, *args, **kwargs)
        self.execute = _track_execute(self.execute)
    
    cls.__init__ = new_init
    return cls
```

I tried something like this but it is bugged basically. My impression is that this should be doable without copy-paste bruteforce but I'm pretty unfamiliar with python wrappers/decorators so...

obliviateandsurrender (approved):

tests/test_base_device.py
```diff
@@ -729,6 +729,66 @@ def test_num_wires_property(self):
...
+        """Test that the tracker works for batched circuits"""
+        dev = qml.device("default.qubit", wires=1, shots=10000)
+        qiskit_dev = QiskitDevice2(wires=1, backend=AerSimulator(), shots=10000)
+
+        x = np.array(0.1, requires_grad=True)
+
+        @qml.qnode(dev, diff_method="parameter-shift")
+        def circuit(x):
+            qml.RX(x, wires=0)
+            return qml.expval(qml.Z(0))
```
lillian542, line 743: This is a single very simple example of what we might track. It would be nice to include some tests that have more complex returns. I would also question the assumption that comparing the tracker on the Qiskit device to the tracker on `default.qubit` for the same circuit is a good test in this case, because, since the devices behave differently, we shouldn't expect their trackers to match consistently.

Here are some examples of possible return statements where I think things are off (the circuits will need to just be called instead of calling `grad` on them, since they have multiple outputs):

- `return qml.expval(qml.Z(0)), qml.expval(qml.X(1)), qml.counts(qml.X(2))` : for `default.qubit`, this should (and does) assume 1 execution, because everything commutes. But `expval` and `counts` go to different tapes, so I would expect the Qiskit device to show 2 executions, since ultimately two tapes are executed when calling `qiskit_circuit(x)`. In this case, the two devices do match, but I would expect them not to.
- `return qml.expval(qml.Z(0)), qml.expval(qml.X(0))`: here I would expect the results for this device and `default.qubit` to match, but instead it splits the `qiskit_device` return into multiple calls. I think this is less of a problem, because the `tracker.history` still has reasonable results, it's just that they are split up differently.

lillian542 (commented):
This behaves a bit differently than I would expect it to, and I think re-uses too much from PennyLane, where the code was written for simulators and without specific regard for the Qiskit interface. 

I would assume that `executions` tracks how many times one of the device execute functions is called, which will depend on measurement types and commutation relations.  The `default.qubit` tracker takes the commutation relation into account when deciding the number of executions, but doesn't know anything about how many tapes are ultimately executed because of things specific to this device regarding Primitive types. I've added a comment with some examples of where this behaves differently than I would have expected. 

I also don't think the tracker for this device should include "simulations". This device requires shots and either connects to or mimics hardware. That's part of the `simulator_tracker` interface, but it doesn't make sense for a hardware plugin.

austingmhuang:
>Right, I'm not sure the Qiskit device tracker should have simulations, since its primary use will be hardware or mimicking hardware. If we move the Aer and BasicSimulator devices to the new API, then this tracker would seem very suitable for them. But for the remote device, I'm wondering if we need something that's less based on the assumptions of PennyLane simulators.

This is fair. I didn't make a decision on whether or not if it should have simulations or not but in my mind, if you were to run this using real hardware and tracked it you should probably just ignore this keyword since it's clearly not relevant for you. But if you used AerSimulator() then maybe this keyword should exist? In the ideal case, we should have a conditional somewhere that basically checks if you have a simulator or a real device but not sure if that's really "worth" the time/effort. Alternatively we just delete the `simulations` key altogether, since I think `executions` pretty much is the same thing? I'm pretty impartial to either solution so I just left it as is.

tests/test_base_device.py
austingmhuang: >`return qml.expval(qml.Z(0)), qml.expval(qml.X(1)), qml.counts(qml.X(2))` : for default.qubit, this should (and does) assume 1 execution, because everything commutes. But expval and counts go to different tapes, so I would expect the Qiskit device to show 2 executions, since ultimately two tapes are executed when calling qiskit_circuit(x). In this case, the two devices do match, but I would expect them not to.

I am a little confused. When I ran this, I got `'executions': 2` for the qiskit device and `'executions:1'` for the default device (Which is... what you want right? This is the intended behaviour on my part as well). I can add an additional test case for this (in fact... I should... my bad... let me do that...) but maybe this is bugged somehow?

>`return qml.expval(qml.Z(0)), qml.expval(qml.X(0))`: here I would expect the results for this device and default.qubit to match, but instead it splits the qiskit_device return into multiple calls. I think this is less of a problem, because the tracker.history still has reasonable results, it's just that they are split up differently.

Here I expect the `qiskit_device` to be split up as well, since it's going to make two separate calls to the Estimator. 

In general, my intention was to have the `executions` to match as closely as possible to the number of calls to the IBM backends, so users don't have to check their IBM account dashboards to know how many calls occurred (although, they really should for peace of mind). So far I think this is pretty much working as intended (although the tests cases do not really "test" for that... which is... very fixable!)

lillian542: Discussed offline, concluded that the tracker treats `latest` differently than I was expecting, but this is alright given how the device differs from the PennyLane devices. Specifically, `latest` reflects the latest call to run on Qiskit, not the latest call of the QNode.

Is summary, the tracker is counting what we want it to count, but the bookkeeping structure is different than what I expected. This seems like more work than it's worth to look into, and the existing version is accurate.

austingmhuang:
> This is fair. I didn't make a decision on whether or not if it should have simulations or not but in my mind, if you were to run this using real hardware and tracked it you should probably just ignore this keyword since it's clearly not relevant for you. But if you used AerSimulator() then maybe this keyword should exist? In the ideal case, we should have a conditional somewhere that basically checks if you have a simulator or a real device but not sure if that's really "worth" the time/effort. Alternatively we just delete the `simulations` key altogether, since I think `executions` pretty much is the same thing? I'm pretty impartial to either solution so I just left it as is.

Simulations keyword deleted :)

lillian542 (approved):

PR #536, opened 2024-05-16
Update reqs to 1.0

austingmhuang:
This new device will no longer support Qiskit <0.46. Therefore testing for those versions is no longer necessary and has been deleted from the CI workflow.


austingmhuang:
[sc-63543]

lillian542 (commented):
Looks good! A few small loose ends to wrap up:

1. lets also rename `tests_qiskit_1.yml` to `tests.yml`, since that is the standard name people will be looking for to find the test CI file
2. there are also 0.46 and 1.0 tests run twice a week for the test matrix (the IBMQ tests) - we should similarly delete the 0.46 version and rename the 1.0 version to `ibmq_tests.yml`. Long term those tests might not make sense as we move to local testing with FakeHardware, but for now let's just do that.
3. Lets add a comment on the PR for the feature branch targeting master, to remind us to update the Plugin Test Matrix when the branch is eventually merged - removing the IBMQ tests will need to be reflected [here](https://github.com/PennyLaneAI/plugin-test-matrix) (but not until the tests are removed on `master`)

lillian542:
Wrt codecov, it's just upset that initializing the `BasicAer` device isn't being tested. 

We can wait to merge this until we've merged a PR removing the `BasicAer` device, and then re-trigger CI and it should pass. Alternatively, we can have someone with authorization override CodeCov and merge.

austingmhuang:
> We can wait to merge this until we've merged a PR removing the `BasicAer` device, and then re-trigger CI and it should pass. Alternatively, we can have someone with authorization override CodeCov and merge.

I think we can wait and merge this after removing the `BasicAer` device. I'll open that PR ASAP.

austingmhuang:
> Looks good! A few small loose ends to wrap up:
> 
> 1. lets also rename `tests_qiskit_1.yml` to `tests.yml`, since that is the standard name people will be looking for to find the test CI file

Ok yeah, I didn't know that :sweat_smile: 

> 2. there are also 0.46 and 1.0 tests run twice a week for the test matrix (the IBMQ tests) - we should similarly delete the 0.46 version and rename the 1.0 version to `ibmq_tests.yml`. Long term those tests might not make sense as we move to local testing with FakeHardware, but for now let's just do that.

Ok, makes sense 👍 


> 3. Lets add a comment on the PR for the feature branch targeting master, to remind us to update the Plugin Test Matrix when the branch is eventually merged - removing the IBMQ tests will need to be reflected [here](https://github.com/PennyLaneAI/plugin-test-matrix) (but not until the tests are removed on `master`)

Done.

.github/workflows/ibmq_tests.yml
```diff
@@ -1,7 +1,7 @@
-name: IBMQ integration tests
+name: IBMQ integration tests with Qiskit 1.0
 on:
   schedule:
-    - cron: '0 0 * * 0,4' # At 00:00 on Sunday and Thursday.
+    - cron: '1 0 * * 0,4' # At 01:00 on Sunday and Thursday.
```
obliviateandsurrender, line 4: why `01:00`? 🤔

.github/workflows/tests.yml
```diff
@@ -72,19 +69,17 @@ jobs:
...
       - name: Install Plugin
         run: |
           python setup.py bdist_wheel
           pip install dist/PennyLane*.whl
-          pip freeze
 
       - name: Run tests
         run: |
-          pl-device-test --device=qiskit.basicaer --tb=short --skip-ops --shots=20000 --device-kwargs backend=qasm_simulator
+          pl-device-test --device=qiskit.basicsim --tb=short --skip-ops --shots=20000 --device-kwargs backend=basic_simulator
```
obliviateandsurrender, line 82: Is there any performance degradation for `basicsim` in comparison to `basicaer`?

obliviateandsurrender (commented):
About the tests being removed, will we be still testing for `0.46` or removing them fully? 🤔

.github/workflows/ibmq_tests.yml
austingmhuang: Oh actually I don't know. I deleted the old tests and renamed this one and I guess they had a different schedule? If it matters I can always change it back 😅

.github/workflows/tests.yml
austingmhuang: `Basicaer` is deprecated in >1.0, and `basicsim` is the 1.0 replacement for it. Shouldn't be any performance degradation as far as I can tell.

austingmhuang:
> About the tests being removed, will we be still testing for `0.46` or removing them fully? 🤔

We will be removing them fully. `0.46` and previous versions will no longer be supported.
TL;DR: In October Qiskit will no longer support 0.46 and previous versions accessing the real hardware, so it doesn't make sense for us to support it on the July release just to deprecate it in October.

tests/test_qiskit_device.py
```diff
@@ -172,17 +172,11 @@ def test_warning_raised_for_hardware_backend_analytic_expval(self, recorder):
...
-            record[0].message.args[0] == "The analytic calculation of "
-            "expectations, variances and probabilities is only supported on "
-            f"statevector backends, not on the {dev.backend.name}. Such statistics obtained from this "
-            "device are estimates based on samples."
-        )
+            _ = qml.device("qiskit.aer", backend="aer_simulator", wires=2, shots=None)
+
+        # Two warnings are being raised: one about analytic calculations and another about deprecation.
+        # Removed specific warning test
+        assert len(record) == 2
```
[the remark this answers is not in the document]
austingmhuang: I can change the test so that it checks the message text still

lillian542 (approved):
Looks good to me, thanks @austingmhuang !

obliviateandsurrender (approved):
:shipit:

PR #538, opened 2024-05-16
Remove use primitives and everything that depends on it

austingmhuang:
This story is one of many stories related to no longer providing backwards support for Qiskit 0.46 and upgrading to Qiskit 1.0. 

Going forward, `use_primitives` will always be `True` since we will not be providing support for using the old primitives-less `backend.run()` function (Qiskit's support for that will be discontinued around mid October). Thus, the keyword argument `use_primitives` is not necessary and should be deleted. 

As a result, the `backend.run()` function is also removed from this PR, since setting `use_primitives=False` is just a way to access this soon-to-be-defunct function. However, certain functionalities relied on the raw samples that the backend.run() function provided (e.g. measuring `qml.Hadamard()`), which means simply removing `backend.run()` will cause these tests to fail. In the interim, we will skip these tests for now and address the functionality in a [separate PR](https://github.com/PennyLaneAI/pennylane-qiskit/pull/539) where we upgrade to V2 primitives and implement all the functionality using those.

Additionally, `split_execution_types`, by necessity, is also modified to no longer have a `no_prim` array where we put measurements that needed to use the `backend.run()` function. We now put everything that is NOT covered by the `Estimator` directly to the `Sampler`. Using `SamplerV2`, we should be able to access the raw samples and get the same functionality that we had from `backend.run()`. 

**TLDR**
`backend.run()` removed
`use_primitives` removed
`split_execution_types` changed; `no_prim` taken over by `sampler`
`execute` changed; uses `sampler`
Functionality tests skipped here, addressed in #539


tests/test_base_device.py
```diff
@@ -1291,16 +1150,18 @@ def circuit():
         ],
     )
     @pytest.mark.filterwarnings("ignore::UserWarning")
+    @pytest.mark.skip(
+        reason="The functionality of using sampler to get the accurate answer is not yet implemented"
+    )
     def test_no_pauli_observable_gives_accurate_answer(self, mocker, observable):
```
austingmhuang, line 1156: We're skipping this until we implement the method in the `Sampler` where we use raw samples to compute the `expval` for things like `qml.Hadamard()` (no `pauli_rep`)

tests/test_base_device.py
```diff
@@ -1008,11 +896,6 @@ def test_shot_vector_warning_mocked(self):
                 dev.execute(qs)
 
 
-@pytest.mark.skipif(
-    Version(qiskit.__version__) < Version("1.0.0"),
-    reason="Session initialization is not supported for local simulators for Qiskit version < 1.0/qiskit_ibm_runtime version < 0.22.0",
-    ## See https://docs.quantum.ibm.com/api/migration-guides/local-simulators for additional details
-)
```
austingmhuang, line 1015: Not needed now that version will always be >1.0

tests/test_base_device.py
```diff
@@ -954,45 +877,10 @@ def test_execute_sampler_mocked(self, mocked_sampler, session):
         # to emphasize, this did nothing except appease CodeCov
         assert isinstance(result[0], Mock)
 
-    @patch("pennylane_qiskit.qiskit_device2.transpile")
-    def test_execute_runtime_service_mocked(self, mocked_transpile):
-        """Test the _execute_sampler function using a mocked version of Sampler
-        that returns a meaningless result."""
-
-        dev = QiskitDevice2(wires=5, backend=backend, use_primitives=True)
```
austingmhuang, line 962: We will write a replacement test for this one using `Sampler` instead for a separate PR.

tests/test_base_device.py
```diff
@@ -142,26 +141,22 @@ class TestSupportForV1andV2:
...
+            (FakeManila(), (1, 1024)),
+            (FakeManilaV2(), (1, 1024)),
         ],
     )
-    @pytest.mark.skipif(
-        Version(qiskit.__version__) < Version("1.0.0"),
-        reason="Session initialization is not supported for local simulators for Qiskit version < 1.0/qiskit_ibm_runtime version < 0.22.0",
-        ## See https://docs.quantum.ibm.com/api/migration-guides/local-simulators for additional details
+    @pytest.mark.skip(
+        reason="The functionality of using sampler to get an accurate answer is not yet implemented"
```
austingmhuang, line 155: Skipping this since we haven't implemented the functionality of using `Sampler` with raw samples. This is done in [this PR](https://github.com/PennyLaneAI/pennylane-qiskit/pull/539)

tests/test_base_device.py
```diff
@@ -325,8 +314,7 @@ class TestDevicePreprocessing:
                 ],
                 [
                     [qml.expval(qml.PauliZ(1)), qml.var(qml.PauliY(0))],
-                    [qml.probs(wires=[2])],
-                    [qml.counts()],
+                    [qml.counts(), qml.probs(wires=[2])],
```
austingmhuang, line 317: No longer separate and now both go into the same Sampler array.

tests/test_base_device.py
```diff
@@ -1176,6 +1055,7 @@ def test_process_estimator_job(self, measurements, expectation):
 
     @pytest.mark.parametrize("num_wires", [1, 3, 5])
     @pytest.mark.parametrize("num_shots", [50, 100])
+    @pytest.mark.skip(reason="Need to replace this with using SamplerV2.")
```
austingmhuang, line 1058: Works as is without skipping but it calls backend.run which we don't really care about anymore.

tests/test_base_device.py
```diff
@@ -439,17 +427,17 @@ def test_observable_stopping_condition(self, obs, expected):
...
-            ([qml.probs(wires=[0]), qml.counts(), qml.var(qml.PauliY(2))], 3),
+            ([qml.probs(wires=[0]), qml.counts(), qml.var(qml.PauliY(2))], 2),
         ],
     )
     def test_preprocess_splits_incompatible_primitive_measurements(self, measurements, num_types):
         """Test that the default behaviour for preprocess it to split the tapes based
-        on meausrement type. Expval and Variance are one type (Estimator), Probs another (Sampler),
-        and everything else a third (raw sample-based measurements)."""
+        on meausrement type. Expval and Variance are one type (Estimator), Probs and raw-sample based measurements
+        are another type(Sampler)."""
```
obliviateandsurrender, line 438:
```suggestion
        are another type (Sampler)."""
```

tests/test_base_device.py
```diff
@@ -887,13 +817,10 @@ def test_execute_pipeline_primitives_no_session(self, mocker):
 
     @pytest.mark.parametrize("backend", [backend, legacy_backend])
     def test_execute_pipeline_with_all_execute_types_mocked(self, mocker, backend):
-        """Test that a device that **is** using Primitives calls the _execute_runtime_service
-        to execute measurements that require raw samples, and the relevant primitive measurements
-        on the other measurements"""
+        """Test that a device executes measurements that require raw samples on the sampler,
+        and the relevant primitive measurements on the estimator"""
```
obliviateandsurrender, line 821:
```suggestion
        """Test that a device executes measurements that require raw samples from the sampler,
        and the relevant primitive measurements from the estimator"""
```

obliviateandsurrender (approved):
:shipit:

PR #539, opened 2024-05-16
Migrate to v2 primitives

austingmhuang:
This story is one of many stories related to no longer providing backwards support for Qiskit 0.46 and upgrading to Qiskit 1.0. 

**Summary of changes:**
In this PR, we upgrade from `EstimatorV1` and `SamplerV1` to their respective `V2` versions. We test and ensure that the format and results from the primitives are what we expect in comparison to the other pennylane devices. All the functionality that used to be addressed by `backend.run()` now uses `SamplerV2` instead (e.g. `expval(qml.Hadamard())`, `qml.probs()`).

**3 other related PRs closed by these changes**
Since the formatting was broken when `use_primitives=True`, we use code from [this pr #530](https://github.com/PennyLaneAI/pennylane-qiskit/pull/530) to fix this issue. For testing the comparison with the qml.probs() function, we use code from [this PR #529](https://github.com/PennyLaneAI/pennylane-qiskit/pull/529). In order to actually run the `Estimator` and the `Sampler`, we need to transpile beforehand as well. So we also use code from #531. Thus, this PR will close these three PRs as well.

**For a separate PR**
Although we have removed the `options` keyword argument and the use of `Options()`, we do not address how to handle the options and kwargs that a user should be able to provide the device with. For now, `_update_kwargs()` is commented out and not actually being called and the primitives are running with default options. Due to UI changes to both the `Estimator` and the `Sampler` (namely, the Estimator does not take in # of shots anymore -- it takes in some precision value), this will likely be a fair amount of changes. Since this PR is already quite large and works on its own, I think it is best to do this in a separate PR.


austingmhuang:
[sc-63540]

pennylane_qiskit/qiskit_device2.py
```diff
@@ -581,36 +556,49 @@ def _execute_runtime_service(self, circuits, session):
...
+        sampler = Sampler(session=session)
+        compiled_circuits = self.compile_circuits(qcirc)
 
-        result = sampler.run(qcirc).result()
-        self._current_job = result
+        result = sampler.run(compiled_circuits).result()[0]
+        attr = dir(result.data)[-1]
+        self._current_job = getattr(result.data, attr)
+
+        results = []
```
austingmhuang, line 567: I don't like this code and would prefer to just use `data.meas.get_counts()` as described in the [docs](https://docs.quantum.ibm.com/api/migration-guides/v2-primitives#run-multiple-experiments-in-a-single-job-1), but it seems that somehow the transpilation process renames this attribute according to what the classical register is named. Basically, `meas` is not guaranteed to be an attribute that you can access and call `get_counts()` on. So this is the solution I've arrived on for now.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -581,36 +556,49 @@ def _execute_runtime_service(self, circuits, session):
...
+            self._samples = self.generate_samples(index)
+            res = [
+                mp.process_samples(self._samples, wire_order=self.wires)
+                for mp in circuit.measurements
+            ]
+            single_measurement = len(circuit.measurements) == 1
+            res = res[0] if single_measurement else tuple(res)
+            results.append(res)
+
+        return tuple(results)
```
austingmhuang, line 584: You now need to process the raw samples that you get from the SamplerV2.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -581,36 +556,49 @@ def _execute_runtime_service(self, circuits, session):
     def _execute_sampler(self, circuit, session):
         """Execution for the Sampler primitive"""
 
-        qcirc = circuit_to_qiskit(circuit, self.num_wires, diagonalize=True, measure=True)
-        if circuit.shots:
-            self.options.execution.shots = circuit.shots.total_shots
-        sampler = Sampler(session=session, options=self.options)
+        qcirc = [circuit_to_qiskit(circuit, self.num_wires, diagonalize=True, measure=True)]
+        sampler = Sampler(session=session)
+        compiled_circuits = self.compile_circuits(qcirc)
```
austingmhuang, line 561: Transpilation has been added in this PR. It was a separate PR before but since it is only a few lines (8 lines) and the tests will fail without transpilation I decided to do it in this PR.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -581,36 +556,49 @@ def _execute_runtime_service(self, circuits, session):
...
         # the Estimator primitive takes care of diagonalization and measurements itself,
         # so diagonalizing gates and measurements are not included in the circuit
-        qcirc = circuit_to_qiskit(circuit, self.num_wires, diagonalize=False, measure=False)
-        if circuit.shots:
-            self.options.execution.shots = circuit.shots.total_shots
-        estimator = Estimator(session=session, options=self.options)
+        qcirc = [circuit_to_qiskit(circuit, self.num_wires, diagonalize=False, measure=False)]
+        estimator = Estimator(session=session)
 
+        compiled_circuits = self.compile_circuits(qcirc)
```
austingmhuang, line 592: Transpilation here as well.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -622,8 +610,9 @@ def _process_estimator_job(measurements, job_result):
         along with some metadata. Extract the relevant number for each measurement process and
         return the requested results from the Estimator executions."""
 
-        expvals = job_result.values
-        variances = [res["variance"] for res in job_result.metadata]
+        expvals = [res.data.evs.item() for res in job_result]
+        variances = [res.data.stds.item()**2*4096 for res in job_result] # this 4096 is the # of shots
+        # ToDo: Track the # of shots and use that to calculate the variance
```
austingmhuang, line 615: Qiskit removed the information for variances, opting to give `stds` instead. This means we have to reverse engineer `var` from `stds`, which requires the # of shots ran. For now I'm using their default # of shots because the options (and by extension shots) processing will be done in a separate PR.

austingmhuang:
[sc-62867]

austingmhuang:
[sc-54389]

austingmhuang:
[sc-62928]

pennylane_qiskit/qiskit_device2.py
```diff
@@ -180,23 +179,11 @@ def reorder_fn(res):
 
         result = dict(zip(flattened_indices, flattened_results))
 
-        return tuple(result[i] for i in sorted(result.keys()))
-
-    return tapes, reorder_fn
+        result = tuple(result[i] for i in sorted(result.keys()))
 
+        return result[0] if len(result) == 1 else result
```
austingmhuang, line 184: This is a bug fix for the formatting of the return values. See #530

pennylane_qiskit/qiskit_device2.py
```diff
@@ -180,23 +179,11 @@ def reorder_fn(res):
...
-    # this will break (or at least overwrite potentially relevant information)
-    # if they name things in some categories on Options the same as things in
-    # other categories, but at that point they've really departed from the kwarg API
-    options_dict = {}
-    for key, val in vars(options).items():
-        if hasattr(val, "__dict__"):
-            options_dict.update(qiskit_options_to_flat_dict(val))
-        elif val is not None:
-            options_dict[key] = val
-    return options_dict
```
austingmhuang, line 199: No more options, no more `qiskit_options_to_flat_dict`

pennylane_qiskit/qiskit_device2.py
```diff
@@ -289,7 +266,7 @@ def __init__(
             raise ValueError(f"Backend '{backend}' supports maximum {available_qubits} wires")
 
         self.reset()
-        self._update_kwargs()
+        # self._update_kwargs()
```
austingmhuang, line 269: Temporary. We come back to this in #547

obliviateandsurrender: Better to remove it with a TODO rather than leaving it as a comment.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -474,9 +427,9 @@ def execute_circuits(session):
             try:
                 for circ in circuits:
                     if circ.shots and len(circ.shots.shot_vector) > 1:
-                        warnings.warn(
+                        raise ValueError(
                             f"Setting shot vector {circ.shots.shot_vector} is not supported for {self.name}."
-                            f"The circuit will be run once with {circ.shots.total_shots} shots instead."
+                            "Please use a single integer number of shots instead when specifying number of shots."
```
obliviateandsurrender, line 432:
```suggestion
                            "Please use a single integer instead when specifying the number of shots."
```

pennylane_qiskit/qiskit_device2.py
```diff
@@ -495,36 +448,50 @@ def execute_circuits(session):
...
-            self.options.execution.shots = circuit.shots.total_shots
-        sampler = Sampler(session=session, options=self.options)
+        qcirc = [circuit_to_qiskit(circuit, self.num_wires, diagonalize=True, measure=True)]
+        sampler = Sampler(session=session)
+        compiled_circuits = self.compile_circuits(qcirc)
 
-        result = sampler.run(qcirc).result()
-        self._current_job = result
+        result = sampler.run(compiled_circuits).result()[0]
+        classical_register_name = compiled_circuits[0].cregs[0].name
```
obliviateandsurrender, line 456: Are we sure we only need just one single register? 🤔

pennylane_qiskit/qiskit_device2.py
```diff
@@ -495,36 +448,50 @@ def execute_circuits(session):
...
 
+        compiled_circuits = self.compile_circuits(qcirc)
         # split into one call per measurement
         # could technically be more efficient if there are some observables where we ask
         # for expectation value and variance on the same observable, but spending time on
         # that right now feels excessive
 
         pauli_observables = [mp_to_pauli(mp, self.num_wires) for mp in circuit.measurements]
-        result = estimator.run([qcirc] * len(pauli_observables), pauli_observables).result()
+        compiled_circuits *= len(pauli_observables)
```
obliviateandsurrender, line 490: Are we making `len(pauli_observables)` copies of the complied circuit here? 
You might want to join lines 473 and 490 together.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -536,8 +503,11 @@ def _process_estimator_job(measurements, job_result):
         along with some metadata. Extract the relevant number for each measurement process and
         return the requested results from the Estimator executions."""
 
-        expvals = job_result.values
-        variances = [res["variance"] for res in job_result.metadata]
+        expvals = [res.data.evs.item() for res in job_result]
+        variances = [
+            res.data.stds.item() ** 2 * 4096 for res in job_result
+        ]  # this 4096 is the # of shots
```
obliviateandsurrender, line 509: A bit confused, the number of shots are hard-coded? Where does this `4096` value come from?

pennylane_qiskit/qiskit_device2.py
```diff
@@ -563,6 +532,7 @@ def generate_samples(self, circuit=None):
         Returns:
              array[complex]: array of samples in the shape ``(dev.shots, dev.num_wires)``
         """
+
```
obliviateandsurrender, line 535: Is this added by `black`?

tests/test_base_device.py
```diff
@@ -434,8 +421,8 @@ def test_observable_stopping_condition(self, obs, expected):
     )
     def test_preprocess_splits_incompatible_primitive_measurements(self, measurements, num_types):
         """Test that the default behaviour for preprocess it to split the tapes based
-        on meausrement type. Expval and Variance are one type (Estimator), Probs another (Sampler),
-        and everything else a third (raw sample-based measurements)."""
+        on meausrement type. Expval and Variance are one type (Estimator), Probs and raw-sample based measurements
+        are another type(Sampler)."""
```
obliviateandsurrender, line 425:
```suggestion
        are another type (Sampler)."""
```

tests/test_base_device.py
```diff
@@ -239,11 +221,12 @@ def test_backend_wire_validation(self, backend):
         with pytest.raises(ValueError, match="supports maximum"):
             QiskitDevice2(wires=500, backend=backend)
 
+    @pytest.mark.skip(reason="Options handling not decided on yet")
     def test_setting_simulator_noise_model(self):
         """Test that the simulator noise model saved on a passed Options
         object is used to set the backend noise model"""
 
-        options = Options()
+        options = {}
```
obliviateandsurrender, line 229: Has the `Options` changed as well?

obliviateandsurrender (commented):
Good work @austingmhuang! Leaving a few initial comments from the first pass.

pennylane_qiskit/qiskit_device2.py
austingmhuang: Don't think so, probably added it by accident during merge conflicts

pennylane_qiskit/qiskit_device2.py
austingmhuang: I'm not 100% sure but it passes all the tests that we have 🤷 

@lillian542 what do you think?

pennylane_qiskit/qiskit_device2.py
austingmhuang: > Are we making len(pauli_observables) copies of the complied circuit here?
Yep
> You might want to join lines 473 and 490 together.
👍

pennylane_qiskit/qiskit_device2.py
austingmhuang: Right now, the number of shots are hard-coded since this PR does not address how we process keyword arguments. Therefore, I've hard coded to # of shots here to be the default amount of shots that QIskit uses to pass the tests for now. In #547, this will be a variable either `circuit.shots.total_shots` or `dev.shots`.

tests/test_base_device.py
austingmhuang: `Options` has been deleted due to changes by Qiskit. [See this](https://docs.quantum.ibm.com/api/migration-guides/v2-primitives#steps-to-migrate-to-estimator-v2).

Now they want you to specify options with either some kind of dictionary using options={"k": v} where they'll parse that object to get keyword arguments or just use keyword arguments to specify options.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -275,9 +254,7 @@ def __init__(
         self._init_kwargs = kwargs
         # _kwargs are used instead of the Options for performing raw sample based measurements (using old Qiskit API)
         # the _kwargs are a combination of information from Options and _init_kwargs
-        self._kwargs = None
-        if self.options.simulator.noise_model:
-            self.backend.set_options(noise_model=self.options.simulator.noise_model)
+        self._kwargs = kwargs
```
obliviateandsurrender, line 257: What is the difference between `_init_kwargs` and `_kwargs`? Do we need some validation for it, or can we just rely on the pipeline to raise sufficiently descriptive errors for unrecognized arguments?

pennylane_qiskit/qiskit_device2.py
obliviateandsurrender: Will the order of these classical registers matter, then? Will the measurement registers always remain on the first index?

pennylane_qiskit/qiskit_device2.py
austingmhuang: I'm deleting `_init_kwargs` in #547 because I don't think it's relevant anymore. In the past we needed `_init_kwargs` because we had the option to change kwargs post-device-initialization. However, after discussion with tom, this isn't really what PL devices do so we are opting to not allow users to change options after device initialization. If they want to, for example, change the optimization_level to 1 instead of 2, then they will have to redefine their device rather than write something like `device.options.optimization_level = 2`.

austingmhuang:
> Will the order of these classical registers matter, then? Will the measurement registers always remain on the first index?

Right now I don't think a User can pass in a circuit that has multiple measurement registers i.e. they can only have one classical register. So basically the order doesn't matter and the index will always be the first one because there is only 1 classical register always.

If we were to implement the ability to have multiple classical registers then I think this would be a problem, but it's not clear to me if pennylane devices support this functionality. Is it possible to specify multiple classical output registers using default.qubit?

obliviateandsurrender (approved):
Looks good to me! Great work @austingmhuang 🚀

tests/test_base_device.py
```diff
@@ -994,75 +902,20 @@ def test_estimator_with_various_multi_qubit_pauli_obs(
         sampler_execute.assert_not_called()
         estimator_execute.assert_called_once()
 
-        assert np.allclose(res, expectation, atol=0.3)  ## atol is high due to high variance
-
-    @pytest.mark.parametrize(
```
lillian542, line 999: we can't just update this test to be based on the new format? its nice to unit test that the process_estimator_job function returns the expected result

tests/test_base_device.py
```diff
@@ -994,75 +902,20 @@ def test_estimator_with_various_multi_qubit_pauli_obs(
...
-        assert isinstance(processed_result, tuple)
-        assert np.allclose(processed_result, expectation, atol=0.1)
+        assert np.allclose(res[0], expectation, atol=0.3)  ## atol is high due to high variance
 
     @pytest.mark.parametrize("num_wires", [1, 3, 5])
     @pytest.mark.parametrize("num_shots", [50, 100])
-    @pytest.mark.skip(reason="Need to replace this with using SamplerV2.")
+    @pytest.mark.skip(
+        reason="Need to replace this with using SamplerV2."
+    )  # ToDo: Do we even need this?
```
lillian542, line 911: should this still be being skipped? Shouldn't it work now? We do definitely want a unit test for `generate_samples`, so unless this is duplicated elsewhere we should keep it.

lillian542 (commented):
Left a few minor suggestions, see what you think. 

Has this been tested with hardware? I'm losing track 😅

pennylane_qiskit/qiskit_device2.py
```diff
@@ -289,7 +266,7 @@ def __init__(
             raise ValueError(f"Backend '{backend}' supports maximum {available_qubits} wires")
 
         self.reset()
-        self._update_kwargs()
+        # ToDo: process or update kwargs in some fashion here
```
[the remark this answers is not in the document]
austingmhuang: [Yep](https://app.shortcut.com/xanaduai/story/63544/remove-options-and-go-back-to-processing-kwargs), forgot to link it though...

pennylane_qiskit/qiskit_device2.py
```diff
@@ -495,36 +448,48 @@ def execute_circuits(session):
...
 
-        return (result.quasi_dists[0],)
+        for index, circ in enumerate([circuit]):
+            self._samples = self.generate_samples(index)
+            res = [
+                mp.process_samples(self._samples, wire_order=self.wires) for mp in circ.measurements
+            ]
+            single_measurement = len(circ.measurements) == 1
+            res = res[0] if single_measurement else tuple(res)
+            results.append(res)
```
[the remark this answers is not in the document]
austingmhuang: True. I just copied and pasted the old code from `backend.run()` and it worked fine so I didn't think that much deeply about it 😅 . Didn't notice that the circuit would always be a single circuit.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -495,36 +448,48 @@ def execute_circuits(session):
...
         # split into one call per measurement
         # could technically be more efficient if there are some observables where we ask
         # for expectation value and variance on the same observable, but spending time on
         # that right now feels excessive
-
-        pauli_observables = [mp_to_pauli(mp, self.num_wires) for mp in circuit.measurements]
-        result = estimator.run([qcirc] * len(pauli_observables), pauli_observables).result()
+        circ_and_obs = [
+            (compiled_circuits[i], pauli_observables[i]) for i in range(len(pauli_observables))
+        ]
```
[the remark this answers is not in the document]
austingmhuang: Yeah looking at it seems I may have looked at the wrong syntax. Fixed now though

tests/test_base_device.py
austingmhuang: This does not work right now cause we can't set a shots value yet (no processing kwargs). I would like to keep a test for `generate_samples` but it is actually kind of tricky to update this function to use the new `_execute_sampler` function.

tests/test_base_device.py
austingmhuang: This one I don't think is too challenging to update to the new format. There's no `EstimatorResult` equivalent for EstimatorV2 as far as I can tell but everything else should be updatable.

tests/test_base_device.py
austingmhuang: Well, I thought it was tricky but it's implemented in the #547 now.

tests/test_base_device.py
austingmhuang: Yeah ok I reworked this test but I didn't exactly care that much about `Type` of the data inside the result, since the focus of the test is on if process_estimator_job is working properly or not.

austingmhuang:
> Left a few minor suggestions, see what you think.
> 
Refactored the code so it's more obvious we're working with just a single circuit here. Also added back some tests.

> Has this been tested with hardware? I'm losing track 😅
No... hardware tests take 3 hours just to queue up 🙃. I think I will run hardware tests when this and #547 are merged. I can then run hardware tests while I work on the documentation and stuff like that.

PR #540, opened 2024-05-17
Remove devices that will not be supported in the new release

austingmhuang:
These devices will no longer work once we set the requirement `Qiskit>1.0`. So we remove them.

In particular we are removing all the ibmq devices (circuit runner, ibmq, sampler) and the older basicaer device. Which is 4 devices. The docs are slightly edited to pass the CI mostly, there will be a second PR cleaning up the docs and making sure everything looks right (this PR is big enough already).


austingmhuang:
[sc-63583]

doc/devices/ibmq.rst
```diff
@@ -1,84 +0,0 @@
-IBM Q Experience
-================
-
-PennyLane-Qiskit supports running PennyLane on IBM Q hardware via the ``qistkit.ibmq`` device.
-You can choose between different backends - either simulators tailor-made to emulate the real hardware,
-or the real hardware itself.
-
-Accounts and Tokens
```
austingmhuang, line 8: Need to write new docs explaining to use ibm_runtime_service instead. Separate PR!

doc/requirements.txt
```diff
@@ -7,10 +7,10 @@ pennylane==0.34
...
 pygments==2.17.2
 pygments-github-lexers==0.0.5
-qiskit==0.45.3
-qiskit-aer==0.13.3
-qiskit-ibm-runtime==0.20.0
-qiskit-ibm-provider==0.10.0
+qiskit==1.0.2
+qiskit-aer==0.14.1
+qiskit-ibm-runtime==0.23.0
+qiskit-ibm-provider==0.11.0
```
austingmhuang, line 13: I changed this because the docs would not build otherwise

requirements.txt
```diff
@@ -1,44 +1,4 @@
...
-stevedore==5.2.0
-symengine==0.11.0
-sympy==1.12
-toml==0.10.2
-urllib3==2.2.1
-websocket-client==1.7.0
+pennylane>=0.32
+qiskit
+numpy
+sympy
```
austingmhuang, line 4: I changed this because the docs would not build otherwise. Plus this was very outdated for a long time already. We can do a separate PR to fix this up to be honest.

trbromley:
> These devices will no longer work once we set the requirement `Qiskit>1.0`. So we remove them.

I have some memory of us discussing deprecations, am I wrong? :thinking: Though I agree if these devices don't work with Qiskit 1.0 then that necessitates a removal rather than a deprecation.

One thing that would be nice in the docs, e.g., [here](https://xanaduai-pennylane--540.com.readthedocs.build/projects/qiskit/en/540/#devices), would be to have a message like "Looking for device X, Y, or Z? You'll need to use the older version of this plugin and Qiskit."

austingmhuang:
> > These devices will no longer work once we set the requirement `Qiskit>1.0`. So we remove them.
> 
> I have some memory of us discussing deprecations, am I wrong? 🤔 Though I agree if these devices don't work with Qiskit 1.0 then that necessitates a removal rather than a deprecation.
> 

Basicaer already doesn't work on Qiskit >1.0 so that needs to be removed 100%. I think last time we did say that we would just deprecate the ibmq devices so I could revert the removal here instead 😅 .

Although there is an argument to be made that we should remove the ibmq devices since

1. IBMProvider has been deprecated for some time now (no telling when it'll become discontinued).
2. The default behaviour of ibmq_devices is to use the 'ibmq_qasm_simulator", which has already been discontinued.
3. These devices will 100% be broken around Oct 15, which was similar reasoning to why we're upgrading to 1.0 in the first place.
4. The functionality can be made available in qiskit.remote. 

I can easily revert the changes here and just remove basicaer only though. Let me know what you think @trbromley!

> One thing that would be nice in the docs, e.g., [here](https://xanaduai-pennylane--540.com.readthedocs.build/projects/qiskit/en/540/#devices), would be to have a message like "Looking for device X, Y, or Z? You'll need to use the older version of this plugin and Qiskit."

👍 Agree

doc/index.rst
```diff
@@ -76,9 +56,9 @@ follows:
 
     dev = qml.device('qiskit.aer', wires=2, backend='unitary_simulator')
 
-PennyLane chooses the ``qasm_simulator`` as the default backend if no backend is specified.
-For more details on the ``qasm_simulator``, including available backend options, see
-`Qiskit Qasm Simulator documentation <https://qiskit.org/ecosystem/aer/stubs/qiskit_aer.QasmSimulator.html>`_.
+PennyLane chooses the ``aer_simulator`` as the default backend if no backend is specified.
+For more details on the ``aer_simulator``, including available backend options, see
+`Qiskit Qasm Simulator documentation <https://qiskit.github.io/qiskit-aer/stubs/qiskit_aer.AerSimulator.html#qiskit_aer.AerSimulator.run>`_.
```
obliviateandsurrender, line 61: We can add the warning here, I guess, regarding outdated-missing devices!

pennylane_qiskit/basic_aer.py
```diff
@@ -97,15 +47,4 @@ class BasicSimulatorDevice(QiskitDevice):
     )
 
     def __init__(self, wires, shots=1024, backend="basic_simulator", **kwargs):
-
-        min_version = Version("1.0.0")
```
obliviateandsurrender, line 101: If we are removing this how do we let users know we no longer support versions prior to `1.0.0` during runtime?

requirements.txt
obliviateandsurrender, line 4:
```suggestion
sympy

```
Do we need to be pinning any of this?

obliviateandsurrender (commented):
Good work @austingmhuang! Just leaving a couple of queries.

pennylane_qiskit/basic_aer.py
austingmhuang: You will get an import error from now on instead from Qiskit's end since I deleted
`if Version(qiskit.__version__) >= Version("1.0.0"):`
   ` from qiskit.providers.basic_provider import BasicProvider`

doc/index.rst
austingmhuang: Agree. The documentation in general needs a lot of updating, so much so that it would probably be a pretty big PR on its own in my opinion.

requirements.txt
austingmhuang: I don't think so... the file used for the ci-tests are not pinned either so I assume that it shouldn't be needed.

PR #545, opened 2024-05-22
Remove devices that will not be supported in the new release

austingmhuang:
[Accidentally merged this PR (somehow two reviews is not necessary?) unintentionally.](https://github.com/PennyLaneAI/pennylane-qiskit/pull/540).

We are removing all the ibmq devices (circuit runner, ibmq, sampler) and the older basicaer device. Which is 4 devices. The docs are slightly edited to pass the CI mostly, there will be a second PR cleaning up the docs and making sure everything looks right (this PR is big enough already).


austingmhuang:
[sc-63583]

PR #546, opened 2024-05-22
Remove basicaer

austingmhuang:
[Accidentally merged this PR (somehow two reviews is not necessary?) unintentionally.](https://github.com/PennyLaneAI/pennylane-qiskit/pull/540).

In this PR, we are just removing basicaer because it will break when we upgrade to 1.0.  Basicaer has been removed as of Qiskit 1.0 and its functionality has been moved to Basicsim.

I separated the ibmq devices from this pr to reduce the size of this PR while we get confirmation regarding deprecation vs removal for the ibmq devices. In this PR, the docs are slightly edited to pass the CI, but are not the main focus of this PR. We will have a separate PR for the docs.


doc/index.rst
```diff
@@ -76,9 +72,9 @@ follows:
 
     dev = qml.device('qiskit.aer', wires=2, backend='unitary_simulator')
 
-PennyLane chooses the ``qasm_simulator`` as the default backend if no backend is specified.
-For more details on the ``qasm_simulator``, including available backend options, see
-`Qiskit Qasm Simulator documentation <https://qiskit.org/ecosystem/aer/stubs/qiskit_aer.QasmSimulator.html>`_.
+PennyLane chooses the ``aer_simulator`` as the default backend if no backend is specified.
+For more details on the ``aer_simulator``, including available backend options, see
+`Qiskit Aer Simulator documentation <https://qiskit.github.io/qiskit-aer/stubs/qiskit_aer.AerSimulator.html#qiskit_aer.AerSimulator.run>`_.
```
austingmhuang, line 77: Qasm simulator has been removed already. However, for the ibmq devices, the default behaviour cannot actually use the local aer simulators. Right now the default behaviour for the ibmq devices is already broken.

austingmhuang:
[sc-63583]

doc/requirements.txt
```diff
@@ -7,10 +7,10 @@ pennylane==0.34
...
 pygments==2.17.2
 pygments-github-lexers==0.0.5
-qiskit==0.45.3
-qiskit-aer==0.13.3
-qiskit-ibm-runtime==0.20.0
-qiskit-ibm-provider==0.10.0
+qiskit==1.0.2
+qiskit-aer==0.14.1
+qiskit-ibm-runtime==0.23.0
+qiskit-ibm-provider==0.11.0
```
austingmhuang, line 13: Changed this to build docs.

doc/index.rst
lillian542: Are we updating the behaviour here? Have we already updated it in a separate PR? Or does the deleted text accurately describe what the devices currently do?

austingmhuang: We need to update the behaviour for the ibmq devices since the ibmq devices use the ibmq_qasm_simulator as their default backend. For qiskit.aer and qiskit.remote the behaviour was already [updated in new_device_feature_branch](https://github.com/PennyLaneAI/pennylane-qiskit/pull/521)

austingmhuang: Oh wait, I just realized this is for qiskit.aer devices. Well, regardless, this was updated to use AerSimulator in that PR so this would be reflecting that change.

austingmhuang: Yeah just double checked. The default behaviour of qiskit.aer is to use AerSimulator as its backend. So... I suppose this is an accidental but correct change of the documentation to reflect reality.

austingmhuang: Ok. After some further investigation, what's going on here is that `qiskit.basicaer` used to default to `QasmSimulator` but `qiskit.aer` has always defaulted to `Aer_Simulator`. The Index.rst here was describing the behaviour of the `basicaer` device. With the removal of `basicaer`, we are now describing the behaviour of the `qiskit.aer`, which defaults to `Aer_Simulator`. World is round.

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@
 * The test suite no longer runs for Qiskit versions below 0.46.
   [(#536)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/536)
 
+* The basic aer device has been removed.
```
lillian542, line 16:
```suggestion
* The ``qiskit.basicaer`` device has been removed because it is not supported for versions of Qiskit above 0.46.
```

doc/index.rst
lillian542: We'll worry about the details when the docs update comes later anyway, I suppose, seems good for now

tests/conftest.py
```diff
@@ -41,7 +41,7 @@
 
 
 if Version(qiskit.__version__) < Version("1.0.0"):
-    test_devices = [AerDevice, BasicAerDevice]
+    test_devices = [AerDevice]
     hw_backends = ["qasm_simulator", "aer_simulator"]
     state_backends = [
```
lillian542, line 46: I think all of the `if Version` checks for qiskit in the `conftest.py` file can just be removed now, right? Since the tests shouldn't be running with `Version(qiskit.__version__) < Version("1.0.0")` anymore.

lillian542: Maybe more relevant [in 536](https://github.com/PennyLaneAI/pennylane-qiskit/pull/536) 🤷‍♀️ But just tidying up either way, since it should always use the `else` block once 536 is merged

lillian542 (approved):
Made a couple of minor suggestions, but looks good to me!

tests/conftest.py
austingmhuang: Yeah all the` if Version` checks should be removed in general; there are quite a few littered throughout the tests file as well that I didn't remove to keep this PR lean. 

I can tidy this up in this PR though since it's within the scope of removing the basicaerdevice though.

obliviateandsurrender (approved):

austingmhuang:
[sc-63833]

PR #547, opened 2024-05-24
Process kwargs

austingmhuang:


pennylane_qiskit/qiskit_device2.py
```diff
@@ -372,8 +369,34 @@ def preprocess(
...
+
+        if "default_shots" in self._kwargs:
+            warnings.warn(
+                f"default_shots was found as a keyword argument, but it is not supported by {self.name}"
+                "Please use the `shots` keyword argument instead. The number of shots "
+                f"{shots} will be used instead."
+            )
+        self._kwargs["default_shots"] = shots
+
+    def get_transpile_args(self):
```
austingmhuang, line 399: Changed the signature of this function because we want to `pop` the kwargs that are going to be passed to the transpiler. Otherwise when these kwargs reach the primitive, it'll raise a `ValidationError`.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -447,12 +472,15 @@ def execute_circuits(session):
...
-
         qcirc = [circuit_to_qiskit(circuit, self.num_wires, diagonalize=True, measure=True)]
         sampler = Sampler(session=session)
         compiled_circuits = self.compile_circuits(qcirc)
+        sampler.options.update(**self._kwargs)
 
-        result = sampler.run(compiled_circuits).result()[0]
+        result = sampler.run(
+            compiled_circuits,
+            shots=circuit.shots.total_shots if circuit.shots.total_shots else None,
```
austingmhuang, line 482: Shots defined in `circuit` take precedence over shots defined on device initialization.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -503,10 +536,8 @@ def _process_estimator_job(measurements, job_result):
 
         expvals = [res.data.evs.item() for res in job_result]
         variances = [
-            res.data.stds.item() ** 2 * 4096 for res in job_result
-        ]  # this 4096 is the # of shots Qiskit uses by default. It is hard-coded here.
-        # ToDo: Track the # of shots and use that instead of hard-coding
-        # to calculate the variance.
+            (res.data.stds.item() / res.metadata["target_precision"]) ** 2 for res in job_result
```
austingmhuang, line 539: We can calculate the variance by using the std and the precision.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -482,14 +510,19 @@ def _execute_estimator(self, circuit, session):
...
         # could technically be more efficient if there are some observables where we ask
         # for expectation value and variance on the same observable, but spending time on
         # that right now feels excessive
         circ_and_obs = [
             (compiled_circuits[i], pauli_observables[i]) for i in range(len(pauli_observables))
         ]
-        result = estimator.run(circ_and_obs).result()
+        result = estimator.run(
+            circ_and_obs,
+            precision=np.sqrt(1 / circuit.shots.total_shots) if circuit.shots else None,
```
austingmhuang, line 523: Essentially, if you set some kind of precision on device initialization, that will get overwritten if you set some number of shots when running the circuit. Estimator does not take in shots at runtime (only on initialization), but since PL does not take in some `precision` value it doesn't make sense to support `precision` as an option. Therefore, this code just converts a shot value to a precision value, so that the estimator can run the shot value that we set :)

tests/test_base_device.py
```diff
@@ -180,38 +179,6 @@ def test_no_shots_warns_and_defaults(self):
 
         assert dev.shots.total_shots == 1024
 
-    @pytest.mark.skip(reason="Options handling not decided on yet")
-    def test_kwargs_on_initialization(self, mocker):
```
austingmhuang, line 184: This is gone since we don't have an Options() object anymore. It is replaced by a different test that handles when options are set via options={"something": 1}.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -372,8 +369,34 @@ def preprocess(
...
+        if "noise_model" in self._kwargs:
+            noise_model = self._kwargs.pop("noise_model")
+            self.backend.set_options(noise_model=noise_model)
+
+        if "options" in self._kwargs:
+            for key, val in self._kwargs.pop("options").items():
+                if key in self._kwargs:
+                    warnings.warn(
+                        "An overlap between what was passed in via options and what was passed in via kwargs was found."
+                        f"The value set in options {key}: {val} will be used."
```
obliviateandsurrender, line 385:
```suggestion
                        f"The value set in options {key}={val} will be used."
```

pennylane_qiskit/qiskit_device2.py
```diff
@@ -372,8 +369,34 @@ def preprocess(
...
+                    )
+                self._kwargs[key] = val
+
+        shots = self._kwargs.pop("shots")
+
+        if "default_shots" in self._kwargs:
+            warnings.warn(
+                f"default_shots was found as a keyword argument, but it is not supported by {self.name}"
```
obliviateandsurrender, line 393:
```suggestion
                f"default_shots was found in the keyword arguments, but it is not supported by {self.name}"
```

pennylane_qiskit/qiskit_device2.py
obliviateandsurrender: Nice!

pennylane_qiskit/qiskit_device2.py
obliviateandsurrender: How do we come up with this precision value? 🤔

austingmhuang:
[sc-63544]

tests/test_base_device.py
```diff
@@ -991,31 +928,6 @@ def test_generate_samples(self, num_wires, num_shots):
         # nothing else is in samples
         assert [s for s in samples if not s in np.array([exp_res0, exp_res1])] == []
 
-    @pytest.mark.skip(
-        reason="Tracking shot information will be addressed in the PR about options handling"
-    )
-    def test_tape_shots_used_for_estimator(self, mocker):
```
austingmhuang, line 997: This has been replaced with the test at Line 835. I don't know why it's misaligned, maybe due to fixing merge conflicts.

pennylane_qiskit/qiskit_device2.py
lillian542: What happens if you run this on two consecutive circuits? Does the second one miss the `transpile` arguments because they've been removed?

Maybe this should be called once at initialization and store all the arguments as a `self._transpile_args` property instead of calling this each time we compile.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -494,37 +472,54 @@ def execute_circuits(session):
...
-        return (result.quasi_dists[0],)
+        self._samples = self.generate_samples(0)
+        res = [
+            mp.process_samples(self._samples, wire_order=self.wires) for mp in circuit.measurements
+        ]
+        single_measurement = len(circuit.measurements) == 1
+        res = res[0] if single_measurement else tuple(res)
+        results.append(res)
+
+        return tuple(results)
```
lillian542, line 503: This looks like it adds an unexpected dimension to the results for the non-`single_measurement case`. I think we want to go straight to `tuple(res)`.

If so, that makes we think maybe we don't have tests covering the case where `len(res) != 1`? So maybe we should also add a test that would have noticed this.

lillian542 (commented):
Looking good! There's a couple of things that I think need a little adjustment.

tests/test_base_device.py
```diff
@@ -222,17 +189,13 @@ def test_backend_wire_validation(self, backend):
         with pytest.raises(ValueError, match="supports maximum"):
             QiskitDevice2(wires=500, backend=backend)
 
-    @pytest.mark.skip(reason="Options handling not decided on yet")
     def test_setting_simulator_noise_model(self):
```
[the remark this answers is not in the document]
lillian542: not relevant, never mind 😅

austingmhuang:
>What happens if you run this on two consecutive circuits? Does the second one miss the transpile arguments because they've been removed?

Hmm yes it seems like the second one should be missing the transpile arguments 😰 ... I think the solution would be to parse through kwargs -> separate and save them into transpile_args and runtime_args on the device in the pre-process step instead.

pennylane_qiskit/qiskit_device2.py
austingmhuang: We don't have tests covering the case where `len(res) != 1` 😅 and you're right that it does in fact add an additional layer of dimensionality. Interestingly enough though this will pass any test where `len(res) != 1` regardless, because of the `reorder_fn(res)` function. Basically we would pass in a result that looks like ((x, y),) rather than (x, y)... but then 

`return result[0] if len(result) == 1 else result` will interpret that as result[0] and we're right back to (x, y) -> so the extra layer of dimensionality would be cancelled out. Tricky bug and I'm actually thinking about reworking this whole post-processing the result with `reorder_fn` thing cause 1. it's confusing, and 2. it messes with the Tracker. But I fixed the bug for now and I'll add a few more test cases.

austingmhuang:
> > What happens if you run this on two consecutive circuits? Does the second one miss the transpile arguments because they've been removed?
> 
> Hmm yes it seems like the second one should be missing the transpile arguments 😰 ... I think the solution would be to parse through kwargs -> separate and save them into transpile_args and runtime_args on the device in the pre-process step instead.
> 
> I would really like to write a good test to make sure that these transpile args are actually being used when compiling it since I would have caught this with a better test probably but it seems like the best we can do is a unit test that checks that dev has a `kwargs` and a `transpile_args` and that they're separated and saved clearly.

This is done now. I also edited the docstrings and added a comment to clarify that process_kwargs will give the device a `transpile_args` attribute since it's not that obvious (There might be a way to make this more obvious via the code than with a comment/docstrings; idk tho)

pennylane_qiskit/qiskit_device2.py
```diff
@@ -203,10 +203,10 @@ class QiskitDevice2(Device):
         compile_backend (Union[Backend, None]): the backend to be used for compiling the circuit that will be
             sent to the backend device, to be set if the backend desired for compliation differs from the
             backend used for execution. Defaults to ``None``, which means the primary backend will be used.
-        **kwargs: transpilation and runtime kwargs to be used for measurements without Qiskit Primitives.
-            If any values are defined both in ``options`` and in the remaining ``kwargs``, the value
-            provided in ``options`` will take precedence. These kwargs will be ignored for all Primitive-based
-            measurements on the device.
+        **kwargs: transpilation and runtime kwargs to be used for measurements with Primitives. If `options` is
```
obliviateandsurrender, line 206:
```suggestion
        **kwargs: transpilation and runtime keyword arguments to be used for measurements with Primitives. If an `options` dictionary is
```

pennylane_qiskit/qiskit_device2.py
```diff
@@ -203,10 +203,10 @@ class QiskitDevice2(Device):
...
+            defined amongst the kwargs, and there are settings that overlap with those in kwargs, the settings
+            in `options` will take precedence over kwargs. Kwargs accepted by both the transpiler
+            and at runtime (e.g. optimization_level) will be passed to the transpiler rather than to the Primitive.
```
obliviateandsurrender, line 209:
```suggestion
            and at runtime (e.g. ``optimization_level``) will be passed to the transpiler rather than to the Primitive.
```

pennylane_qiskit/qiskit_device2.py
```diff
@@ -203,10 +203,10 @@ class QiskitDevice2(Device):
         compile_backend (Union[Backend, None]): the backend to be used for compiling the circuit that will be
             sent to the backend device, to be set if the backend desired for compliation differs from the
             backend used for execution. Defaults to ``None``, which means the primary backend will be used.
-        **kwargs: transpilation and runtime kwargs to be used for measurements without Qiskit Primitives.
-            If any values are defined both in ``options`` and in the remaining ``kwargs``, the value
-            provided in ``options`` will take precedence. These kwargs will be ignored for all Primitive-based
-            measurements on the device.
+        **kwargs: transpilation and runtime kwargs to be used for measurements with Primitives. If `options` is
+            defined amongst the kwargs, and there are settings that overlap with those in kwargs, the settings
+            in `options` will take precedence over kwargs. Kwargs accepted by both the transpiler
```
obliviateandsurrender, line 208:
```suggestion
            in `options` will take precedence over kwargs. Keyword arguments accepted by both the transpiler
```

pennylane_qiskit/qiskit_device2.py
```diff
@@ -266,7 +263,7 @@ def __init__(
             raise ValueError(f"Backend '{backend}' supports maximum {available_qubits} wires")
 
         self.reset()
-        # ToDo: process or update kwargs in some fashion here
+        self._process_kwargs()  # processes kwargs and separates transpilation arguments to dev._transpile_args
```
obliviateandsurrender, line 266: I would prefer this to return the `kwargs` and `transpiler_kwargs` instead of them changing it outside the `__init__` context. 
```suggestion
       self._transpile_args, self._kwargs = self._process_kwargs(kwargs)  # processes kwargs and separates transpilation arguments to dev._transpile_args
```

pennylane_qiskit/qiskit_device2.py
```diff
@@ -372,8 +369,37 @@ def preprocess(
...
+            warnings.warn(
+                f"default_shots was found in the keyword arguments, but it is not supported by {self.name}"
+                "Please use the `shots` keyword argument instead. The number of shots "
+                f"{shots} will be used instead."
+            )
+        self._kwargs["default_shots"] = shots
+
+        self._transpile_args = self.get_transpile_args()
+
+    def get_transpile_args(self):
```
obliviateandsurrender, line 402: Same suggestion for this - 
```suggestion
    def get_transpile_args(self, kwargs): --> _transpile_args
```
That it processes the `kwargs` given to it and not the ones contained in the object directly.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -500,11 +535,7 @@ def _process_estimator_job(measurements, job_result):
         return the requested results from the Estimator executions."""
 
         expvals = job_result[0].data.evs
-        variances = (
-            job_result[0].data.stds ** 2 * 4096
-        )  # this 4096 is the # of shots Qiskit uses by default. It is hard-coded here.
-        # ToDo: Track the # of shots and use that instead of hard-coding
-        # to calculate the variance.
+        variances = (job_result[0].data.stds / job_result[0].metadata["target_precision"]) ** 2
```
obliviateandsurrender, line 538: Nice!

obliviateandsurrender (commented):
Few more suggestions!

pennylane_qiskit/qiskit_device2.py
austingmhuang: Makes sense. I think that's cleaner as well.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -250,11 +250,7 @@ def __init__(
         self._service = getattr(backend, "_service", None)
         self._session = session
 
-        # initial kwargs are saved and referenced every time the kwargs used for transpilation and execution
-        self._init_kwargs = kwargs
-        # _kwargs are used instead of the Options for performing raw sample based measurements (using old Qiskit API)
-        # the _kwargs are a combination of information from Options and _init_kwargs
-        self._kwargs = kwargs
+        kwargs["shots"] = shots
```
obliviateandsurrender, line 253: What happens if the `options` have `shots`? They won't be able to override it directly? 🤔

obliviateandsurrender (approved):
Good work @austingmhuang, let's :shipit:!

pennylane_qiskit/qiskit_device2.py
austingmhuang: If you pass in options={"shots": 30}, you get a warning but options will override it. 

```        
  if "options" in kwargs:
      for key, val in kwargs.pop("options").items():
          if key in kwargs:
              warnings.warn(
                  "An overlap between what was passed in via options and what was passed in via kwargs was found."
                  f"The value set in options {key}={val} will be used."
              )
          kwargs[key] = val
     ```

**Comment by lillian542** [comment]
>I would really like to write a good test to make sure that these transpile args are actually being used when compiling it since I would have caught this with a better test probably but it seems like the best we can do is a unit test that checks that dev has a kwargs and a transpile_args and that they're separated and saved clearly.

It's also tough because we rarely test what happens if we do the same thing *twice* to make sure it goes right the _second_ time 😅 And we often create a new device for each test. I think this kind of thing will generally slip past standard tests.

File: pennylane_qiskit/qiskit_device2.py (hunk 9 above)
Comment (lillian542) [reply pennylane_qiskit/qiskit_device2.py:509]: This still seems off to me. It looks like for the len(res) != 1 case, we've just gone from :

```
>>> result = []
>>> res = tuple([1, 2])
>>> result.append(res)
>>> tuple(result)
((1, 2),)
```

to 

```
>>> result = []
>>> res = [1, 2]
>>> result.append(res)
>>> tuple(result)
([1, 2],)
```

when I think what we generally expect is:

```
>>> result = []
>>> res = [1, 2]
>>> tuple(res)
(1, 2)
```
But maybe I'm mixing things up here

Comment (lillian542) [reply pennylane_qiskit/qiskit_device2.py:509]: I guess as long as it comes out as expected in the end, it doesn't matter 😅

Comment (austingmhuang) [reply pennylane_qiskit/qiskit_device2.py]: Essentially when the measurementProcess is something like expval() it doesn't work as expected but when it's something like samples() then it will behave more normally.

File: tests/test_base_device.py [hunk 18]
```diff
@@ -1158,3 +1077,42 @@ def qiskit_circuit(x):
...
+                qml.sample(qml.Y(0)),
+                qml.expval(qml.X(1)),
+                qml.var(qml.Y(0)),
+                qml.counts(),
+            )
+
+        res = circuit(np.pi / 2)
+        qiskit_res = qiskit_circuit(np.pi / 2)
+
+        print(res, qiskit_res)
```
lillian542, line 1114:
```suggestion
```

lillian542 (approved):
Looks good to me 🚀

tests/test_base_device.py
austingmhuang: 😅

obliviateandsurrender:
It probably needs to merge changes from the `master` onto the feature branch for the CI to pass.

PR #549, opened 2024-05-30
Example added `BasicSim` docs that shows default `shots`

isaacdevlugt:
In Qiskit, when you use this device / backend you’re hard-corded at `shots=1024`. I assume Qiskit users are aware of this. However, when I used this device in PennyLane, I ran this circuit not expecting it to work because I didn’t specify shots. 

```
pl_qfunc = qml.from_qiskit(qc, measurements=[qml.classical_shadow(wires=range(n))])
pl_circuit = qml.QNode(pl_qfunc, device=qml.device("qiskit.basicsim", wires=n))
pl_circuit()
```

I had to go digging into the source code to figure out what was going on. As a PennyLane user, I found this confusing (does that matter? lol). Anyway, a simple addition to the docs seems fine to me.

### DIFF
--- a/doc/devices/basicsim.rst
+++ b/doc/devices/basicsim.rst
@@ -14,6 +14,21 @@ that can be accessed in this plugin through:
 
 This device uses the Qiskit ``BasicSimulator`` backend from the
 `basic_provider <https://docs.quantum.ibm.com/api/qiskit/providers_basic_provider>`_ module in Qiskit.
+In Qiskit, ``BasicSimulator`` uses ``shots=1024`` by default. As such, the same convention has been applied
+to `'qiskit.basicsim'` in PennyLane. 
+
+.. code-block:: python
+
+    @qml.qnode(dev)
+    def circuit():
+        qml.Hadamard(0)
+        qml.Hadamard(1)
+        return qml.count(wires=1)
+
+.. code-block:: pycon
+    
+    >>> circuit()
+    {'0': tensor(520, requires_grad=True), '1': tensor(504, requires_grad=True)} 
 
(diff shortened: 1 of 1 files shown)


isaacdevlugt:
Screenshot 2024-05-30 at 11 27 48 AM

isaacdevlugt:
Hey @austingmhuang! I mean the latter. 1024 is hardcoded as the _default_ number (can be overridden). 

- the plugin: https://github.com/PennyLaneAI/pennylane-qiskit/blob/eb865f7920c59d5580771517ea1e65ad97cdc56c/pennylane_qiskit/basic_aer.py#L99
- qiskit: https://github.com/Qiskit/qiskit/blob/df379876ba10d6f490a96723b6dbbf723ec45d7a/qiskit/providers/basic_provider/basic_simulator.py#L260

trbromley:
Why don't we update `BasicSim` to make `shots` a required argument?

tests/test_qiskit_device.py
```diff
@@ -167,22 +173,46 @@ def test_transpilation_option_update(self, device, wires, device_options, transp
...
+        ),
+    ]
+
+    @pytest.mark.parametrize(
+        "device_backend_msg",
+        testcases_warning_raised_for_hardware_backend_analytic_expval,
+    )
+    def test_warning_raised_for_hardware_backend_analytic_expval(
+        self, device_backend_msg, recorder
+    ):
```
isaacdevlugt, line 204: Not sure if this is best practice since it's kinda messy. Just didn't want to have (nearly) identical tests for two devices 😅

PR #550, opened 2024-05-31
Remove ibmq devices

austingmhuang:
After discussion with @trbromley, it was decided to remove these devices. This PR simply deletes all the devices and where they are referenced in the docs/code. No other changes were needed.


austingmhuang:
[sc-63583]

obliviateandsurrender (approved):
It looks good to me. Few points -
1. We are removing them without any deprecation cycle? 
2. We should check whether our demos depend on these devices.
3. Is there any replacement for these devices from Qiskit's side? Or is it just the `remote` devices?

austingmhuang:
> It looks good to me. Few points -
> 
> 1. We are removing them without any deprecation cycle?

Yes. There are a lot of reasons as to why but basically yes.

> 2. We should check whether our demos depend on these devices.

This is a good point.

> 3. Is there any replacement for these devices from Qiskit's side? Or is it just the `remote` devices?

Just the remote devices. Sampler/Circuit-runner (raw samples) are replaced with SamplerV2, which is what we are going to use going forward as well.

lillian542 (approved):
Is there a story for updating the demos?

austingmhuang:
> Is there a story for updating the demos?

No, not yet. I'll create a story for that.

PR #551, opened 2024-06-03
Qiskit session

austingmhuang:
**Context:** 
Session options like `max_time` were not supported. If the Users wanted to use `qiskit_session`, they wouldn't be able to set any settings for that session.

**What this PR does:**
This PR allows users to set options that the Session initializer accepts. This is to make this relatively future-proof, where if Qiskit decides that the Session should accept other keyword arguments then we would be able to simply pass that along. We do not allow users to modify two specific settings when using `qiskit_session`, namely the `backend` and the `service`. This is because we force the user to specify a backend when we initialize the device in the first place, by proxy, what service. Basically, it doesn't make sense to let them modify this on session initialization. We do want `max_time` and potentially other keyword arguments in the future to be able to be modified when initializing the `qiskit_session` though.

Any session that the user set on device initialization is overwritten by this context manager (unless the context manager is trying to do something strange, like change the backend/service). The examples/docstrings try to make this as clear as possible.


austingmhuang:
[sc-54393]

pennylane_qiskit/qiskit_device2.py
```diff
@@ -98,7 +98,31 @@ def circuit(x):
...
+        elif existing_session and (
+            hasattr(existing_session, "_" + k) or hasattr(existing_session, k)
+        ):
+            warnings.warn(
+                f"`{k}` was set in the Session passed to the device. Using `{k}` '{v}' set in `qiskit_session`.",
+                UserWarning,
+            )
+            session_options[k] = v
+        else:
+            session_options[k] = v
```
austingmhuang, line 123: If Qiskit releases new Session options, this code would still capture those options and raise appropriate warnings if overlapped (unless they make one of the session attributes have extremely weird naming convention)

pennylane_qiskit/qiskit_device2.py
```diff
@@ -98,7 +98,31 @@ def circuit(x):
...
+        # Need "_" since `max_time` attribute on Session is `_max_time`
+        # When there is overlap between the Session options on the device and the session options
+        # passed in via qiskit_session, we prefer the ones passed in via qiskit_session
+        elif existing_session and (
+            hasattr(existing_session, "_" + k) or hasattr(existing_session, k)
+        ):
+            warnings.warn(
+                f"`{k}` was set in the Session passed to the device. Using `{k}` '{v}' set in `qiskit_session`.",
+                UserWarning,
+            )
```
lillian542, line 120: To me this feels like it implies that if I *don't* pass a value for `max_time`, and my initial session did have a value for `max_time`, then my initial value will be used, since I'm notified that its being overwritten - as if its stored somewhere and will remain the same until I change it. But that's not true, right? The default value will just be used?

austingmhuang: In the case where:
1. There is an initial session with some value for `max_time` passed to the device.
2. And then you run `qiskit_session(dev)` without setting `max_time`. 

Would result in
1. Don't get any warnings
2. But the session will use `None` for `max_time`, which Qiskit in turn interprets as some default value.
3. When you close the session and run the circuit again it will use the max_time stored on the device.

I can push a fix to use the `max_time` set in the initial session (i.e. don't override with `None`). Alternatively, we could raise a warning letting them know this is going to be ran with whatever Qiskit set as the default...

pennylane_qiskit/qiskit_device2.py
```diff
@@ -55,7 +55,7 @@
 
 # pylint: disable=protected-access
 @contextmanager
-def qiskit_session(device):
+def qiskit_session(device, **kwargs):
```
obliviateandsurrender, line 58: The docstring must be updated to account for the ``kwargs``. Also, do we mention allowed arguments for these keyword arguments? If they are too much, maybe you could like qiskit's docs for the users.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -98,7 +98,34 @@ def circuit(x):
...
+
+    for k, v in kwargs.items():
+        # Options like service and backend should be tied to the settings set on device
+        if k in session_options and k != "max_time":
+            warnings.warn(f"Using '{k}' set in device, {getattr(device, k)}", UserWarning)
+
+        # Need "_" since `max_time` attribute on Session is `_max_time`
+        # When there is overlap between the Session options on the device and the session options
+        # passed in via qiskit_session, we prefer the ones passed in via qiskit_session
+        elif existing_session and getattr(existing_session, "_" + k):
```
obliviateandsurrender, line 119: Do we need to worry about setting a default for the `getattr` here?

pennylane_qiskit/qiskit_device2.py
```diff
@@ -98,7 +98,34 @@ def circuit(x):
...
+            warnings.warn(
+                f"`{k}` was also set in the Session passed to the device. Using `{k}` '{v}' set in `qiskit_session`.",
+                UserWarning,
+            )
+            session_options[k] = v
```
obliviateandsurrender, line 124: I think we can combine the `elif` and `else` and use the `elif` case just to raise the warning.

pennylane_qiskit/qiskit_device2.py
lillian542: It's not so much that I would expect it to use the `max_time` from the initial session. It's more that I *wouldn't* expect that, but the warning to me implies that that's what happens, which I think I would find confusing.

I see that we want to get `service` and `backend` from the device, so raising a warning there if they try to change it I understand. But if I set `max_time` the first time and the second time to separate values, I think those are two independent things - I would find a warning for that, as if something undefined or unexpected is happening, confusing.

What if we just assume that all sessions are related to the device, and must have a `backend` and `service` from the device, but they aren't related to each other and so we don't compare them at all?

lillian542: > When there are overlapping settings in the Session passed in on device initialization and the qiskit_session, we will raise a warning letting them know this is strange, and explain what exactly happened.

Basically this. Imagining I'm a user - I'm surprised that it would be strange to use a context manager to temporarily use a session with different settings, since that seems to be what the context manager is provided for. So when a warning is raised to let me know it's strange, I feel confused about what I did wrong, and worried I've misunderstood something.

austingmhuang: Ah, I see what you mean now. Yes that is a bit of a strange warning now that I think about it. I will remove the warning for `max_time` but maybe it is useful to keep it in for the `service` and `backend`?

pennylane_qiskit/qiskit_device2.py
austingmhuang: I will update the docstring and also add the qiskit docs for users for reference.

pennylane_qiskit/qiskit_device2.py
austingmhuang: I don't think so. `existing_session`, if it exists, should always be some `Session()` which has all those attributes. I suppose if the User themselves wrote their own version of `Session` (e.g. our mock) this could be a problem, but I don't see why a User would want to do that...? I suppose if Qiskit significantly changes the attributes of Session or adds an attribute without an "_" in front of it this would be a problem. I guess it wouldn't hurt to set a default :shrug: but not sure how much value it provides.

pennylane_qiskit/qiskit_device2.py
lillian542: >maybe it is useful to keep it in for the service and backend?

Agreed! For `service` and `backend` the user is passing something unexpected that we will overwrite, so a warning seems good to me

pennylane_qiskit/qiskit_device2.py
```diff
@@ -55,14 +55,18 @@
...
     on the device while the context manager is active. Using the context manager
     will ensure the Session closes properly and is removed from the device after
     completing the tasks.
 
     Args:
         device (QiskitDevice2): the device that will create remote tasks using the session
+        **kwargs: session keyword arguments to be used for settings for the Session. At the
+        time of writing, the only relevant keyword argument is "max_time", which lets you
+        set the maximum amount of time the sessin is open. For the most up to date information,
+        please refer to <https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.Session>
```
obliviateandsurrender, line 69: I think you add the reference as a hyperlink instead of complete bare link as text.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -55,14 +55,18 @@
...
     on the device while the context manager is active. Using the context manager
     will ensure the Session closes properly and is removed from the device after
     completing the tasks.
 
     Args:
         device (QiskitDevice2): the device that will create remote tasks using the session
+        **kwargs: session keyword arguments to be used for settings for the Session. At the
+        time of writing, the only relevant keyword argument is "max_time", which lets you
+        set the maximum amount of time the sessin is open. For the most up to date information,
+        please refer to `the Qiskit Session documentation <https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.Session>`_.
```
obliviateandsurrender, line 69:
```suggestion
        please refer to the Qiskit Session `documentation <https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.Session>`_.
```

tests/test_base_device.py
```diff
@@ -232,6 +233,94 @@ def test_using_session_context(self, mock_session, initial_session):
...
+
+        with qiskit_session(dev, max_time=30) as session:
+            assert dev._session == session
+            assert dev._session is not None
+            assert dev._session._max_time == 30
+
+        assert dev._session is None
+
+    def test_error_when_passing_unexpected_kwarg(self):
+        """Test that we accept any keyword argument that the User wants to supply so that if
```
obliviateandsurrender, line 250:
```suggestion
        """Test that we accept any keyword argument that the user wants to supply so that if
```

obliviateandsurrender (approved):
Thanks @austingmhuang!

pennylane_qiskit/qiskit_device2.py
```diff
@@ -98,7 +103,27 @@ def circuit(x):
     """
     # Code to acquire session:
     existing_session = device._session
-    session = Session(backend=device.backend)
+
+    # When an existing session exists, we want to use its settings unless overwritten
+    # by settings in the qiskit_session
```
lillian542, line 108: I still consider the behaviour quite unintuitive. If we want to go for this, can we include information in the docstring about what will happen if the device that is passed already has a session on it?

austingmhuang: No I think it is kind of unintuitive to do it this way as well.

The best way is probably to just make the docstring more clear that if you initialize `qiskit_session` you will not be using your existing session set in the device.

The logic is that I think the existing session set in device should only have value in a very specific case: batched circuits. In all other cases, the session is just necessary to access the primitives but doesn't necessarily _do_ anything. 

Therefore if a user has gone out of their way to use `qiskit_session`, they probably want to do some kind of like VQE or QAOA or whatever iterative thing, so they probably want those settings. I'll make it clearer that a User **doesn't** accidentally think they should set the Session in the device.

tests/test_base_device.py
```diff
@@ -232,6 +233,94 @@ def test_using_session_context(self, mock_session, initial_session):
...
+            assert dev._session._max_time != initial_session._max_time
+
+        assert dev._session == initial_session
+        assert dev._session._max_time == initial_session._max_time
+
+    def test_warnings_when_overriding_session_context_options(self, recorder):
+        """Test that warnings are raised when there are is an overlap between the Session options
+        on the device and the session options passed in via qiskit_session. Also ensures that the
+        session options passed in from the qiskit_session take precedence except for `backend` or
+        `service`"""
```
lillian542, line 286: docstring could use clarification following updates in the PR

lillian542 (approved):
Just a couple of minor suggestions!

PR #552, opened 2024-06-04
Docstrings for converter functions and new qiskit device

austingmhuang:
With all the changes to the device and the update to PrimitivesV2, a lot of the docstrings need to be updated. This PR updates the docstrings to improve readability and consistency.


austingmhuang:
[sc-54385]

pennylane_qiskit/qiskit_device2.py
```diff
@@ -208,6 +208,61 @@ class QiskitDevice2(Device):
             with those in kwargs, the settings in `options` will take precedence over kwargs. Keyword
             arguments accepted by both the transpiler and at runtime (e.g. ``optimization_level``)
             will be passed to the transpiler rather than to the Primitive.
+
+    **Example:**
```
obliviateandsurrender, line 212: I don't see this `class` (and hence this code block) in the built docs. Am I missing something?

austingmhuang: Oh. Hmm, think I need to update the .rst file.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -208,6 +208,61 @@ class QiskitDevice2(Device):
...
+        @qml.qnode(dev)
+        def circuit(x):
+            qml.RX(x, wires=[0])
+            qml.CNOT(wires=[0, 1])
+            return qml.expval(qml.PauliZ(1))
+
+    >>> circuit(np.pi/3, shots=1024)
+    0.529296875
+
+    This device also supports the use of local simulators such as FakeManila or AerSimulator
```
obliviateandsurrender, line 232:
```suggestion
    This device also supports the use of local simulators such as ``AerSimulator`` or fake backends such as ``FakeManila``. 
```

pennylane_qiskit/qiskit_device2.py
```diff
@@ -208,6 +208,61 @@ class QiskitDevice2(Device):
...
+        @qml.qnode(dev)
+        def circuit(x):
+            qml.RX(x, wires=[0])
+            qml.CNOT(wires=[0, 1])
+            return qml.expval(qml.PauliZ(1))
+
+    >>> circuit(np.pi/3, shots=1024)
+    0.49755859375
+
+    Internally, the device uses the EstimatorV2 and the SamplerV2 runtime primitives to execute
```
obliviateandsurrender, line 251:
```suggestion
    Internally, the device uses the EstimatorV2 and the SamplerV2 runtime primitives to execute
```
Could we link `EstimatorV2` and `SamplerV2` to qiskit pages for them?

pennylane_qiskit/qiskit_device2.py
```diff
@@ -208,6 +208,61 @@ class QiskitDevice2(Device):
...
+    the measurements. To set options for transpilation or runtime (see <https://docs.quantum.ibm.com/run/configure-runtime-compilation>
+    and https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/options), simply pass the
+    keyword arguments into the device. If you wish to change options other than shots,
```
obliviateandsurrender, line 254:
```suggestion
    keyword arguments into the device. If you wish to change options other than ``shots``,
```

pennylane_qiskit/qiskit_device2.py
```diff
@@ -208,6 +208,61 @@ class QiskitDevice2(Device):
...
+
+    >>> circuit(np.pi/3, shots=1024)
+    0.529296875
+
+    This device also supports the use of local simulators such as FakeManila or AerSimulator
+
+    .. code-block:: python
+
+        import pennylane as qml
+        from qiskit_ibm_runtime.fake_provider import FakeManilaV2
```
obliviateandsurrender, line 237: Can we change this example to use ``AerSimulator`` and last example can then still use ``FakeManilaV2``.

pennylane_qiskit/qiskit_device2.py
```diff
@@ -208,6 +208,61 @@ class QiskitDevice2(Device):
...
+    PennyLane requires you to re-initialize the device to do so.
+
+    .. code-block:: python
+
+        import pennylane as qml
+        from qiskit_ibm_runtime.fake_provider import FakeManilaV2
+
+        backend = FakeManilaV2()
+        dev = qml.device("qiskit.remote", wires=5, backend=backend, resilience_level=1, optimization_level=1, seed_transpiler=42)
+        # to change options, re-initialize the device
```
obliviateandsurrender, line 264: Can we also show an example of changing `shots` as well?

obliviateandsurrender (changes requested):

pennylane_qiskit/qiskit_device2.py
soranjh: If the users are not supposed to use QiskitDevice2 in any scenario, then it can stay hidden as is.

pennylane_qiskit/remote.py
```diff
@@ -25,16 +25,95 @@ class RemoteDevice(QiskitDevice):
...
-            BackendV1 or BackendV2 instance.
-        shots (int or None): number of circuit evaluations/random samples used
-            to estimate expectation values and variances of observables. For statevector backends,
-            setting to ``None`` results in computing statistics like expectation values and variances analytically.
+            or strings (``['aux_wire', 'q1', 'q2']``).
+        backend (Backend): the initialized Qiskit backend
 
     Keyword Args:
-        name (str): The name of the circuit. Default ``'circuit'``.
+        shots (int or None): number of circuit evaluations/random samples used
```
obliviateandsurrender, line 32:
```suggestion
        shots (Union[int, None]): number of circuit evaluations/random samples used
```

pennylane_qiskit/remote.py
```diff
@@ -25,16 +25,95 @@ class RemoteDevice(QiskitDevice):
...
+
+        @qml.qnode(dev)
+        def circuit(x):
+            qml.RX(x, wires=[0])
+            qml.CNOT(wires=[0, 1])
+            return qml.expval(qml.PauliZ(1))
+
+        circuit(np.pi/3) # this will run with 4000 shots
+        circuit(np.pi/3, shots=10000) # this will run with 10000 shots
+        circuit(np.pi/3) # this will run with 4000 shots
```
obliviateandsurrender, line 100: Maybe use the `sample` measurement and show the length of the output to confirm with comments.

pennylane_qiskit/remote.py
```diff
@@ -25,16 +25,95 @@ class RemoteDevice(QiskitDevice):
...
+
+    .. code-block:: python
+
+        import pennylane as qml
+        from qiskit_ibm_runtime.fake_provider import FakeManilaV2
+
+        backend = FakeManilaV2()
+        dev = qml.device("qiskit.remote", wires=5, backend=backend, resilience_level=1, optimization_level=1, seed_transpiler=42)
+        # to change options, re-initialize the device
+        dev = qml.device("qiskit.remote", wires=5, backend=backend, resilience_level=1, optimization_level=2, seed_transpiler=24)
```
lillian542, line 116: These lines are very long and don't as nicely in the docs (you get a long scroll bar in the code block). Could we add a line-break after each arg/kwarg being passed to `device` to keep the lines short?

lillian542 (commented):
It feels a little strange to update the docstring of the `remote` device to reflect changes to the new device, when it hasn't been switched over yet. I guess it will all happen in the end anyway, but should we do that first (in a separate PR)? Might also be time to rename them `QiskitDevice` --> `QiskitDeviceLegacy` and `QiskitDevice2` --> `QiskitDevice`.

pennylane_qiskit/remote.py
austingmhuang: > It feels a little strange to update the docstring of the remote device to reflect changes to the new device, when it hasn't been switched over yet

I thought about this and I think either works for me. I wanted to rename the devices in this PR actually (if the changes weren't large) but it seems like there are some additional changes that warrant a separate PR. It does feel a little backwards to update the docstrings in this PR before renaming the devices so I think it's fine to revert this change here and just open a separate PR that will change all of it at the same time. Let me open that PR now so it's obvious.

lillian542: That sounds like extra work and I don't think it matters much, let's not worry about it :)

lillian542 (approved):
🚀

obliviateandsurrender (approved):
Should be good to go once CI passes!

PR #553, opened 2024-06-06
Tests with fakehardware

austingmhuang:
This PR parametrizes certain tests to run with fakehardware instead. These tests have been modified (when needed) to test that formats are matching rather than the answers are correct. Additionally we make our tests flaky rather than have a high number of shots in order to stop the CI from failing due to random chance.


austingmhuang:
[sc-63546]

tests/test_base_device.py
```diff
@@ -1113,36 +1125,37 @@ def qiskit_circuit():
...
-        assert np.allclose(res, qiskit_res, atol=0.03)
 
-    def test_sampler_output_shape(self):
+    @pytest.mark.parametrize("backend", [backend, FakeManila(), FakeManilaV2()])
+    def test_sampler_output_shape(self, backend):
         """Test that the shape of the results produced from the sampler for the Qiskit device
         is consistent with Pennylane"""
-        dev = qml.device("default.qubit", wires=[0, 1, 2, 3], shots=1024)
-        qiskit_dev = QiskitDevice2(wires=[0, 1, 2, 3], backend=backend)
+        dev = qml.device("default.qubit", wires=5, shots=1024)
```
austingmhuang, line 1133: Need to change number of wires to match the number of qubits for the fake hardware.

tests/test_base_device.py
```diff
@@ -1255,18 +1269,19 @@ def circuit():
...
+        qiskit_dev = QiskitDevice2(wires=3, backend=backend, shots=4000)
 
         @qml.qnode(qiskit_dev)
         def qiskit_circuit():
             qml.RX(np.pi / 3, 0)
             qml.RZ(np.pi / 3, 0)
             return observable()
 
-        dev = qml.device("default.qubit", wires=3, shots=30000)
+        dev = qml.device("default.qubit", wires=3, shots=4000)
```
austingmhuang, line 1284: Dropped the number of shots since we have flaky tests instead.

tests/test_base_device.py
```diff
@@ -204,13 +205,15 @@ def test_setting_simulator_noise_model(self):
 class TestQiskitSessionManagement:
     """Test using Qiskit sessions with the device"""
 
-    def test_default_no_session_on_initialization(self):
+    @pytest.mark.parametrize("backend", [backend, FakeManila(), FakeManilaV2()])
```
lillian542, line 208: could we rename `backend` to `aer_backend` so its immediately clear what it is?

lillian542 (commented):
What would happen if we added the two FakeManilla backends to the list of `pl_devices` in the integration tests? I know many of those tests are specific to the old API, but I wonder how many aren't (or wouldn't be with minor tweaks).

austingmhuang:
> What would happen if we added the two FakeManilla backends to the list of `pl_devices` in the integration tests? I know many of those tests are specific to the old API, but I wonder how many aren't (or wouldn't be with minor tweaks).

Hmm, well the `pl_devices` in the integration tests are not actually the devices but the providers. So it's not just a matter of appending FakeManila to the list, you'll probably have to tweak the whole thing to move off of using providers+string to using devices instead. Doable but likely much easier after #566 is merged I think.

lillian542 (approved):

PR #555, opened 2024-06-07
add qasm measurement

MashAliK:
For https://github.com/PennyLaneAI/pennylane/issues/5646.
Updated `load_qasm` to take optional measurements as an argument and added the ability for it to make mid-circuit measurements.
Tests added in `tests/test_converter.py`.

### DIFF
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -20,6 +20,7 @@ import warnings
 from functools import partial, reduce
 
 import numpy as np
+import qiskit.qasm2
 from qiskit import QuantumCircuit
 from qiskit.circuit import Parameter, ParameterExpression, ParameterVector
 from qiskit.circuit import Measure, Barrier, ControlFlowOp, Clbit
@@ -584,14 +585,17 @@ def load(quantum_circuit: QuantumCircuit, measurements=None):
     return _function
 
 
-def load_qasm(qasm_string: str):
+def load_qasm(qasm_string: str, measurements=None):
     """Loads a PennyLane template from a QASM string.
     Args:
         qasm_string (str): the name of the QASM string
+        measurements (None | pennylane.measurements.MeasurementProcess | list[pennylane.measurements.MeasurementProcess]):
+            the PennyLane `measurements <https://docs.pennylane.ai/en/stable/introduction/measurements.html>`_
+            that override the terminal measurements that may be present in the input circuit
     Returns:
         function: the new PennyLane template
     """
-    return load(QuantumCircuit.from_qasm_str(qasm_string), measurements=[])
+    return load(qiskit.qasm2.loads(qasm_string), measurements=measurements)
 
 
 def load_qasm_from_file(file: str):
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -587,11 +587,13 @@ def load(quantum_circuit: QuantumCircuit, measurements=None):
 
 def load_qasm(qasm_string: str, measurements=None):
     """Loads a PennyLane template from a QASM string.
+
     Args:
         qasm_string (str): the name of the QASM string
         measurements (None | pennylane.measurements.MeasurementProcess | list[pennylane.measurements.MeasurementProcess]):
             the PennyLane `measurements <https://docs.pennylane.ai/en/stable/introduction/measurements.html>`_
             that override the terminal measurements that may be present in the input circuit
+
     Returns:
         function: the new PennyLane template
     """
--- a/tests/test_converter.py
+++ b/tests/test_converter.py
@@ -1165,8 +1165,8 @@ class TestConverterQasm:
         with recorder:
             quantum_circuit(params={})
 
-        # only X and CNOT queued (not 4 x qml.measure)
-        assert len(recorder.queue) == 2
+        # X and CNOT queued with 4 x qml.measure
+        assert len(recorder.queue) == 6
 
         assert recorder.queue[0].name == "PauliX"
         assert recorder.queue[0].parameters == []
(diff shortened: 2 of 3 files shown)


trbromley:
Thanks @MashAliK! :star_struck: I've requested reviews from our team.

pennylane_qiskit/converter.py
```diff
@@ -584,14 +585,17 @@ def _function(*args, params: dict = None, wires: list = None, **kwargs):
     return _function
 
 
-def load_qasm(qasm_string: str):
+def load_qasm(qasm_string: str, measurements=None):
```
lillian542, line 588: This is more of a discussion point, but I wonder if the default value here should be `[]` instead of `None`? 

That would preserve the current behaviour if someone is using this without passing measurements. But then this should also carry through to the PennyLane PR, and it will mean that the behaviour between `from_qiskit` and `from_qasm` is inconsistent without measurements.

I would be in favour of updating to this implementation, but then I guess this should be included in the changelog as a breaking change. For context, [here is a previous PR](https://github.com/PennyLaneAI/pennylane-qiskit/pull/469) where we specifically defaulted to `[]` when calling `load` (below) to _avoid_ making this breaking change. But I think if we add the `measurements` UI that `from_qiskit` has to this function, then its time to make the breaking change. @trbromley @obliviateandsurrender

tests/test_converter.py
```diff
@@ -1165,8 +1165,8 @@ def test_qasm_(self, recorder):
         with recorder:
             quantum_circuit(params={})
 
-        # only X and CNOT queued (not 4 x qml.measure)
-        assert len(recorder.queue) == 2
+        # X and CNOT queued with 4 x qml.measure
+        assert len(recorder.queue) == 6
```
lillian542, line 1169: this is what "breaks" by defaulting to `None` instead of `[]` above

pennylane_qiskit/converter.py
lillian542: See comment in the test file below for a concrete example of the behaviour that will change with this update

lillian542:
This is looking good, thanks @MashAliK 🎉  There are some failing tests, but I don't think they are related to your PR (most of them definitely aren't). We'll work on getting that updated and get back to you soon, don't worry about it for now.

Don't forget to update the changelog to include the change, and list yourself as a contributor! This should definitely go under "Improvements". As mentioned above, we may also want to include a second entry for it in "Breaking changes" - let's see what people think makes sense.

pennylane_qiskit/converter.py
lillian542: @MashAliK , we've discussed this offline and we'll stick with the implementation you have here for the PR, no need to update anything

pennylane_qiskit/converter.py
```diff
@@ -584,14 +585,17 @@ def _function(*args, params: dict = None, wires: list = None, **kwargs):
...
     """Loads a PennyLane template from a QASM string.
     Args:
         qasm_string (str): the name of the QASM string
+        measurements (None | pennylane.measurements.MeasurementProcess | list[pennylane.measurements.MeasurementProcess]):
+            the PennyLane `measurements <https://docs.pennylane.ai/en/stable/introduction/measurements.html>`_
+            that override the terminal measurements that may be present in the input circuit
```
lillian542, line 594: Something here is causing Sphinx build for the docs to fail - I think you need to add a blank line here before the `Returns` section for the docs to render correctly

tests/test_converter.py
```diff
@@ -1176,6 +1176,66 @@ def test_qasm_(self, recorder):
         assert recorder.queue[1].parameters == []
         assert recorder.queue[1].wires == Wires([2, 0])
 
+        for i in range(2, 6):
+            assert recorder.queue[i].name == "MidMeasureMP"
+            assert recorder.queue[i].wires == Wires([i - 2])
+
+    def test_qasm_measure(self):
+        """Tests that measurements specified as an argument are added to the converted circuit."""
+        qasm_string = 'include "qelib1.inc";' "qreg q[2];" "creg c[2];" "h q[0];" "cx q[0], q[1];"
```
lillian542, line 1185: Reformatting this string will fix the formatting complaints in CI regarding this line. For details on what the formatting check is catching, you can see [the documentation here](https://peps.python.org/pep-3126/)

tests/test_converter.py
```diff
@@ -1176,6 +1176,66 @@ def test_qasm_(self, recorder):
...
+            "cx q[0], q[1];"
+            "measure q -> c;"
+        )
+        dev = qml.device("default.qubit")
+        loaded_circuit = load_qasm(qasm_string)
+
+        # loaded circuit
+        @qml.qnode(dev)
+        def quantum_circuit1():
+            mid_measure, m0, m1 = loaded_circuit()
```
lillian542, line 1221: The formatting checker in CI doesn't like unused variables - this or a similar update would fix it:

```suggestion
            mid_measure, _, m1 = loaded_circuit()
```

tests/test_converter.py
```diff
@@ -1176,6 +1176,66 @@ def test_qasm_(self, recorder):
...
+            return qml.expval(mid_measure), qml.expval(m1)
+
+        # native pennylane circuit
+        @qml.qnode(dev)
+        def quantum_circuit2():
+            qml.Hadamard(0)
+            mid_measure = qml.measure(0)
+            qml.RZ(0.24, 0)
+            qml.CNOT([0, 1])
+            m0 = qml.measure([0])
```
lillian542, line 1232: Same as above wrt unused variables
```suggestion
            qml.measure([0])
```

lillian542 (commented):
The testing issue from the master branch should be resolved very soon! 

In the meantime, there are a couple of minor checks that are failing for documentation and formatting that should be updated here, so I've left a few comments :)

pennylane_qiskit/converter.py
obliviateandsurrender, line 594:
```suggestion
            that override the terminal measurements that may be present in the input circuit

```

tests/test_converter.py
obliviateandsurrender, line 1185: Could we rewrite the string as the one below in `test_qasm_mid_circuit_measure`? (i.e., one instruction in one line)

obliviateandsurrender (commented):
Thanks @MashAliK, good work! Leaving a couple of suggestions similar to Lillian's.

MashAliK:
@lillian542 @obliviateandsurrender Thanks for the suggestions!
I've made an update to try to resolve these problems.

tests/test_converter.py
```diff
@@ -1176,6 +1176,72 @@ def test_qasm_(self, recorder):
         assert recorder.queue[1].parameters == []
         assert recorder.queue[1].wires == Wires([2, 0])
 
+        for i in range(2, 6):
+            assert recorder.queue[i].name == "MidMeasureMP"
+            assert recorder.queue[i].wires == Wires([i - 2])
+
+    def test_qasm_measure(self):
```
obliviateandsurrender, line 1183: Can you run `black -l 100 pennylane-qiskit tests` to fix the `black` formatting errors?

pennylane_qiskit/converter.py
lillian542: Great, looks like the build succeeds now!

tests/test_converter.py
MashAliK: When I run this command, the only reformatting done is that the change I made to make `qasm_string` so that it is multiline gets reverted back into a single-line definition like it was before. Should I keep it single-line as Black is suggesting?

obliviateandsurrender: Yes, please keep the formatting the black suggests for the `Formatting checks` to pass.
If you get the single-line definition, then add the strings components explicitly with `+`: 
`"abc" + "def" == "abcdef"`

obliviateandsurrender (approved):
Looks good to go!

lillian542 (approved):
Nice work @MashAliK!

PR #557, opened 2024-06-11
Account for new warning messages from Qiskit

austingmhuang:
CI on Master breaks due to some new warning messages from Qiskit. This was already fixed on #536, but we'll fix it here on master as well.


austingmhuang:
Closing this since @obliviateandsurrender's PR #559 covers this as well.

PR #558, opened 2024-06-11
Diagonalize gates

austingmhuang:
**Context:** We didn't have `split_non_commuting` before, which meant observables with commuting measurements would return incorrect answers

**Changes**:
The`split_non_commuting` transform has been added and will run when given non-commuting measurements. Observables like Prod or Sum are now supported as well since these transforms are now working. We also add a lot of tests for both the Sampler and the Estimator.

**Notes**:
There is a small bug with the formatting from the output for the transform, but it is not due to the behaviour of the plugin. There is also a bug on Qiskit's end with their Estimator and their way of calculating the variance that we cannot work around.


pennylane_qiskit/qiskit_device2.py
```diff
@@ -219,6 +218,10 @@ class QiskitDevice2(Device):
         "Hadamard",
         "Hermitian",
         "Projector",
+        "Prod",
+        "Sum",
+        "LinearCombination",
+        "SProd",
```
austingmhuang, line 224: I didn't add Hamiltonian or SparseHamiltonian because well
for Hamiltonian, this just gets casted to LinearCombination from what I can tell so it's kind of extraneous
for SparseHamiltonian, it seems like it's a bit of a tricky case to handle and... not worth it effort to reward ratio wise. Could be nice for the backlog :shrug:

tests/test_base_device.py
```diff
@@ -1114,3 +1174,171 @@ def qiskit_circuit(x):
...
+        def circuit():
+            qml.RX(np.pi / 3, 0)
+            qml.RZ(np.pi / 3, 0)
+            return observable()
+
+        qiskit_res = qiskit_circuit()
+        res = circuit()
+
+        for i in range(len(res)):  # pylint: disable=consider-using-enumerate
+            assert abs(sum(qiskit_res[i]) / 30000 - sum(res[i]) / 30000) < 0.05
```
austingmhuang, line 1344: A crude way of comparing expectation values but it works :shrug:

austingmhuang:
[sc-54392]

austingmhuang:
[sc-62047]

pennylane_qiskit/qiskit_device2.py
obliviateandsurrender: Might be nice to leave a `#TODO` there.

tests/test_base_device.py
```diff
@@ -1114,3 +1174,171 @@ def qiskit_circuit(x):
...
```
obliviateandsurrender, line 1344: What does `ith` index give here for the results?
Can we do something like `np.abs(np.sum(qiskit_res - res, axis=0)) / 30000 < 0.05`

tests/test_base_device.py
```diff
@@ -1114,3 +1174,171 @@ def qiskit_circuit(x):
...
+
+        qiskit_res = qiskit_circuit()
+        res = circuit()
+
+        assert np.allclose(res, qiskit_res, atol=0.05)
+
+    @pytest.mark.parametrize(
+        "observable",
+        [
+            lambda: [qml.counts(qml.X(0) + qml.Y(0)), qml.counts(qml.X(0))],
```
obliviateandsurrender, line 1254: Instead of using `lambda`, you can use `qml.apply(measurement)` to queue the measurement.

austingmhuang: Ohh didn't know that. Thanks @obliviateandsurrender!

tests/test_base_device.py
austingmhuang: `res` or `qiskit_res` are an array of arrays, where the inner arrays are the results of the `qml.sample`. 
Could do `assert np.all(np.abs(np.sum(qiskit_res, axis=1) / 30000 - np.sum(res, axis=1) / 30000) < 0.05)` instead.

tests/test_base_device.py
austingmhuang: Seems like qml.apply is bugged for operators like `[qml.counts(qml.X(0) + qml.Y(0)), qml.counts(qml.X(0))]`

```
dev = qml.device("default.qubit", wires=2, shots=1024)

op = qml.sample(qml.X(0) + qml.Y(0)), qml.sample(qml.X(0))
@qml.qnode(dev)
def circuit():
    qml.RX(np.pi / 3, 0)
    qml.RZ(np.pi / 3, 0)
    return qml.apply(op)

>>> circuit()
()
```

austingmhuang: For now I will stick with lambda since it works. If this is a bug I can put a TODO and fix it for later, but I don't think it's that big of a deal.

austingmhuang: Seems like it is not designed to work like this, I will use lambda.

tests/test_base_device.py
```diff
@@ -1114,3 +1174,170 @@ def qiskit_circuit(x):
...
+            return observable()
+
+        qiskit_res = qiskit_circuit()
+        res = circuit()
+
+        equals = lambda a, b: abs(a - b) < 300  # tolerance of 1% of # of shots
+
+        for i in range(len(res)):  # pylint: disable=consider-using-enumerate
+            for key in set(res[i].keys()) & set(qiskit_res[i].keys()):
+                assert equals(res[i][key], qiskit_res[i][key])
```
obliviateandsurrender, line 1287:
```suggestion
         assert len(qiskit_res) == len(res)
         for res1, res2 in zip(qiskit_res, res):
             assert all([res1[key] - res2.get(key, 0) < 300 for key in res1])
```

tests/test_base_device.py
```diff
@@ -1114,3 +1174,170 @@ def qiskit_circuit(x):
...
+        @qml.qnode(dev)
+        def circuit():
+            qml.RX(np.pi / 3, 0)
+            qml.RZ(np.pi / 3, 0)
+            return observable()
+
+        qiskit_res = qiskit_circuit()
+        res = circuit()
+
+        assert np.all(np.abs(np.sum(qiskit_res, axis=1) - np.sum(res, axis=1)) / 30000 < 0.05)
```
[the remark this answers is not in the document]
austingmhuang: Ohh didn't know this existed. Thanks @obliviateandsurrender !

obliviateandsurrender (approved):

lillian542 (approved):
This looks good to me! 🎉

PR #559, opened 2024-06-11
Fix asserts for `Conditionals` for controlled flow operations

obliviateandsurrender:
`Conditional` checks were failing as they were trying to check equality between underlying measurements, which will turn out to be false due to different hash resulting from different `id`.

### DIFF
--- a/pennylane_qiskit/converter.py
+++ b/pennylane_qiskit/converter.py
@@ -867,11 +867,15 @@ def _process_switch_condition(condition, mid_circ_regs):
         if all(clbit in mid_circ_regs for clbit in clbits):
             # Build an integer representation for each switch case
             meas_pl_op = sum(2**idx * mid_circ_regs[clbit] for idx, clbit in enumerate(clbits))
+            # Non Expr-based condition can have 2**#clbits Boolean outputs: 0, ..., 2**#clbits - 1
+            # If all of them are already covered in the given cases, skip the default case.
             use_switch_default = bool(set(condition[1]) ^ set(range(2 ** len(clbits))))
 
     # if the target is an Expr
     else:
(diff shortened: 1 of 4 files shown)


pennylane_qiskit/converter.py
```diff
@@ -866,17 +867,19 @@ def _process_switch_condition(condition, mid_circ_regs):
         if all(clbit in mid_circ_regs for clbit in clbits):
             # Build an integer representation for each switch case
             meas_pl_op = sum(2**idx * mid_circ_regs[clbit] for idx, clbit in enumerate(clbits))
+            use_switch_default = bool(set(condition[1]) ^ set(range(2 ** len(clbits))))
```
lillian542, line 870: I'm a little lost wrt what is going on here, could we add a comment saying what is happening?

obliviateandsurrender: Added notes for concise explanation.

pennylane_qiskit/converter.py
```diff
@@ -866,17 +867,23 @@ def _process_switch_condition(condition, mid_circ_regs):
         if all(clbit in mid_circ_regs for clbit in clbits):
             # Build an integer representation for each switch case
             meas_pl_op = sum(2**idx * mid_circ_regs[clbit] for idx, clbit in enumerate(clbits))
+            # Non Expr-based condition can have 2**#clbits Boolean outputs: 0, ..., 2**#clbits - 1
```
obliviateandsurrender, line 870:
```suggestion
            # Non Expr-based condition can have 2**#clbits outputs: 0, ..., 2**#clbits - 1
```

lillian542 (approved):
Thanks @obliviateandsurrender!

austingmhuang (approved):
Looks good to me!

PR #566, opened 2024-06-21
Use new qiskit device as the base for remote

austingmhuang:
**Context**
The new qiskit device is shaping up to look pretty good so it's time to update `qiskit.remote` so that it uses the new qiskit device api instead of the legacy device api. 

**Changes**
This PR makes `qiskit.remote` inherit from the new qiskit device. Basically, we rename `QiskitDevice2.py` to `QiskitDevice.py` and the original `QiskitDevice.py` to `QiskitDeviceLegacy.py`. Tests are rewritten and imports are redefined accordingly. 

**Notes**
For now, we still inherit from `QiskitDeviceLegacy` for `qiskit.aer` and `qiskit.basicsim`. We could, however, opt to make this change in this PR as well, although it seems like there is a significant amount of refactoring that will be needed to make this work. My personal opinion is that if we do this, we should do this in a separate PR since I think the changes are already a bit confusing as is.


austingmhuang:
[sc-54388]

pennylane_qiskit/qiskit_device_legacy.py
austingmhuang, line 1: Renamed from Qiskit_device to Qiskit_device_legacy

pennylane_qiskit/qiskit_device2.py
austingmhuang, line 1: Renamed from `qiskit_device2` to `qiskit_device`

pennylane_qiskit/qiskit_device.py
austingmhuang, line 1: Nothing was changed in the original `qiskit_device2` and all the changes in this file are just a result of the renaming.

tests/test_base_device.py
austingmhuang, line 1: These are ONLY changes that result from changing `QiskitDevice2` to QiskitDevice`

tests/test_integration.py
```diff
@@ -23,42 +23,29 @@
...
-        """check compatibility of provided backend"""
-        dev_name, _ = pldevice
-        if (dev_name == "qiskit.aer" and "aer" not in backend_name) or (
-            dev_name == "qiskit.basicaer" and "aer" in backend_name
-        ):
-            return (False, "Only the AerSimulator is supported on AerDevice")
-        return True, None
-
-else:
-    from qiskit.providers.basic_provider import BasicProvider
```
austingmhuang, line 49: Although unrelated to this PR it is deleted since we don't run below v1.0 anymore.

tests/test_qiskit_device.py
```diff
@@ -109,7 +109,7 @@ class TestSupportForV1andV2:
     )
     def test_v1_and_v2_mocked(self, dev_backend):
         """Test that device initializes with no error mocked"""
-        dev = qml.device("qiskit.remote", wires=10, backend=dev_backend, use_primitives=True)
+        dev = qml.device("qiskit.aer", wires=10, backend=dev_backend)
```
austingmhuang, line 112: This change is just to keep code cov happy.

tests/test_integration.py
```diff
@@ -90,31 +77,18 @@ def test_load_remote_device_with_backend_instance(self, d, backend):
...
-        assert dev.shots == 1024
-        assert dev.short_name == "qiskit.remote"
-        assert dev.provider is None
-        assert dev.capabilities()["returns_state"] == (backend in state_backends)
-
-    @pytest.mark.parametrize("d", pldevices)
-    def test_load_remote_device_by_name(self, d, backend):
-        """Test that the qiskit.remote device loads correctly when passed a provider and a backend
-        name. This test is equivalent to `test_load_device` but on the qiskit.remote device instead
-        of specialized devices that expose more configuration options."""
```
austingmhuang, line 104: We delete this test since `provider` is no longer an accepted keyword argument.

tests/test_integration.py
```diff
@@ -129,12 +103,6 @@ def test_incorrect_backend_wires(self):
         ):
             qml.device("qiskit.aer", wires=100, method="statevector")
 
-    def test_remote_device_no_provider(self):
-        """Test that the qiskit.remote device raises a ValueError if passed a backend
-        by name but no provider to look up the name on."""
-        with pytest.raises(ValueError, match=r"Must pass a provider"):
-            qml.device("qiskit.remote", wires=2, backend="aer_simulator_statevector")
-
```
austingmhuang, line 137: No more provider so this error message is gone as well.

pennylane_qiskit/converter.py
```diff
@@ -36,7 +36,7 @@
 import pennylane as qml
 import pennylane.ops as pennylane_ops
 from pennylane.tape.tape import rotations_and_diagonal_measurements
-from pennylane_qiskit.qiskit_device import QISKIT_OPERATION_MAP
+from pennylane_qiskit.qiskit_device_legacy import QISKIT_OPERATION_MAP
```
lillian542, line 39: since `QISKIT_OPERATION_MAP` isn't really a "legacy" object but will be used by the new device and conversion going forward, should we move it to either the new `qiskit_device` folder or to `converter`, and update imports accordingly?

tests/test_integration.py
```diff
@@ -90,32 +76,18 @@ def test_load_remote_device_with_backend_instance(self, d, backend):
...
-        of specialized devices that expose more configuration options."""
-
-        # check compatibility between provider and backend, and skip if incompatible
-        is_compatible, failure_msg = check_provider_backend_compatibility(d, backend)
-        if not is_compatible:
-            pytest.skip(failure_msg)
-
-        _, provider = d
+        if backend_instance.configuration().n_qubits is None:
+            pytest.skip("No qubits?")
```
lillian542, line 80: When does this happen? If its simulators with an arbitrary number of qubits, could we do 

```
n_qubits = backend_instance.configuration().n_qubits
n_wires = n_qubits if n_qubits else 10
```
and then pass `n_wires` below to avoid skipping those devices in tests?

lillian542 (commented):
Looking good! I have a couple of suggestions, see what you think :)

pennylane_qiskit/converter.py
austingmhuang: Hmm I personally lean towards moving it to converter (to me it seems more like a utility thing?) so that, suppose in 5 years we have a new new device, this import doesn't need to change. Plus it just makes sense intuitively to import the `QISKIT_OPERATION_MAP` from the file where it is actually defined.

tests/test_integration.py
austingmhuang: This was probably added when I was experimenting with adding FakeManila to the pl_devices. I probably was in the wrong branch lol. I deleted it from this branch (since it's not relevant) but something similar might come back when addressing using FakeManila for the integration tests.

pennylane_qiskit/qiskit_device.py
```diff
@@ -12,124 +12,204 @@
...
 
-class QiskitDevice(QubitDevice, abc.ABC):
-    r"""Abstract Qiskit device for PennyLane.
+@transform
+def split_execution_types(
+    tape: qml.tape.QuantumTape,
+) -> (Sequence[qml.tape.QuantumTape], Callable):
+    """Split into separate tapes based on measurement type. Counts and sample-based measurements
+    will use the Qiskit Sampler. ExpectationValue and Variance will use the Estimator, except
+    when the measured observable does not have a `pauli_rep`. In that case, the Sampler will be
```
obliviateandsurrender, line 130: When do we not have a `pauli_rep` for the observables? 🤔 
For `Projectors` and `Hermitian` unitaries, can we not rely on decomposing them to a Pauli basis either via PL or qiskit and then use them with `Estimator`? For the rest, if this fails the fallback method can still be the `Sampler` approach.

pennylane_qiskit/qiskit_device.py
```diff
@@ -12,124 +12,204 @@
...
+            ]
+        )
+
+    def reorder_fn(res):
+        """re-order the output to the original shape and order"""
+
+        flattened_indices = [i for group in order_indices for i in group]
+        flattened_results = [r for group in res for r in group]
+
+        result = dict(zip(flattened_indices, flattened_results))
```
obliviateandsurrender, line 179: We should `assert` the length before zipping so we can ensure we don't miss anything. You can skip codecov check for the condition in case you don't have a unit test for it.

pennylane_qiskit/qiskit_device.py
austingmhuang: For example, observables like `qml.Hadamard` don't have a `pauli_rep`. In this case we use the `Sampler` instead and diagonalize ourselves. 

>For Projectors and Hermitian unitaries, can we not rely on decomposing them to a Pauli basis either via PL or qiskit and then use them with Estimator? 

I _think_ this is doable but is the extra work worth it when we could just use the Sampler approach instead?

lillian542 (approved):
Looks good, thanks! 🚀

doc/requirements.txt
```diff
@@ -3,7 +3,7 @@ docutils==0.20.1
 ipykernel==6.29.2
 jinja2==3.1.3
 nbsphinx==0.9.3
-pennylane==0.37
+pennylane==0.34
```
austingmhuang, line 6: I do not know why but sphinx does not build at v0.36 or 0.37.

pennylane_qiskit/qiskit_device.py
```diff
@@ -12,525 +12,664 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 r"""
-This module contains a base class for constructing Qiskit devices for PennyLane.
+This module contains a prototype base class for constructing Qiskit devices
+for PennyLane with the new device API.
 """
-# pylint: disable=too-many-instance-attributes,attribute-defined-outside-init
+# pylint: disable=too-many-instance-attributes,attribute-defined-outside-init, missing-function-docstring
```
obliviateandsurrender, line 18: Why do we need `missing-function-docstring`? Can we not assign docstrings to all the functions here?

obliviateandsurrender (approved):
A couple of non-blocking questions otherwise, good to go!

pennylane_qiskit/qiskit_device.py
austingmhuang: I think you can safely delete this without issue since `execute()` has a docstring now.

PR #569, opened 2024-07-05
Add conversion support for Qiskit's `NoiseModel`s

obliviateandsurrender:
**Context:** Add support for converting Qiskit's noise models to PennyLane's noise models. 

1. `noise_models.py` contains helper methods for particular things
    - Converting a Qiskit's `Error` object to that of a corresponding Qiskit's instructions consisting of their Pauli or Kraus representation
    - Using the above representation to map it to a corresponding PennyLane Channel's description.
    - Converting a Qiskit's `NoiseModel` object to a list of PennyLane Channel objects that are constructed based on the above description.
2. `converter.py` contains the `load_noise_model` method, which accepts a Qiskit noise model along with some mentioned keyword arguments and returns the corresponding PennyLane noise model.

**Benefits:**
Noise model construction becomes easier and hardware noise models can be better studied. 

**Possible Drawbacks:** 
1. Kraus representations of errors are not unique. Since Qiskit simply stores the final description in a crude way consisting either of the Pauli words or via the Kraus operators, the conversion mapping doesn't remain one-to-one and could lead to a channel corresponding to a different error but with the same canonicalized Kraus representation for the computed parameters.
2. Readout errors are not yet supported. 

**Related GitHub Issues:** [sc-68045]


pennylane_qiskit/noise_models.py
```diff
@@ -0,0 +1,421 @@
...
+    for gate_name, qubit_dict in noise_model._local_quantum_errors.items():
+        for qubits, error in qubit_dict.items():
+            noise_op = _build_qerror_op(error, gate_name=gate_name, **kwargs)
+            qerror_dmap[noise_op][qubits].append(qiskit_op_map[gate_name])
+
+    # TODO: Add support for the readout error
+    rerror_dmap = defaultdict(lambda: defaultdict(list))
+    if noise_model._default_readout_error or noise_model._local_readout_errors:
+        warn("Readout errors are not supported currently and will be skipped.")
+
```
isaacdevlugt, line 420: @trbromley we have in the requirements that if readout noise is present, PennyLane raises an error. But, I'm wondering if raising a warning instead and _skipping_ the readout noise makes more sense? 🤔

obliviateandsurrender:
A minimalistic version of this PR is implemented [here](https://github.com/PennyLaneAI/pennylane-qiskit/pull/577).

PR #576, opened 2024-07-12
Add error message for shot vectors for QiskitDevice

austingmhuang:
As mentioned by @CatalinaAlbornoz in issue #575, there's a bug that's ultimately related to `qiskit.aer` not supporting shot vectors (@albi3ro). While the functionality is still not supported, in this PR, we add a more useful error message to capture this behaviour.

This error message already exists in the new device API so there's no additional work on that front.

### DIFF
--- a/tests/test_qiskit_device.py
+++ b/tests/test_qiskit_device.py
@@ -326,3 +326,15 @@ class TestBatchExecution:
         res = barrier_func()
         assert barrier_func.tape.operations[0] == qml.Barrier([0, 1])
         assert np.allclose(res, dev.batch_execute([barrier_func.tape]), atol=0)
+
+def test_aer_device_shots_value_error():
+    """Tests that aer device raises an error when given a shot vector"""
+    with pytest.raises(
+        ValueError, match="Shots needs to be an integer value. Shot vectors are not supported"
+    ):
+        AerDevice(backend="aer_simulator", wires=1, shots=(1, 1, 1))
+
+    with pytest.raises(
+        ValueError, match="Shots needs to be an integer value. Shot vectors are not supported"
+    ):
+        AerDevice(backend="aer_simulator", wires=1, shots=[1, 1, 1])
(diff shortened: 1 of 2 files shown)


tests/test_qiskit_device.py
```diff
@@ -326,3 +326,16 @@ def barrier_func():
         res = barrier_func()
         assert barrier_func.tape.operations[0] == qml.Barrier([0, 1])
         assert np.allclose(res, dev.batch_execute([barrier_func.tape]), atol=0)
+
+
```
soranjh, line 330: You can use @pytest.mark.parametrize.

soranjh (approved):

albi3ro (commented):
You may also need to update the setter for shots.

If you do:
```
dev.shots = (10,100,100)
```

Do you get an error? This is what is used for dynamic shots.


We should double check that bug #575 is actually fixed.

austingmhuang:
> You may also need to update the setter for shots.
> 
> If you do:
> 
> ```
> dev.shots = (10,100,100)
> ```
> 
> Do you get an error? This is what is used for dynamic shots.
> 
> We should double check that bug #575 actually provides an informative error message now.


```
dev.shots = (10, 100, 100)
```

Does not raise an error for `qiskit.aer` since it does not have a setter for shots.
Seems like the error message doesn't show up for #575 when using `qiskit.aer`.

---------------------------------------------------------------

I tried using `qiskit.remote` with `AerSimulator` instead. This does raise an error for `dev.shots=(10, 100, 100)` but for the example in #575 the error message does not show up. Not sure how to proceed.

albi3ro:
> > You may also need to update the setter for shots.
> > If you do:
> > ```
> > dev.shots = (10,100,100)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do you get an error? This is what is used for dynamic shots.
> > We should double check that bug #575 actually provides an informative error message now.
> 
> ```
> dev.shots = (10, 100, 100)
> ```
> 
> Does not raise an error for `qiskit.aer` since it does not have a setter for shots (inherits from the legacy device). Seems like the error message doesn't show up for #575 when using `qiskit.aer`.
> 
> I tried using `qiskit.remote` with `AerSimulator` instead. This does raise an error for `dev.shots=(10, 100, 100)` but for the example in #575 the error message does not show up. Not sure how to proceed.

We may need to define a shots setter to add some validation is a shot vector is attempted to be set.

PR #577, opened 2024-07-15
Add conversion support for Qiskit noise models

obliviateandsurrender:
**Context:** Add support for converting Qiskit's noise models to PennyLane's noise models. 

1. `noise_models.py` contains helper methods for converting a Qiskit's `NoiseModel` object to a list of PennyLane Channel objects that are constructed based on their Kraus representation obtained via Qiskit.
2. `converter.py` contains the `load_noise_model` method, which accepts a Qiskit noise model along with some mentioned keyword arguments and returns the corresponding PennyLane noise model.

**Benefits:**
Noise model construction can be done in PennyLane from a Qiskit one. 

**Possible Drawbacks:** 
1. The Readability of converted noise models will not be the best as `QubitChannels` will only give Kraus representation and not any information on the corresponding error.

**Related GitHub Issues:** [sc-68045]


pennylane_qiskit/converter.py
```diff
@@ -1042,3 +1046,94 @@ def _expr_eval_clvals(clbits, clvals, expr_func, bitwise=False):
...
+    qerror_dmap, _ = _build_noise_model_map(noise_model, **kwargs)
+    model_map = {}
+    for error, wires_map in qerror_dmap.items():
+        conditions = []
+        for wires, operations in wires_map.items():
+            cond = qml.noise.op_in(operations)
+            if wires != AnyWires:
+                cond &= WiresIn(wires)
+            conditions.append(cond)
+        fcond = reduce(lambda cond1, cond2: cond1 | cond2, conditions)
```
austingmhuang, line 1131: Question: Rather than do an array of conditions and ORing across all elements, can we just set `fcond = False` at the beginning and simply do 
```
if qml.noise.op_in(operations):
    fcond = True
```
Effectively I think it's the same unless you need the `conditions` array for something else.

pennylane_qiskit/noise_models.py
```diff
@@ -0,0 +1,126 @@
...
+    "ecr": "ECR",
+    "sdg": qml.adjoint(qml.S),
+    "tdg": qml.adjoint(qml.T),
+    "sxdg": qml.adjoint(qml.SX),
+    "reset": qml.measure(AnyWires, reset=True),  # TODO: Improve reset support
+}
+
+
+def _build_qerror_op(error, **kwargs) -> qml.QubitChannel:
+    """Builds a PennyLane error channel from a Qiksit's QuantumError object.
```
austingmhuang, line 70:
```suggestion
    """Builds a PennyLane error channel from a Qiksit QuantumError object.
```

pennylane_qiskit/noise_models.py
```diff
@@ -0,0 +1,126 @@
...
+        quantum_error (bool): include quantum errors in the converted noise model. Default is ``True``
+        readout_error (bool): include readout errors in the converted noise model. Default is ``True``
+        decimals (int): number of decimal places to round the Kraus matrices. Default is ``10``
+
+    Returns:
+        (dict, dict): returns mappings for ecountered quantum errors and readout errors.
+
+    For plugin developers: noise model map tuple consists of following two (nested) mappings:
+        * qerror_dmap: noise_operation -> wires -> target_gate
+        * rerror_dmap: noise_operation -> wires -> target_measurement
```
austingmhuang, line 104: Prefer to be more verbose here, something like readout_error_dmap?

pennylane_qiskit/noise_models.py
```diff
@@ -0,0 +1,126 @@
...
+        noise_model (qiskit_aer.noise.NoiseModel): Qiskit's noise model
+        kwargs: Optional keyword arguments for providing extra information
+
+    Keyword Arguments:
+        quantum_error (bool): include quantum errors in the converted noise model. Default is ``True``
+        readout_error (bool): include readout errors in the converted noise model. Default is ``True``
+        decimals (int): number of decimal places to round the Kraus matrices. Default is ``10``
+
+    Returns:
+        (dict, dict): returns mappings for ecountered quantum errors and readout errors.
```
austingmhuang, line 100:
```suggestion
        (dict, dict): returns mappings for given quantum errors and readout errors.
```

pennylane_qiskit/noise_models.py
```diff
@@ -0,0 +1,126 @@
...
+    try:
+        kraus_matrices = np.round(Kraus(error).data, decimals=kwargs.get("decimals", 10))
+    except Exception as exc:  # pragma: no cover
+        raise ValueError(f"Error {error} could not be converted.") from exc
+
+    return qml.QubitChannel(K_list=kraus_matrices, wires=AnyWires)
+
+
+def _build_noise_model_map(noise_model, **kwargs) -> Tuple[dict, dict]:
+    """Builds a noise model map from which a PennyLane noise model can be constructed efficiently.
```
austingmhuang, line 88:
```suggestion
    """Builds a noise model map from a Qiskit noise model. This noise model map can be used to efficiently construct a PennyLane noise model.
```

Maybe this is better? Just want to clarify that the function takes in a qiskit noise model.

tests/test_converter.py
```diff
@@ -2206,3 +2207,79 @@ def test_convert_with_invalid_operator(self):
...
+    qiksit = pytest.importorskip("qiskit", "1.0.0")
+
+    @staticmethod
+    def _kraus_to_choi(krau_mats, optimize=False) -> np.ndarray:
+        r"""Transforms Kraus representation of a channel to its Choi representation."""
+        kraus_vecs = np.array([kraus.ravel(order="F") for kraus in krau_mats])
+        return np.einsum("ij,ik->jk", kraus_vecs, kraus_vecs.conj(), optimize=optimize)
+
+    def test_build_noise_model(self):
+        """Tests that _build_noise_model_map constructs correct model map for a noise model"""
```
austingmhuang, line 2225:
```suggestion
        """Tests that _build_noise_model_map constructs correct model map for a Qiskit noise model"""
```

tests/test_noise_models.py
```diff
@@ -0,0 +1,230 @@
...
+        error_1 = noise.depolarizing_error(depol1, 1)
+        error_2 = noise.depolarizing_error(depol2, 2)
+        error_3 = noise.phase_amplitude_damping_error(0.14, 0.24, excited_state_population=exc_pop)
+
+        noise_model = noise.NoiseModel()
+        noise_model.add_all_qubit_quantum_error(error_1, ["rz", "sx", "x"])
+        noise_model.add_all_qubit_quantum_error(error_2, ["cx"])
+        noise_model.add_all_qubit_quantum_error(error_3, ["ry", "rx"])
+
+        model_map, _ = _build_noise_model_map(noise_model)
```
austingmhuang, line 172:
```suggestion
        quantum_error_model_map, _ = _build_noise_model_map(noise_model)
```

Might be better to stay consistent across files

tests/test_noise_models.py
```diff
@@ -0,0 +1,230 @@
...
+    @pytest.mark.parametrize(
+        "combination, p_error",
+        [
+            (lambda err1, err2: err1.compose(err2), 0.052),
+            (lambda err1, err2: err1.tensor(err2), 0.037),
+            (lambda err1, err2: err1.expand(err2), 0.094),
+        ],
+    )
+    def test_composition_error_ops(self, combination, p_error):
+        """Tests that combination of quantum errors can be correctly converted into a PennyLane QubitChannel."""
```
austingmhuang, line 218:
```suggestion
        """Tests that a combination of quantum errors can be correctly converted into a PennyLane QubitChannel."""
```

tests/test_converter.py
austingmhuang: After reading the function -- are we not building a noise model here instead of a noise model map? Should update the docstring in general.

pennylane_qiskit/noise_models.py
```diff
@@ -0,0 +1,126 @@
...
+    Keyword Arguments:
+        quantum_error (bool): include quantum errors in the converted noise model. Default is ``True``
+        readout_error (bool): include readout errors in the converted noise model. Default is ``True``
+        decimals (int): number of decimal places to round the Kraus matrices. Default is ``10``
+
+    Returns:
+        (dict, dict): returns mappings for ecountered quantum errors and readout errors.
+
+    For plugin developers: noise model map tuple consists of following two (nested) mappings:
+        * qerror_dmap: noise_operation -> wires -> target_gate
```
austingmhuang, line 103: Might be better to just give an example of what qerror_dmap looks like?

austingmhuang:
```
qerror_dmap = {
    noise_op1: {
        AnyWires: [qiskit_op1, qiskit_op2],
        (0, 1): [qiskit_op3],
        (2,): [qiskit_op4]
    },
    noise_op2: { 
        AnyWires: [qiskit_op5],
        (1, 2): [qiskit_op6, qiskit_op7]
    }
}
```

Potentially takes a little too much room

austingmhuang (changes requested):
Looking good @obliviateandsurrender! Don't think I totally understand how the tests work for the noise models at a detailed level but I think I get it on a high-level basis. Seems like you take a qiskit noise model -> quantum error map (converting qiskit errors to kraus matrices) -> pennylane noise model?

pennylane_qiskit/converter.py
```diff
@@ -1042,3 +1046,94 @@ def _expr_eval_clvals(clbits, clvals, expr_func, bitwise=False):
...
+
+        c0 = qml.noise.op_eq(qml.RZ) | qml.noise.op_eq(qml.RY)
+        c1 = qml.noise.op_eq(qml.CNOT)
+
+        n0 = qml.noise.partial_wires(qml.QubitChanel(kraus_ops1))
+        n1 = qml.noise.partial_wires(qml.QubitChanel(kraus_ops2))
+
+        equivalent_pl_noise_model = qml.NoiseModel({c0: n0, c1: n1})
+    """
+    qerror_dmap, _ = _build_noise_model_map(noise_model, **kwargs)
```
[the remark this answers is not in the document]
obliviateandsurrender: Yep, `qerror_dmap` does stand for quantum error default map. I generally err to the side of using concise enough variable names with minimal underscores to keep the code clean. I will add a comment over the top to help with verbosity.

pennylane_qiskit/noise_models.py
obliviateandsurrender: Made the note more verbose.

pennylane_qiskit/noise_models.py
```diff
@@ -0,0 +1,139 @@
...
+        noise_op = _build_qerror_op(error, **kwargs)
+        qerror_dmap[noise_op][AnyWires].append(qiskit_op_map[gate_name])
+
+    # Add specific qubit errors
+    for gate_name, qubit_dict in noise_model._local_quantum_errors.items():
+        for qubits, error in qubit_dict.items():
+            noise_op = _build_qerror_op(error, **kwargs)
+            qerror_dmap[noise_op][qubits].append(qiskit_op_map[gate_name])
+
+    # TODO: Add support for the readout error
```
austingmhuang, line 134: Do we have a shortcut story for this?

austingmhuang (approved):
No major comments or issues from my side. Just left some nitpicky stuff 😅

pennylane_qiskit/noise_models.py
austingmhuang: We should probably add that to the comment here then 😄

lillian542 (approved):
Looks good to me! 🎉

PR #578, opened 2024-07-29
Parity of noise model conversion `args`

obliviateandsurrender:
Brings `kwargs` for noise models conversion functionality to `args`


pennylane_qiskit/converter.py
```diff
@@ -1281,14 +1281,14 @@ def load_noise_model(noise_model, **kwargs) -> qml.NoiseModel:
...
-        if isinstance(error, qml.QubitChannel) and not kwargs.get("verbose", False):
+        if isinstance(error, qml.QubitChannel) and not verbose:
             kraus_shape = qml.math.shape(error.data)
             num_kraus, num_wires = kraus_shape[0], int(np.log2(kraus_shape[1]))
             noise = _rename(f"QubitChannel(num_kraus={num_kraus}, num_wires={num_wires})")(noise)
 
-        if isinstance(error, qml.QubitChannel) and kwargs.get("verbose", False):
-            if (decimals := kwargs.get("decimal_places", None)) is not None:
-                kraus_matrices = list(np.round(error.data, decimals=decimals))
+        if isinstance(error, qml.QubitChannel) and verbose:
```
isaacdevlugt, line 1289: Can this branch of the control flow be `elif`? Does it matter? 🤔

isaacdevlugt (approved):
Thanks @obliviateandsurrender

pennylane_qiskit/converter.py
obliviateandsurrender: Doesn't really matter, but changed it! 👍

pennylane_qiskit/converter.py
```diff
@@ -1281,14 +1281,13 @@ def load_noise_model(noise_model, **kwargs) -> qml.NoiseModel:
...
-        if isinstance(error, qml.QubitChannel) and kwargs.get("verbose", False):
-            if (decimals := kwargs.get("decimal_places", None)) is not None:
-                kraus_matrices = list(np.round(error.data, decimals=decimals))
+        if isinstance(error, qml.QubitChannel):
+            if not verbose:
+                kraus_shape = qml.math.shape(error.data)
+                n_kraus, n_wires = kraus_shape[0], int(np.log2(kraus_shape[1]))
+                noise = _rename(f"QubitChannel(num_kraus={n_kraus}, num_wires={n_wires})")(noise)
+            elif verbose and decimal_places is not None:
+                kraus_matrices = list(np.round(error.data, decimals=decimal_places))
```
austingmhuang, line 1290: What do we do when `isinstance(error, qml.QubitChannel)` is `False`? Is it possible to use a guard clause instead?

austingmhuang (commented):
Looks good @obliviateandsurrender! Just some minor comments/questions.

pennylane_qiskit/converter.py
```diff
@@ -1225,19 +1225,19 @@ def _expr_eval_clvals(clbits, clvals, expr_func, bitwise=False):
     return condition_res
 
 
-def load_noise_model(noise_model, **kwargs) -> qml.NoiseModel:
+def load_noise_model(
+    noise_model, verbose: bool = False, decimal_places: Union[int, None] = None
+) -> qml.NoiseModel:
```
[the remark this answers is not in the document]
obliviateandsurrender: Afaik, we have been using type hints when possible in the functions in the `converter.py` since it's massive and it is the least we can do for maintainability.

pennylane_qiskit/converter.py
obliviateandsurrender: If we don't have a `QubitChannel` then we don't modify their `repr` and use them as it is. This is just a special case for handling Kraus matrices representation so there's no need for using a guard clause.

pennylane_qiskit/converter.py
```diff
@@ -1225,19 +1225,19 @@ def _expr_eval_clvals(clbits, clvals, expr_func, bitwise=False):
...
-        kwargs: optional keyword arguments for the conversion of the noise model
-
-    Keyword Arguments:
-        verbose (bool): show a complete list of Kraus matrices for ``qml.QubitChannel`` instead of
-            number of Kraus matrices and the number of qubits they act on. The default is ``False``
-        decimal_places (int): number of decimal places to round the Kraus matrices when they are
-            being displayed for each ``qml.QubitChannel`` with ``verbose=False``.
+        verbose (bool): when printing a ``NoiseModel``, a complete list of Kraus matrices for each ``qml.QubitChannel``
+            is displayed with ``verbose=True``. By default, ``verbose=False`` and only the number of Kraus matrices and
+            the number of qubits they act on is displayed for brevity.
```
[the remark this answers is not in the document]
obliviateandsurrender: This is in parity with the discussions we had on documentation on the PL side. 👍

PR #579, opened 2024-07-29
Delete ibmq tests

austingmhuang:
The`ibmq_tests.yml` file tests two files, namely `test_ibmq.py` `test_runtime.py`,  that have been deleted with the merging of the new Qiskit Device. This PR removed the `ibmq_tests.yml` file.

We no longer have these ibmq tests because there is no cloud simulator for us to run the tests on anymore. We don't have any tests that require an IBM Quantum account token at the moment (we use local simulators for everything). There is the option to replace these tests with tests that use a real IBM device, but that would potentially require a paid plan depending on the number of calls to the backend.


austingmhuang:
[sc-69774]

soranjh (approved):

obliviateandsurrender (approved):

PR #580, opened 2024-08-27
Plugin bug fix

austingmhuang:
Fixes errors in plugin text matrix.

### DIFF
--- a/tests/test_base_device.py
+++ b/tests/test_base_device.py
@@ -1081,6 +1081,9 @@ class TestExecution:
         ],
     )
     @flaky(max_runs=10, min_passes=7)
+    @pytest.mark.xfail(
+        reason="Qiskit variances is different from PennyLane variances, discussion pending on resolution"
+    )
     def test_estimator_with_various_multi_qubit_pauli_obs(
         self, mocker, wire, angle, op, multi_q_obs
     ):
@@ -1090,8 +1093,8 @@ class TestExecution:
         correspond correctly (wire ordering convention in Qiskit and PennyLane don't match.)
         """
 
-        pl_dev = qml.device("default.qubit", wires=[0, 1, 2, 3])
-        dev = QiskitDevice(wires=[0, 1, 2, 3], backend=aer_backend)
+        pl_dev = qml.device("default.qubit", wires=4)
+        dev = QiskitDevice(wires=4, backend=aer_backend)
 
         sampler_execute = mocker.spy(dev, "_execute_sampler")
         estimator_execute = mocker.spy(dev, "_execute_estimator")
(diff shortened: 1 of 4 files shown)


tests/test_base_device.py
```diff
@@ -1090,8 +1093,8 @@ def test_estimator_with_various_multi_qubit_pauli_obs(
         correspond correctly (wire ordering convention in Qiskit and PennyLane don't match.)
         """
 
-        pl_dev = qml.device("default.qubit", wires=[0, 1, 2, 3])
-        dev = QiskitDevice(wires=[0, 1, 2, 3], backend=aer_backend)
+        pl_dev = qml.device("default.qubit", wires=4)
+        dev = QiskitDevice(wires=4, backend=aer_backend)
```
austingmhuang, line 1097: I changed this since the previous way was a little redundant.

tests/test_base_device.py
```diff
@@ -1081,6 +1081,9 @@ def test_estimator_with_different_pauli_obs(self, mocker, wire, angle, op, expec
         ],
     )
     @flaky(max_runs=10, min_passes=7)
+    @pytest.mark.xfail(
+        reason="Qiskit variances is different from PennyLane variances, discussion pending on resolution"
+    )
```
austingmhuang, line 1086: We xfail these tests because the qiskit's std is different from what is expected.

Long story short: if you compute var(O) manually using pennylane/expectation values using `<O^2> - <O>^2` where O is the operator, and then calculate the `std` using sqrt(var(O))/sqrt(N), you will get a different `std` from Qiskit 1.2.0. In Qiskit <1.1.0, you will get the same `std` for commuting observables but not for non-commuting observables.

If O is a linear combination of pauli terms (Pi, Pj...), Qiskit computes the variance as Var(Pi) + Var(Pj) + ..., which doesn't seem right to me. Perhaps they're doing some kind of approximation?

tests/test_qiskit_device.py
```diff
@@ -182,8 +182,8 @@ def test_warning_raised_for_hardware_backend_analytic_expval(self, recorder):
             "device are estimates based on samples."
         )
 
-        # Two warnings are being raised: one about analytic calculations and another about deprecation.
-        assert len(record) == 2
+        # Three warnings are being raised: one about analytic calculations.
```
soranjh, line 185: What are the other two warning?

tests/test_base_device.py
soranjh: Sounds good to me.

tests/test_qiskit_device.py
austingmhuang, line 185:
```suggestion
        # One warning raised about analytic calculations.
```

austingmhuang:
[sc-72050]

PR #581, opened 2024-08-28
Change back to `QuantumCircuit.from_qasm_str`

lillian542:
When updating QASM load functionality recently, we switched from `QuantumCircuit.from_qasm_str` to `qiskit.qasm2.loads` for loading a QASM string into a `QiskitCircuit`. But `qiskit.qasm2.loads` seems not to support as many operations as `QuantumCircuit.from_qasm_str`, and specifically is failing to load the `u` gate.


KetpuntoG (approved):
LGTM, thanks!

PR #591, opened 2024-10-21
Fix test improperly comparing measurement values

albi3ro:
A test was comparing equality of measurment values using `==`, even though that returns another measurement value instead of boolean.  Now that measruement values are intrinsically truthy and instead raise an error when used as a boolean, this test is raising an error.

This change fixes it to use `qml.equal` and `qml.assert_equal` instead of `bool(m1 == m2)`.

### DIFF
--- a/tests/test_converter.py
+++ b/tests/test_converter.py
@@ -1647,10 +1647,12 @@ class TestConverterIntegration:
         qc1.measure(1, 1)
 
         qtemp, qtemp1 = load(qc), load(qc1)
-        assert qtemp()[0] == qml.measure(0) and qtemp1()[0] == qml.measure(2)
+        assert qml.equal(qtemp()[0], qml.measure(0))
+        assert qml.equal(qtemp1()[0], qml.measure(2))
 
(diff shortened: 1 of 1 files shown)


tests/test_converter.py
```diff
@@ -1647,10 +1647,12 @@ def test_diff_meas_circuit(self):
         qc1.measure(1, 1)
 
         qtemp, qtemp1 = load(qc), load(qc1)
-        assert qtemp()[0] == qml.measure(0) and qtemp1()[0] == qml.measure(2)
+        assert qml.equal(qtemp()[0], qml.measure(0))
```
mudit2812, line 1650: `qml.equal` for `MeasurementValue` compares the stored `MidMeasureMP`s, and their equality depends on the `id` of the measurement. So you would have to instead compare the `wires`, `reset`, and `postselect` attributes manually.

Maybe also worth changing `qml.equal` wrt `MidMeasureMP` 😅 .

albi3ro: 😬

mudit2812 (approved):

PR #593, opened 2024-10-29
Let users import from pennylane_qiskit

austingmhuang:
In the documentation for qiskit_session and [the demo for using IBM with pennylane-qiskit](https://pennylane.ai/qml/demos/ibm_pennylane/), we import qiskit_session using `from pennylane_qiskit import qiskit_session`. But this doesn't actually work because it is not in `__init__.py`. 

[sc-77085]


CatalinaAlbornoz (approved):
Thanks for the fix!

lillian542 (approved):

PR #619, opened 2025-02-21
Enable Job mode when no session is required

felipelewyee:
Hello, I would like to propose to:
- Enable the "Job" mode when no session is explicitly used.
- Group estimator circuits and sampler circuits into pubs for a single call to each primitive.

Sessions can still be used as before if explicitly declared, remaining useful when continuous job submission is required.

**What problem does this solve?**
The current behavior is to use Sessions always. However, this leads to expensive usage of the quantum computer. For example, the following requires to run two circuits:
```python
import pennylane as qml
from qiskit_ibm_runtime import QiskitRuntimeService

service = QiskitRuntimeService(channel="ibm_quantum")
backend = service.least_busy(simulator=False, operational=True)

dev = qml.device("qiskit.remote", wires=127, backend=backend, shots=1000) 
@qml.qnode(dev)
def test():
    qml.Hadamard(wires=0)
    return qml.expval(qml.PauliX(0)), qml.expval(qml.PauliY(0))
test()
```

Each circuit takes around 3 seconds, but the actual execution takes 33 seconds because the session remains active while the classic computer is doing transpilation and postporcessing between executions. The problem becomes more serious for larger circuits with many measurements, leading to sessions of more than an hour for circuits that run in a few minutes.

With this proposed PR, the two circuits will be transpiled and no session will be opened, leading to only 6 seconds of quantum computer usage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Remove support for Qiskit 1.0 and add support for Qiskit 2.0.
   [(#653)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/653)
+  [(#659)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/659)
 
 * Remove support for Python 3.10 and add support for Python 3.13.
   [(#646)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/646)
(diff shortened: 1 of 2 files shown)


CatalinaAlbornoz:
Hi @felipelewyee, thank you for opening this PR! Our team will review when they have capacity. It might take several weeks though since we have a new PennyLane release coming up in the next few weeks.

PR #627, opened 2025-04-28
Update Sphinx-action version to tag

runora95:
docs.yml currently uses sphinx-action to run the documentation checks. Sphinx-action is being upgraded to python 3.10 which is incompatible with the existing docs.yml workflow. This PR will update docs.yml to use a version of sphinx-action that is pegged to python 3.9 and a lower version of sphinx. 

This repo is already using sphinx-7.4.7 with python3.9.


JerryChen97 (approved):
Thanks a lot for the fixing!

PR #640, opened 2025-07-09
Fix number of wires in `QubitChannel` in converted noise models

obliviateandsurrender:
Use information from the Kraus matrices to fix the number of wires in `QubitChannel` in converted noise models.

More context - 

For a noise model, where noise function is constrcuted with a `qml.noise.conditional.partial_wires` - we zip over the list of temporary wires (now called as "DummyTempWire") and operator wires to obtain the set of wires on which the noise has to be applied. This can be seen ([here](https://github.com/PennyLaneAI/pennylane/blob/08b3c20567d29698b3f4fa86e85efa80f0ea8e9c/pennylane/noise/conditionals.py#L760)). Currently, by default, the number of temporary wires were being set as `one`, but in practice, they should match the wires on which the included Kraus matrices will be applied. So in this PR, we are taking that into account.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,11 +41,15 @@
 
 * Stops queuing a mid circuit measurement on first use of `qml.from_qiskit`.
  [(#630)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/630)
- 
+
+* `qml.QubitChannel` in a converted noise model should now act on correct number of qubits.
+  [(#640)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/640)
+
  ### Contributors ✍️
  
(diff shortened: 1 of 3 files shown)


JerryChen97 (changes requested):
Would you like to elaborate over the analysis about what happened in the bug that user reported? TBH I'm lost here... 😓

JerryChen97:
Plus it will be better if we can add a minimized version code sample derived from the bug we are fixing, into the test

JerryChen97 (approved):

astralcai (approved):

andrijapau (changes requested):
Ignore me.

PR #642, opened 2025-07-11
Remove ibm_provider from setup.py

austingmhuang:
IBM provider is deprecated and it is raising errors in our quantum volume demo even though the plugin doesn't actually use it. We just need to remove it from setup.py and update conftest to use runtime instead.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -20,7 +20,7 @@ import pytest
 import numpy as np
 
 import pennylane as qml
-from qiskit_ibm_provider import IBMProvider
+from qiskit_ibm_runtime import QiskitRuntimeService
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -50,17 +50,21 @@
 
 @pytest.fixture
 def skip_if_no_account():
+    """Skips test if no IBM Quantum account is found."""
     t = os.getenv("IBMQX_TOKEN", None)
+    if t is None:
+        if not QiskitRuntimeService.saved_accounts():
```
obliviateandsurrender, line 56: We can combine them.

andrijapau (approved):

PR #650, opened 2025-08-07
Move `UserWarning` regarding `shots=None` to preprocess

JerryChen97:
qiskit does not support analytic mode right now. In the past, we prevent this by warning at init and convert to 1024. But it should belong to the preprocessing.
Previous behaviour of hard-converting `device.shots` blocks us from properly deprecating the `shots` on device init at https://github.com/PennyLaneAI/pennylane-qiskit/pull/652
[sc-97314]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,12 +11,12 @@
 
 * `QiskitDevice` no longer warns or changes `shots=None` at initialization.
   Instead, the `analytic_warning` transform issues a warning only at execution time,
-  and leaves `shots=None` unchanged so the Qiskit backend can default to 1024 shots.
+  and leaves `shots=None` unchanged; the Qiskit backend will then set it's own default for the number of shots.
   To ensure a consistent experience, use `qml.set_shots(shots)` on each QNode executed
   with `QiskitDevice`:
 
   ```python
-  dev = qml.device("qiskit.aer", wires=2, shots=None)  # Previously issued warning now
+  dev = qml.device("qiskit.aer", wires=2)
 
   @qml.set_shots(1000)
   @qml.qnode(dev)
@@ -25,7 +25,6 @@
       qml.CNOT(wires=[0, 1])
       return qml.expval(qml.PauliZ(0))
 
-  # Now you must set shots before execution to avoid errors
   # or equivalently, without the decorator:
   #  circuit = qml.set_shots(circuit, shots=1000)
   result = circuit()
--- a/pennylane_qiskit/qiskit_device.py
+++ b/pennylane_qiskit/qiskit_device.py
@@ -64,8 +64,8 @@ def analytic_warning(tape):
     """Transform that adds a warning for circuits without shots set."""
     if not tape.shots:
         warnings.warn(
-            "Expected an integer number of shots, but received shots=None. Defaulting "
-            "to 1024 shots. The analytic calculation of results is not supported on "
+            "Expected an integer number of shots, but received shots=None. A default "
+            "number of shots will be selected by the Qiskit backend. The analytic calculation of results is not supported on "
             "this device. All statistics obtained from this device are estimates based "
             "on samples.",
             UserWarning,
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,10 @@
 
 ### Internal changes ⚙️
 
+* `QiskitDevice` does not error out `None` shots at initialization any more.
```
lillian542, line 16: It never errored out, it just issued a warning and set device shots to 1024.

```suggestion
* `QiskitDevice` does not issue a warning when device shots are `None` shots at initialization any more.
```

lillian542 (commented):
Change log should be updated to reflect that this was not an error, but a warning.

I think this is worth running past @isaacdevlugt - its in line with the intended deprecation, but it does mean we are introducing an error where we previously would have raised a warning and handled the mistake for the user, so technically this is a breaking change, and its not impossible that it breaks someone's workflow. And if it does, the error message is confusing, because previously when they ran that code, it _didn't_ produce an analytic execution.

JerryChen97:
> Change log should be updated to reflect that this was not an error, but a warning.
> 
> I think this is worth running past @isaacdevlugt - its in line with the intended deprecation, but it does mean we are introducing an error where we previously would have raised a warning and handled the mistake for the user, so technically this is a breaking change, and its not impossible that it breaks someone's workflow. And if it does, the error message is confusing, because previously when they ran that code, it _didn't_ produce an analytic execution.

Thanks for raising this point! Yeah this is very critical. I think if we want to keep the same support as before, then we need to introduce another transform to automatically assign back `shots=1024` here in preprocess; or we leave it as breaking change.

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@
 
 ### Internal changes ⚙️
 
+* `QiskitDevice` does not issue a warning when device shots are `None` shots at initialization any more.
+  Instead, a `no_analytic` preprocess transform is applied.
+  [(#650)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/650)
```
lillian542, line 18: I'm sure Isaac will have thoughts on the exact messaging, but I would think this should also be under `Breaking changes`, and should tell people how to fix their problem using the new and approved `set_shots` method. Something like:

```suggestion
* `QiskitDevice` does not issue a warning when device shots are `None` shots at initialization any more.
  Instead, a `no_analytic` preprocess transform is applied. This means an 
  error will now be raised if shots are not specified. Please use... 
  
[[ insert example with `set shots` here ]]
 
 on all `qnodes` that will be executed on the `QiskitDevice`.
  [(#650)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/650)
```

lillian542:
Also tagging @gabrielasd for visibility :)

gabrielasd:
> > so we warn people that what would normally happen in PennyLane with any other device ... isn't what is happening
> 
> That's a good point 🤔... @gabrielasd what do you think?

I like the idea that the users is clear of what is happening, so from that perspective I am in favor of the warning. Or if not a warning, could a clarification that if shots=None, Qiskit will set shots to 1024 be added to the documentation of the shots parameter?

JerryChen97:
It seems that I can't add @gabrielasd as reviewer :sweat:

isaacdevlugt (approved):

lillian542 (commented):
Since we are letting Qiskit pick the default instead of setting it ourselves to 1024, I think we it might be better to reword the warning so that we say Qiskit will select a default. It's more likely to stay accurate if we don't make promises about what they will do. I left a few suggestions to that effect.

Also there are still a couple of places that reference raising en error, but we've chosen to stick with a warning.

CHANGELOG.md
```diff
@@ -9,6 +9,31 @@
 * Remove support for Python 3.10 and add support for Python 3.13.
   [(#646)](https://github.com/PennyLaneAI/pennylane-qiskit/pull/646)
 
+* `QiskitDevice` no longer warns or changes `shots=None` at initialization.
+  Instead, the `analytic_warning` transform issues a warning only at execution time,
+  and leaves `shots=None` unchanged so the Qiskit backend can default to 1024 shots.
```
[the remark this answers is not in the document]
JerryChen97: good idea!!!

PR #653, opened 2025-08-14
feat: bump `qiskit` eco-system to latest

andrijapau:
## Context:

This plug-in has been on an outdated version of the Qiskit eco-system. We want to be compatible with the latest version of Qiskit to un-block users.

## Description of change:

This PR was generated by bumping the ecosystem to latest and then fixing all of the CI failures until it is 💚 .

[sc-85737]
[sc-88695]


requirements.txt
```diff
@@ -1,5 +1,5 @@
 pennylane
-qiskit<1.3
-qiskit-ibm-runtime<=0.29
-numpy
-sympy
+qiskit>=2.0
+qiskit-aer>=0.17
+qiskit-ibm-runtime>=0.41
+sympy
```
andrijapau, line 5:
```suggestion
sympy

```

tests/test_converter.py
andrijapau, line 1: Support from @obliviateandsurrender here. 👍🏼

tests/test_qiskit_device.py
andrijapau, line 1: Again get rid of V1 pipeline

pennylane_qiskit/converter.py
```diff
@@ -586,7 +583,7 @@ def _function(*args, params: dict = None, wires: list = None, **kwargs):
                     )
 
             # Check if it is a conditional operation or conditional instruction
-            if instruction.condition or isinstance(instruction, ControlFlowOp):
+            if isinstance(instruction, ControlFlowOp):
```
andrijapau, line 586: instruction.condition no longer exists

pennylane_qiskit/converter.py
```diff
@@ -941,10 +938,6 @@ def _conditional_funcs(inst, operation_class, branch_funcs, ctrl_flow_type):
     """
     true_fns, false_fns = [operation_class], [None]
 
-    # Logic for using legacy c_if
-    if not isinstance(inst, ControlFlowOp):
-        return true_fns, false_fns, inst.condition
```
andrijapau, line 946: legacy code

pennylane_qiskit/qiskit_device.py
```diff
@@ -193,7 +193,7 @@ def circuit(x):
     # Code to acquire session:
     existing_session = device._session
 
-    session_options = {"backend": device.backend, "service": device.service}
+    session_options = {"backend": device.backend}
```
andrijapau, line 196: service doesn't exist as a session option anymore

pennylane_qiskit/qiskit_device_legacy.py
```diff
@@ -342,9 +340,15 @@ def compile(self):
...
-        compiled_circuits = transpile(self._circuit, backend=compile_backend, **self.transpile_args)
-        return compiled_circuits
+        return transpile(
+            self._circuit,
+            backend=self.compile_backend or self.backend,
+            **self.transpile_args,
+            # Disable routing to keep SWAP in circuit rather than
+            # using ElidePermutation to shuffle qubit indices around.
+            # See https://github.com/Qiskit/qiskit/issues/13144 for more details.
+            routing_method="none",
```
lillian542, line 350: I can see how this enables SWAP, but will it make it hard to run on hardware that needs to be transpiled with respect for physical routing constraints? I'm just wondering if maybe we're fixing one problem by creating a worse problem.

tests/conftest.py
```diff
@@ -114,13 +114,11 @@ def hardware_backend(request):
 
 @pytest.fixture(params=test_devices)
 def device(request, backend, shots):
-    print("getting a device")
     if backend not in state_backends:
         if shots is None:
             pytest.skip("Hardware simulators do not support analytic mode")
 
     if backend == "aer_simulator" and not issubclass(request.param, AerDevice):
-        print("I should be skipping this test")
```
lillian542, line 123: oops, my bad 🙈

tests/test_base_device.py
```diff
@@ -175,7 +153,7 @@ def test_compile_backend_kwarg(self):
         assert dev2._compile_backend != dev2._backend
         assert dev2._compile_backend == compile_backend
 
-    @pytest.mark.parametrize("backend", [aer_backend, legacy_backend])
+    @pytest.mark.parametrize("backend", [aer_backend])
```
lillian542, line 156: Huh. Isn't it kind of weird that we don't run this test with `FakeManilaV2`?

tests/test_base_device.py
```diff
@@ -138,7 +117,6 @@ def test_v1_and_v2_mocked(self, backend):
     @pytest.mark.parametrize(
         "backend, shape",
         [
-            (FakeManila(), (1024,)),
             (FakeManilaV2(), (1024,)),
         ],
     )
```
lillian542, line 122: For tidiness, can we go ahead and just un-parametrize this test, and then also update the name and docstring to get rid of the V1 mentions.

tests/test_base_device.py
```diff
@@ -128,7 +107,7 @@ class TestSupportForV1andV2:
 
     @pytest.mark.parametrize(
         "backend",
-        [legacy_backend, aer_backend, mocked_backend],
+        [aer_backend, mocked_backend],
     )
     def test_v1_and_v2_mocked(self, backend):
```
lillian542, line 112:
```suggestion
    def test_mocked_backend(self, backend):
```

lillian542 (commented):
Looks good, mostly minor comments. My only real concern is whether the solution of skipping routing in transpiration to support the SWAP gates is going to be a problem for transpiration for hardware.

tests/test_qiskit_device.py
```diff
@@ -115,7 +93,6 @@ def test_v1_and_v2_mocked(self, dev_backend):
     @pytest.mark.parametrize(
         "dev_backend",
         [
-            FakeManila(),
             FakeManilaV2(),
         ],
     )
```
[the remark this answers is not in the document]
andrijapau: done for this file in https://github.com/PennyLaneAI/pennylane-qiskit/pull/653/commits/d127ef786e4fe0772168b7e50fd2d3c6eae5a831

tests/test_base_device.py
andrijapau: Done for this file in https://github.com/PennyLaneAI/pennylane-qiskit/pull/653/commits/064cb9cf6fdc1a58f736e0d01ceee0bb4f4dad4f

tests/test_base_device.py
andrijapau: Hmm ... looking into it.

andrijapau: Fixed in https://github.com/PennyLaneAI/pennylane-qiskit/pull/653/commits/e80f6ceac9f0d792395b349de13a3f41f6c93c7e

pennylane_qiskit/qiskit_device_legacy.py
```diff
@@ -130,9 +130,7 @@ def __init__(self, wires, provider, backend, shots=None, **kwargs):
 
         # Perform validation against backend
         backend_qubits = (
-            backend.num_qubits
-            if isinstance(backend, BackendV2)
-            else self.backend.configuration().n_qubits
+            backend.num_qubits if isinstance(backend, BackendV2) else self.backend.num_qubits
```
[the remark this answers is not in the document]
andrijapau: Actually, I think the previous implementation is a big sloppy. We should be using ONLY `self.backend` as this block above this logic,
```python
        if isinstance(backend, Backend):
            self._backend = backend
            self.backend_name = _get_backend_name(backend)
        elif provider is None:
            raise ValueError("Must pass a provider if the backend is not a Backend instance.")
        else:
            try:
                self._backend = provider.get_backend(backend)
            except QiskitBackendNotFoundError as e:
                available_backends = list(map(_get_backend_name, provider.backends()))
                raise ValueError(
                    f"Backend '{backend}' does not exist. Available backends "
                    f"are:\n {available_backends}"
                ) from e

            self.backend_name = _get_backend_name(self._backend)

        self._capabilities["returns_state"] = self._is_state_backend
```
is essentially handling the `backend` kwarg which gets processed into a `self.backend` object.

Long story short, I think we can just use `self.backend.num_qubits` and we're safe.

pennylane_qiskit/qiskit_device_legacy.py
lillian542: You can use a fake backend with routing requirements. `GenericBackendV2` with a coupling map. I think we use it in some of the tests, and it shouldn't require hardware access.

lillian542: Discussed offline. This is only affecting the legacy device, which now is only used by simulators, so we don't need to worry about this after all 😌

lillian542 (commented):
Actually one more minor thing (only if it's not too much hassle to do it): could you change the name of the tests from `Tests for 1.0` to just `Tests`? 

It's a relic from when we were running both 0.46 and 1.0 tests, and now its no longer accurate 😅

andrijapau:
> Actually one more minor thing (only if it's not too much hassle to do it): could you change the name of the tests from `Tests for 1.0` to just `Tests`?
> 
> It's a relic from when we were running both 0.46 and 1.0 tests, and now its no longer accurate 😅

Done in https://github.com/PennyLaneAI/pennylane-qiskit/pull/653/commits/3b4f5dc87746f3f2e42018a462ec88b4b3d1c421 😎

andrijapau:
ℹ️  Performed some offline testing of the `qiskit.remote` backend and basic circuits are executing without error.

PR #654, opened 2025-08-19
Overload `batch_transform`

JerryChen97:
In favour of our new shots api (i.e. the shots information should be only stored on qnode and tapes, instead of devices), the shots setting at device initialization has been deprecated on PennyLane. To ease the transition on the side of plugins, qiskit plugin need to also change the default shots to `None` and also when and where to emit the warning. The best candidate up to now is the `expand_fn` or `batch_execute` because legacy facade will apply these two first before actually executing any circuit(s) on the legacy device.

### DIFF
--- a/pennylane_qiskit/qiskit_device.py
+++ b/pennylane_qiskit/qiskit_device.py
@@ -341,7 +341,7 @@ class QiskitDevice(Device):
         self,
         wires,
         backend,
-        shots=1024,
+        shots=None,
         session=None,
         compile_backend=None,
         **kwargs,
@@ -524,7 +524,7 @@ class QiskitDevice(Device):
                 "Please use the `shots` keyword argument instead. The number of shots "
                 f"{shots} will be used instead."
             )
-        kwargs["default_shots"] = shots
+        kwargs["default_shots"] = shots or 1024
 
         kwargs, transpile_args = self.get_transpile_args(kwargs)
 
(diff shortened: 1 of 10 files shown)


pennylane_qiskit/qiskit_device.py
```diff
@@ -524,7 +524,7 @@ def _process_kwargs(self, kwargs):
                 "Please use the `shots` keyword argument instead. The number of shots "
                 f"{shots} will be used instead."
             )
-        kwargs["default_shots"] = shots
+        kwargs["default_shots"] = shots or 1024
```
andrijapau, line 527: This magic number bothers me so much lol.

pennylane_qiskit/qiskit_device_legacy.py
```diff
@@ -149,6 +142,14 @@ def __init__(self, wires, provider, backend, shots=1024, **kwargs):
 
         self.process_kwargs(kwargs)
 
+    def batch_transform(self, circuit):
+        """Batch transform the circuit"""
+        if not (circuit.shots or self.shots or self._is_state_backend):
+            warnings.warn(self.analytic_warning_message.format(self.backend_name), UserWarning)
+            circuit = circuit.copy(shots=1024)
```
andrijapau, line 149: Since when do we have a copy method?

andrijapau (commented):
Few more comments, but otherwise I'm pretty happy with it

pennylane_qiskit/qiskit_device.py
JerryChen97: Yeah, let's have a follow-up PR to break this feature next iteration lol anyways we have to prepare for the new version of qiskit soon, time to drop all of these cryptic magic

PR #655, opened 2025-08-19
create `pytest.ini` to raise PLDW as errors

JerryChen97:
So that we can catch `PennyLaneDeprecationWarning` before it errors out in PTM


andrijapau (approved):
Nice! 🚀

PR #656, opened 2025-09-02
A better `test_no_shots_warning`

JerryChen97:
Previously, the `test_no_shots_warning` in `test_base_device.py` has a ~2% chance to fail theoretically.
With a newly appointed circuit to test against, we shall be able to see it pass deterministically as long as the finite-shot mode was used.


andrijapau (approved):
LGTM

PR #677, opened 2025-12-11
Fix a bug with Barrier not being supported in Azure-qiskit

DSGuala:
**Context:**
The current circuit conversion inserts a `barrier` between the `tape.operations` and the terminal measurements when converting a PL circuit to a Qiskit circuit. Some devices don't support `barrier`, and as far as we can see, it isn't technically necessary.

This PR removes a line in `circuit_to_qiskit` that added a barrier between a circuit's gates and measurements.

Also the new version of `black` has added some extra whitespace etc. There are only 4 lines of actual changes.

**Benefits:**
Allows devices like Azure that don't include `barrier` support to work with the Remote device.

**Possible Drawbacks:**
We aren't _aware_ of anything that was dependent on the `barrier`, and removing it doesn't  cause any tests to fail, but it's not impossible that some device somewhere was expecting it to be there. We can cross that bridge if we come to it.

**Related GitHub Issues:**
https://github.com/PennyLaneAI/pennylane/issues/8657

[sc-104090]


JerryChen97 (approved):

JerryChen97 (commented):
Actually, mind adding a changelog? @DSGuala @lillian542

JerryChen97 (changes requested):
LGTM!

JerryChen97 (approved):

PR #684, opened 2026-01-28
Compatibility with the graph-based decompositions

astralcai:
Make the necessary updates for the pennylane-qiskit plugin to be compatible with graph decomposition enabled.


JerryChen97 (approved):
Thanks for fixing!

PR #706, opened 2026-05-13
Adding pyproject.toml and updating workflows

runora95:
Changing `pennylane-qiskit` from using setup.py to using pyproject.toml. The workflows have been updated accordingly.


setup.py
```diff
@@ -1,90 +1,3 @@
-# Copyright 2018 Carsten Blank
+from setuptools import find_packages, setup
```
andrijapau: We don't need `find_packages` here right?

andrijapau (approved):
LGTM! 🚀 

Happy to see these requirements files gone.

JerryChen97 (approved):
🚀

